PR #29, opened 2026-01-21
feat(auth): add automatic token refresh using refresh_token

BYK:
## Summary

Implements transparent OAuth token refresh so users stay authenticated indefinitely without manual re-login (unless the refresh token is revoked).

## Changes

- **Proactive refresh**: Token is automatically refreshed when <10% of lifetime remains
- **Reactive refresh**: On 401 API response, forces token refresh and retries request once
- **Concurrency protection**: Multiple simultaneous requests share a single refresh operation
- **Graceful degradation**: On refresh failure, auth is cleared and user is prompted to re-login
- **Manual refresh command**: `sentry auth refresh` for testing/debugging (`--force`, `--json` flags)
- **Status display**: `sentry auth status` now shows auto-refresh enabled/disabled

## API Design

```typescript
// refreshToken() with options object
type RefreshTokenOptions = { force?: boolean };
type RefreshTokenResult = { token, refreshed, expiresAt?, expiresIn? };

refreshToken()                    // Refresh if <10% lifetime remaining
refreshToken({ force: true })     // Always refresh (used by 401 handler)
```

## Files Changed

| File | Changes |
|------|---------|
| `src/lib/config.ts` | New `refreshToken({ force })` API with rich result object |
| `src/lib/api-client.ts` | Uses `refreshToken()` + 401 handler with `{ force: true }` |
| `src/lib/oauth.ts` | Added `refreshAccessToken()` (RFC 6749 Section 6) |
| `src/lib/formatters/human.ts` | Added `formatDuration()` helper |
| `src/types/config.ts` | Added `issuedAt` field to track token lifetime |
| `src/commands/auth/refresh.ts` | Manual refresh command |
| `src/commands/auth/status.ts` | Shows auto-refresh status |
| `test/lib/api-client.test.ts` | Tests for 401 retry with force refresh |

## Usage

Token refresh happens automatically - no user action needed!

```bash
# View auth status (shows auto-refresh enabled/disabled)
sentry auth status

# Manually trigger refresh (for testing)
sentry auth refresh

# Force refresh even if token is still valid
sentry auth refresh --force
```

## Testing

```bash
bun test test/lib/api-client.test.ts
bun test test/lib/config.test.ts
```

### DIFF
--- a/test/e2e/auth.test.ts
+++ b/test/e2e/auth.test.ts
@@ -82,20 +82,24 @@ describe("sentry auth status", () => {
 });
 
 describe("sentry auth login --token", () => {
-  test("stores valid API token", async () => {
-    const result = await runCli(["auth", "login", "--token", TEST_TOKEN], {
-      env: { SENTRY_CLI_CONFIG_DIR: testConfigDir },
-    });
+  test(
+    "stores valid API token",
+    async () => {
+      const result = await runCli(["auth", "login", "--token", TEST_TOKEN], {
+        env: { SENTRY_CLI_CONFIG_DIR: testConfigDir },
+      });
 
-    expect(result.stdout).toContain("Authenticated");
-    expect(result.exitCode).toBe(0);
+      expect(result.stdout).toContain("Authenticated");
+      expect(result.exitCode).toBe(0);
 
-    // Verify token was stored
-    const statusResult = await runCli(["auth", "status"], {
-      env: { SENTRY_CLI_CONFIG_DIR: testConfigDir },
-    });
-    expect(statusResult.stdout).toContain("Authenticated");
-  });
+      // Verify token was stored
+      const statusResult = await runCli(["auth", "status"], {
+        env: { SENTRY_CLI_CONFIG_DIR: testConfigDir },
+      });
+      expect(statusResult.stdout).toContain("Authenticated");
+    },
+    { timeout: 15_000 }
+  );
 
   test("rejects invalid token", async () => {
     const result = await runCli(


src/commands/auth/refresh.ts
```diff
@@ -0,0 +1,152 @@
...
+          "Run 'sentry auth login' to authenticate with OAuth and enable auto-refresh."
+      );
+    }
+
+    // Check if refresh is needed (unless --force)
+    if (!flags.force && config.auth.expiresAt) {
+      const remainingMs = config.auth.expiresAt - Date.now();
+      const remainingMinutes = Math.floor(remainingMs / 60_000);
+
+      // Only skip refresh if token has >5 minutes remaining
```
BYK: I thought we agreed to refresh the token when there's less than 10% of its total life is left. Why the logic change here?

src/commands/auth/refresh.ts
```diff
@@ -0,0 +1,152 @@
...
+          );
+        }
+        return;
+      }
+    }
+
+    // Perform refresh
+    const tokenResponse = await refreshAccessToken(config.auth.refreshToken);
+
+    // Store new tokens (server may rotate refresh token)
```
BYK: Server _does_ rotate the refresh tokens. Comment is redundant or a bit misleading.

src/lib/api-client.ts
```diff
@@ -95,6 +100,34 @@ async function createApiClient(): Promise<KyInstance> {
...
+        async (request, options, response) => {
+          // On 401, try refreshing token and retry once
+          if (response.status === 401 && !hasAttemptedRefreshRetry) {
+            hasAttemptedRefreshRetry = true;
+            try {
+              // Get a fresh token (will trigger refresh if needed)
+              const newToken = await getValidAuthToken();
+
+              // Retry the request with the new token
+              return ky(request.url, {
```
BYK: Don't know what `ky` is but it surely is a non-descriptive name. Can we do better?

test/lib/token-refresh.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      },
+    });
+
+    try {
+      await getValidAuthToken();
+      expect.unreachable("Should have thrown AuthError");
+    } catch (err) {
+      expect(err).toBeInstanceOf(AuthError);
+      expect((err as Error).message).toMatch(/no refresh token/i);
+    }
```
BYK: Why not `expect(getValidAuthToken()).rejects.toThrow()` like above?

test/lib/token-refresh.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        token: "needs-refresh",
+        issuedAt: now - lifetime + remaining,
+        expiresAt: now + remaining,
+        refreshToken: "refresh-token",
+      },
+    });
+
+    // This would attempt refresh if refreshAccessToken wasn't mocked
+    // Since it's not mocked, this will fail to refresh, but proves the threshold triggers
+    await expect(getValidAuthToken()).rejects.toThrow();
```
BYK: Can we mock and ensure the refresh flow runs?

src/lib/api-client.ts
```diff
@@ -95,6 +95,29 @@ async function createApiClient(): Promise<KyInstance> {
...
+                headers: retryHeaders,
+                retry: 0,
+              });
+            } catch {
+              return response;
+            }
+          }
+          return response;
+        },
+      ],
```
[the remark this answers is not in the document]
BYK: Fixed in ada77e3. 

Refactored `getValidAuthToken()` to `refreshToken({ force?: boolean })`. The 401 handler now calls `refreshToken({ force: true })` which bypasses the threshold check and always performs a refresh when the server rejects a token.

Added test coverage for the force refresh behavior.

PR #34, opened 2026-01-22
fix(ci): correct Craft version tag (no v prefix)

BYK:
Craft tags don't use the `v` prefix - should be `2.19.0` not `v2.19.0`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
         token: ${{ steps.token.outputs.token }}
         fetch-depth: 0
     - name: Prepare release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         token: ${{ steps.token.outputs.token }}
         fetch-depth: 0
     - name: Prepare release
-      uses: getsentry/craft@v2.19.0
+      uses: getsentry/craft@2.19.0
```
BYK:
```suggestion
      uses: getsentry/craft@v2
```

PR #50, opened 2026-01-23
ref: Rename config folder from .sentry-cli-next to .sentry

BYK:
## Summary

Renames the config folder from `~/.sentry-cli-next` to `~/.sentry` and updates the environment variable from `SENTRY_CLI_CONFIG_DIR` to `SENTRY_CONFIG_DIR`.

## Changes

- Update default config directory path in `src/lib/config.ts` and `src/context.ts`
- Rename environment variable for custom config directory
- Update JSDoc comments in DSN cache/types modules
- Update README documentation
- Update all test files to use the new environment variable name

### DIFF
--- a/test/e2e/auth.test.ts
+++ b/test/e2e/auth.test.ts
@@ -5,10 +5,9 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdirSync, rmSync } from "node:fs";
-import { join } from "node:path";
 import { CONFIG_DIR_ENV_VAR, setAuthToken } from "../../src/lib/config.js";
 import { runCli } from "../fixture.js";
+import { cleanupTestDir, createTestConfigDir } from "../helpers.js";
 
 // Test credentials from environment - these MUST be set
 const TEST_TOKEN = process.env.SENTRY_TEST_AUTH_TOKEN;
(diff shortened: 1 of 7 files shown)


test/e2e/auth.test.ts
```diff
@@ -24,13 +24,13 @@ let testConfigDir: string;
 let originalConfigDir: string | undefined;
 
 beforeEach(() => {
-  originalConfigDir = process.env.SENTRY_CLI_CONFIG_DIR;
+  originalConfigDir = process.env[CONFIG_DIR_ENV_VAR];
   testConfigDir = join(
-    process.env.SENTRY_CLI_CONFIG_DIR || "/tmp",
+    process.env[CONFIG_DIR_ENV_VAR] || "/tmp",
```
BYK: This pattern seems to be repeating. It also has the following problems:

1. It does not isolate tests at all
2. It assumes `/tmp` is writeable and is the place for temp files, ignoring system env variables such as TMP_DIR etc and would 100% not work on Windows.

BYK: Addressed in 6e974ae - created `test/helpers.ts` with `createTestConfigDir()` that uses `fs.promises.mkdtemp(join(tmpdir(), prefix))` for proper cross-platform temp directory creation and test isolation.

PR #53, opened 2026-01-23
feat(commands): rename get commands to view and add -w browser flag

BYK:
## Summary

- Rename all `get` commands to `view` for better UX
- Add `-w` / `--web` flag to open entities in browser with QR code fallback

### Commands Renamed
| Old | New |
|-----|-----|
| `sentry org get` | `sentry org view` |
| `sentry project get` | `sentry project view` |
| `sentry issue get` | `sentry issue view` |
| `sentry event get` | `sentry event view` |

### New `-w` / `--web` Flag
All view commands now support a `-w` flag that opens the entity in the browser:
- Opens browser if available
- Falls back to displaying URL + QR code (same pattern as `sentry auth login`)

### New Utilities
- `src/lib/sentry-urls.ts` - URL builders for Sentry web pages (supports self-hosted via `SENTRY_URL` env var)
- `openOrShowUrl()` helper in `src/lib/browser.ts`

### DIFF
--- a/test/e2e/event.test.ts
+++ b/test/e2e/event.test.ts
@@ -30,10 +30,10 @@ afterEach(async () => {
   await cleanupTestDir(testConfigDir);
 });
 
-describe("sentry event get", () => {
+describe("sentry event view", () => {
   test("requires authentication", async () => {
     const result = await runCli(
-      ["event", "get", "abc123", "--org", TEST_ORG, "--project", TEST_PROJECT],
+      ["event", "view", "abc123", "--org", TEST_ORG, "--project", TEST_PROJECT],
       { env: { [CONFIG_DIR_ENV_VAR]: testConfigDir } }
     );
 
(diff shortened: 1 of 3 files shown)


src/lib/browser.ts
```diff
@@ -59,3 +62,54 @@ export async function openBrowser(url: string): Promise<boolean> {
...
+ *
+ * @param stdout - Output stream for messages
+ * @param url - The URL to open or display
+ * @returns true if browser opened, false if showing fallback
+ */
+export async function openOrShowUrl(
+  stdout: Writer,
+  url: string
+): Promise<boolean> {
+  const opened = await openBrowser(url);
```
BYK, line 80: Should we wrap this in a `try-catch` block? I remember having failures when `xdg-open` was not available etc.

src/lib/browser.ts
```diff
@@ -59,3 +62,54 @@ export async function openBrowser(url: string): Promise<boolean> {
...
+ * Open URL in browser, or show URL + QR code as fallback.
+ *
+ * Attempts to open the browser first. If that fails (no browser available,
+ * headless environment, etc.), displays the URL and a QR code for mobile scanning.
+ *
+ * @param stdout - Output stream for messages
+ * @param url - The URL to open or display
+ * @returns true if browser opened, false if showing fallback
+ */
+export async function openOrShowUrl(
```
BYK, line 76: I think we should use this helper in `auth login` for device login flow too.

PR #69, opened 2026-01-26
feat: Added CLI Skill

MathurAditya724:
Closes #55

PR #105, opened 2026-01-29
feat: Auto-generate SKILL.md and extract version bump script

BYK:
## Summary

Addresses two issues from #69:
1. The `.craft.yml` version bump scripts were unwieldy inline Node.js code
2. The SKILL.md file was manually maintained and could become stale

### Version Bump Script (`script/bump-version.ts`)
- Extracts the inline version bump logic from `.craft.yml` into a standalone Bun script
- Handles both pre-release (`--pre`) and post-release (`--post`) scenarios
- Keeps `package.json` and `plugin.json` versions in sync

### Skill Generation (`script/generate-skill.ts`)
- Generates `SKILL.md` by introspecting the Stricli route tree
- Extracts command paths, flags, aliases, and descriptions automatically
- Produces structured documentation with prerequisites, commands, workflows, and error resolution

### CI Integration

**Check Job** (`check-skill` in `ci.yml`):
- Runs on every PR to detect if `SKILL.md` is stale
- On failure, outputs instructions to the job summary with two fix options

**Generate Workflow** (`.github/workflows/generate-skill.yml`):
- Manual `workflow_dispatch` trigger with branch input
- For internal PRs, the check job links directly to this workflow with the branch pre-filled
- Regenerates and commits SKILL.md automatically
- External contributors (fork PRs) only see local command instructions since they can't trigger workflows

### Cursor Support
- Added symlink at `.cursor/skills/sentry-cli/SKILL.md` pointing to the plugins location

## New Scripts

```bash
bun run generate:skill  # Generate SKILL.md from CLI metadata
bun run check:skill     # Check if SKILL.md is up to date
```

### DIFF
--- a/.github/workflows/generate-skill.yml
+++ b/.github/workflows/generate-skill.yml
@@ -19,7 +19,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.branch }}
-          # Use default GITHUB_TOKEN - commits will be attributed to the user who triggered the workflow
 
       - uses: oven-sh/setup-bun@v2
 
@@ -49,6 +48,8 @@ jobs:
       - name: Commit and push
         if: steps.diff.outputs.changed == 'true'
         run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
           git add plugins/sentry-cli/skills/sentry-cli/SKILL.md
           git commit -m "chore: Regenerate SKILL.md"
           git push
--- a/plugins/sentry-cli/skills/sentry-cli/SKILL.md
+++ b/plugins/sentry-cli/skills/sentry-cli/SKILL.md
@@ -14,10 +14,9 @@ The CLI must be installed and authenticated before use.
 ### Installation
 
 ```bash
-# Install script
 curl https://cli.sentry.dev/install -fsS | bash
 
-# Or use npm/pnpm/bun
+# Or install via npm/pnpm/bun
 npm install -g sentry
 ```
 
--- a/script/generate-skill.ts
+++ b/script/generate-skill.ts
@@ -23,6 +23,9 @@ const FRONTMATTER_REGEX = /^---\n[\s\S]*?\n---\n/;
 /** Regex to match code blocks with optional language specifier */
 const CODE_BLOCK_REGEX = /```(\w*)\n([\s\S]*?)```/g;
 
+/** Regex to extract npm command from PackageManagerCode Astro component (handles multi-line) */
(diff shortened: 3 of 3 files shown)


.github/workflows/generate-skill.yml
```diff
@@ -0,0 +1,56 @@
...
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+            echo "SKILL.md has changes"
+          fi
+
+      - name: Commit and push
+        if: steps.diff.outputs.changed == 'true'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
BYK: Let's not set these manually
```suggestion
```

plugins/sentry-cli/skills/sentry-cli/SKILL.md
```diff
@@ -42,89 +42,159 @@ sentry auth status
...
-sentry project list --org my-org           # List projects in specific org
-sentry project list --platform javascript  # Filter by platform
-sentry project view my-project             # View project details
-sentry project view my-project -w          # Open in browser
-```
+Authenticate with Sentry
 
-### Issues
+**Flags:**
+- `--token <value> - Authenticate using an API token instead of OAuth`
```
BYK, line 43: We should find a way to generate these examples back. I'm okay if those examples live in the original help text as that's also very useful.

script/generate-skill.ts
```diff
@@ -0,0 +1,550 @@
...
+
+function isRouteMap(target: RouteTarget): target is RouteMap {
+  return "getAllEntries" in target;
+}
+
+function isCommand(target: RouteTarget): target is Command {
+  return "parameters" in target && !("getAllEntries" in target);
+}
+
+type CommandInfo = {
```
BYK, line 409: Are these types/interfaces not available from stricli?

script/generate-skill.ts
```diff
@@ -0,0 +1,550 @@
...
+
+/**
+ * Generate the prerequisites section
+ */
+function generatePrerequisites(): string {
+  return `## Prerequisites
+
+The CLI must be installed and authenticated before use.
+
+### Installation
```
BYK: Ideally these would come from `/docs`

script/generate-skill.ts
```diff
@@ -0,0 +1,550 @@
...
+sentry issue list --query "is:unresolved" --sort date
+
+# View issue details
+sentry issue view PROJ-ABC
+
+# Get AI root cause analysis
+sentry issue explain PROJ-ABC
+
+# Open in browser for full context
+sentry issue view PROJ-ABC -w
```
BYK: We should not have these hard-coded in the script, it defeats the purpose of auto-generation!

PR #127, opened 2026-01-30
feat(telemetry): improve Sentry instrumentation

BYK:
## Summary

Improves Sentry telemetry for better observability and debugging of CLI usage.

## Changes

- **Better tracing**: Add spans for HTTP requests, database operations, and serialization
- **User context**: Track authenticated user ID and CLI instance ID (no PII)
- **Command context**: Tag spans with command name, org/project when available
- **Auth feedback**: Show username after successful login
- **Faster exit**: Patch `@sentry/core` to prevent 3s flush timeout from blocking CLI exit

## Technical Notes

- Uses `@sentry/bun` for native builds, aliased to `@sentry/node` for npm bundle
- Adds `user_info` and `instance_info` tables (schema v2 migration)
- UUIDv7 for instance IDs via `Bun.randomUUIDv7()` (polyfilled for Node.js)

### DIFF
--- a/src/commands/auth/login.ts
+++ b/src/commands/auth/login.ts
@@ -81,15 +81,10 @@ export const loginCommand = buildCommand({
       // Save token first, then validate by fetching user info
       await setAuthToken(flags.token);
 
-      // Validate token and fetch user info in one call
+      // Validate token by fetching user info
       let user: SentryUser;
       try {
         user = await getCurrentUser();
-        setUserInfo({
-          userId: user.id,
-          email: user.email,
-          username: user.username,
-        });
       } catch {
         // Token is invalid - clear it and throw
         await clearAuth();
@@ -97,6 +92,18 @@ export const loginCommand = buildCommand({
           "invalid",
           "Invalid API token. Please check your token and try again."
         );
+      }
+
+      // Store user info for telemetry (non-critical, don't block auth)
+      try {
+        setUserInfo({
+          userId: user.id,
+          email: user.email,
+          username: user.username,
+        });
+      } catch (error) {
+        // Report to Sentry but don't block auth - user info is not critical
+        Sentry.captureException(error);
       }
 
       stdout.write(`${success("✓")} Authenticated with API token\n`);


src/commands/auth/login.ts
```diff
@@ -1,16 +1,47 @@
...
 import { AuthError } from "../../lib/errors.js";
 import { muted, success } from "../../lib/formatters/colors.js";
 import { formatDuration } from "../../lib/formatters/human.js";
 import { completeOAuthFlow, performDeviceFlow } from "../../lib/oauth.js";
 import { generateQRCode } from "../../lib/qrcode.js";
 
+/** Maximum number of retries for fetching user info */
+const USER_FETCH_MAX_RETRIES = 3;
+/** Delay between retries in milliseconds */
+const USER_FETCH_RETRY_DELAY = 1000;
```
betegon: it could make sense to use `ky` for this as we're already using it and should have the features like retries: https://github.com/sindresorhus/ky

see api-client.ts

src/commands/auth/login.ts
```diff
@@ -1,16 +1,35 @@
...
+async function fetchAndStoreUserInfo(): Promise<void> {
+  try {
+    const user = await getCurrentUser();
+    setUserInfo({
+      userId: user.id,
+      email: user.email,
+      username: user.username,
+    });
+  } catch {
+    // Silently continue - user info is not critical for auth
```
BYK: At least report this to Sentry with error details.

src/commands/auth/login.ts
BYK: I think we can use `fetchAdnStoreUserInfo()` here for validation now?

src/commands/auth/logout.ts
```diff
@@ -28,6 +29,7 @@ export const logoutCommand = buildCommand({
     }
 
     await clearAuth();
+    await clearUserInfo();
```
BYK: This should be folded into `clearAuth()`

src/commands/issue/list.ts
```diff
@@ -319,6 +319,18 @@ export const listCommand = buildCommand({
...
+    // Set telemetry context (single project mode gets both org and project)
+    const telemetryTarget = targets[0];
+    if (targets.length === 1 && telemetryTarget) {
+      setContext(telemetryTarget.org, telemetryTarget.project);
+    } else if (targets.length > 1 && telemetryTarget) {
+      // Multi-project: set org if all targets share the same org
+      const orgs = new Set(targets.map((t) => t.org));
+      if (orgs.size === 1) {
+        setContext(telemetryTarget.org);
+      }
```
BYK: Just make org and project arrays at all times to simplify this.

src/lib/db/instance.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Uses Bun's native randomUUIDv7 if available (Bun 1.0.30+),
+ * otherwise falls back to a simple implementation for Node.js (npm bundle).
+ */
+function generateUUIDv7(): string {
+  // Bun 1.0.30+ has native UUIDv7 support
+  if (typeof Bun !== "undefined" && typeof Bun.randomUUIDv7 === "function") {
+    return Bun.randomUUIDv7();
+  }
+
+  // Fallback implementation for Node.js (npm bundle)
```
BYK: There must be an existing package on npm for this. Also check Node API docs to see if it was added in newer versions of node (22+)

src/lib/db/schema.ts
```diff
@@ -99,7 +115,28 @@ function getSchemaVersion(db: Database): number {
 export function runMigrations(db: Database): void {
   const currentVersion = getSchemaVersion(db);
 
-  // Add migrations here as schema evolves
+  // Migration from v1 to v2: Add user_info and instance_info tables
+  if (currentVersion < 2) {
```
BYK, line 126: This looks like the repetition of the things above?

src/lib/telemetry.ts
```diff
@@ -10,35 +10,79 @@
...
+    if (user) {
+      // Only send user ID - email/username are PII
+      Sentry.setUser({ id: user.userId });
+    }
+
+    if (instanceId) {
+      Sentry.setTag("instance_id", instanceId);
+    }
+  } catch {
+    // Context initialization is not critical - continue without it
```
BYK, line 43: We should capture this error to Sentry too.

src/lib/oauth.ts
```diff
@@ -328,65 +333,68 @@ export async function refreshAccessToken(
     );
   }
 
-  let response: Response;
+  // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: OAuth error handling requires branching
+  return withHttpSpan("POST", "/oauth/token/", async () => {
```
BYK, line 320: Will this `withHttpSpan` wrapper interfere with our existing `fetch` instrumentation?

src/lib/oauth.ts
```diff
@@ -328,65 +333,68 @@ export async function refreshAccessToken(
...
-
-  if (!response.ok) {
-    let errorDetail = "Token refresh failed";
     try {
-      const errorData = await response.json();
-      const errorResult = TokenErrorResponseSchema.safeParse(errorData);
-      if (errorResult.success) {
-        errorDetail =
-          errorResult.data.error_description ?? errorResult.data.error;
+      response = await fetch(`${SENTRY_URL}/oauth/token/`, {
```
BYK: Why are we not using `ky` here?

src/lib/oauth.ts
```diff
@@ -73,105 +74,109 @@ function sleep(ms: number): Promise<void> {
...
+          `Cannot connect to Sentry at ${SENTRY_URL}`,
+          0,
+          "Check your network connection and SENTRY_URL configuration"
+        );
+      }
+      throw error;
+    }
 
-    if (isConnectionError) {
+    if (!response.ok) {
```
BYK: This pattern seems repeated, can we abstract it out?

src/lib/db/schema.ts
```diff
@@ -4,7 +4,35 @@
 
 import type { Database } from "bun:sqlite";
 
-const CURRENT_SCHEMA_VERSION = 1;
+const CURRENT_SCHEMA_VERSION = 2;
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Table Definitions (single source of truth)
+// ─────────────────────────────────────────────────────────────────────────────
```
BYK: This comment is useless/redundant.

src/lib/db/schema.ts
```diff
@@ -4,7 +4,35 @@
...
+/** Instance identifier for telemetry (single row, id=1) */
+const INSTANCE_INFO_TABLE = `
+  CREATE TABLE IF NOT EXISTS instance_info (
+    id INTEGER PRIMARY KEY CHECK (id = 1),
+    instance_id TEXT NOT NULL,
+    created_at INTEGER NOT NULL DEFAULT (unixepoch() * 1000)
+  )
+`;
+
+// ─────────────────────────────────────────────────────────────────────────────
```
BYK: Redundant.

src/lib/db/schema.ts
```diff
@@ -89,6 +120,10 @@ export function initSchema(db: Database): void {
   }
 }
 
+// ─────────────────────────────────────────────────────────────────────────────
```
BYK: Find all rendundant comments like these and remove them all.

src/lib/db/instance.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (existingRow) {
+    return existingRow.instance_id;
+  }
+
+  // Generate and store new instance ID
+  // Use INSERT OR IGNORE to handle race condition when multiple CLI processes
+  // start simultaneously - only the first insert succeeds
+  // Bun.randomUUIDv7() is native in Bun, polyfilled via uuidv7 package for Node.js
+  const instanceId = Bun.randomUUIDv7();
```
BYK, line 32: Wait, I wonder if SQLite itself has a built-in `uuidv7` function?

BYK:
## Addressing PR Review Comments

I've made the following code changes in commit 912781f:

1. **Add Sentry.captureException in telemetry.ts** - Added error capture in the `initTelemetryContext()` catch block
2. **Abstract connection error handling** - Created `fetchWithConnectionError()` helper in oauth.ts to DRY up the repeated pattern
3. **Remove redundant divider comments** - Cleaned up `// ───...` comment blocks in schema.ts and oauth.ts

---

### Responses to Questions:

**Q: Will `withHttpSpan` wrapper interfere with existing `fetch` instrumentation?**

No, they complement each other:
- `httpIntegration()` auto-instruments all fetch calls globally
- `withHttpSpan` adds **explicit spans** for OAuth calls that need custom error handling and aren't using `ky`
- The spans will nest properly - the http integration creates the outer span, withHttpSpan creates a more specifically-named inner span

**Q: Why not using `ky` in oauth.ts?**

OAuth uses native fetch because:
- Device flow has specific polling requirements (interval, slow_down handling)
- Different error response schemas (TokenErrorResponse vs standard API errors)
- Needs fine-grained control over retry behavior (polling intervals from server)
- The error handling is more complex than standard API calls (multiple response schemas to parse)

**Q: Does SQLite have a built-in `uuidv7` function?**

No, SQLite doesn't have built-in UUIDv7 support. Standard SQLite only provides:
- `randomblob(16)` for random bytes
- `hex()` for hex encoding

UUIDv7 requires specific timestamp encoding in the first 48 bits that SQLite doesn't provide natively. Using `Bun.randomUUIDv7()` (with `uuidv7` npm polyfill for Node.js) is the correct approach.

---

### Already Addressed in Previous Commits:

- ✅ Use ky for login.ts retries → Retry logic removed; uses `getCurrentUser()` which uses ky
- ✅ Report errors to Sentry in catch block → Added `Sentry.captureException(error)` in OAuth flow
- ✅ Use fetchAndStoreUserInfo for validation → Refactored to use `getCurrentUser()` directly
- ✅ Fold clearUserInfo into clearAuth → `clearAuth()` now deletes from user_info table
- ✅ Make org/project arrays → `setOrgProjectContext()` accepts string arrays
- ✅ Use npm package for UUIDv7 → Uses `Bun.randomUUIDv7()` with uuidv7 polyfill
- ✅ Schema repetition → Table definitions extracted to constants, used in both init and migrations

src/commands/auth/login.ts
```diff
@@ -70,6 +80,9 @@ export const loginCommand = buildCommand({
       }
 
       stdout.write(`${success("✓")} Authenticated with API token\n`);
+      stdout.write(
+        `  Logged in as ${muted(`${user.username} <${user.email}>`)}\n`
+      );
```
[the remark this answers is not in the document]
BYK: Fixed in 1a7d68d - Added `formatUserIdentity()` helper that gracefully handles missing username/email:
- Both present: `username <email>`
- Only username: `username`
- Only email: `email`
- Neither: `user {id}`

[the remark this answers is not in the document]
BYK: Fixed in 1a7d68d - Same fix as above. Added `formatUserIdentity()` helper to handle optional username/email fields gracefully.

PR #151, opened 2026-02-01
feat: add update available notification

BYK:
## Summary

Adds a "new version available" notification that displays after CLI commands complete, notifying users when a newer version is available on GitHub releases.

## Key Features

- **Background check**: Probabilistic check (~once per day) that doesn't block CLI execution
- **Cached results**: Version info stored in SQLite metadata table for instant notifications on subsequent runs
- **Smart suppression**: Notifications suppressed for `upgrade`, `--version`, `-V`, and `--json` flags
- **Opt-out**: `SENTRY_CLI_NO_UPDATE_CHECK=1` environment variable to disable entirely
- **Observability**: Errors reported to Sentry via detached span
- **Node.js support**: Polyfill for `Bun.semver` using the `semver` package

## Example Output

```
$ sentry issue list
...command output...

Update available: 0.5.0 → 0.6.0  Run "sentry upgrade" to update.
```

### DIFF
--- a/script/node-polyfills.ts
+++ b/script/node-polyfills.ts
@@ -5,7 +5,7 @@ import { execSync, spawn as nodeSpawn } from "node:child_process";
 import { access, readFile, writeFile } from "node:fs/promises";
 import { DatabaseSync } from "node:sqlite";
 
-import { compare as semverCompare, satisfies as semverSatisfies } from "semver";
+import { compare as semverCompare } from "semver";
 import { glob } from "tinyglobby";
 import { uuidv7 } from "uuidv7";
 
@@ -167,7 +167,6 @@ const BunPolyfill = {
 
   semver: {
     order: semverCompare,
-    satisfies: semverSatisfies,
   },
 };
 
--- a/src/lib/version-check.ts
+++ b/src/lib/version-check.ts
@@ -73,9 +73,16 @@ export function abortPendingVersionCheck(): void {
  * Start a background check for new versions.
  * Does not block - fires a fetch and lets it complete in the background.
  * Reports errors to Sentry in a detached span for visibility.
+ * Never throws - errors are caught and reported to Sentry.
  */
 function checkForUpdateInBackgroundImpl(): void {
-  if (!shouldCheckForUpdate()) {
+  try {
+    if (!shouldCheckForUpdate()) {
+      return;
+    }
+  } catch (error) {
+    // DB access failed - report to Sentry but don't crash CLI
+    Sentry.captureException(error);
     return;
   }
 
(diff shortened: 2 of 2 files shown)


script/node-polyfills.ts
```diff
@@ -163,6 +164,15 @@ const BunPolyfill = {
   randomUUIDv7(): string {
     return uuidv7();
   },
+
+  semver: {
+    order(a: string, b: string): -1 | 0 | 1 {
+      return semverCompare(a, b);
+    },
```
BYK:
```suggestion
    order: semverCompare,
```

script/node-polyfills.ts
```diff
@@ -163,6 +164,15 @@ const BunPolyfill = {
...
+    satisfies(version: string, range: string): boolean {
+      return semverSatisfies(version, range);
+    },
```
BYK:
```suggestion
    satisfies: semverSatisfies,
```

src/lib/version-check.ts
```diff
@@ -0,0 +1,134 @@
...
+  const probability = 1 - Math.exp(-elapsed / effectiveInterval);
+
+  return Math.random() < probability;
+}
+
+/**
+ * Check if update notifications should be suppressed for these args.
+ */
+export function shouldSuppressNotification(args: string[]): boolean {
+  return args.some((arg) => SUPPRESSED_ARGS.has(arg) || arg === "--json");
```
BYK: Why not add `--json` to `SUPPRESSED_ARGS` too?

src/lib/version-check.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Start a background check for new versions.
+ * Does not block - fires a fetch and lets it complete in the background.
+ * Reports errors to Sentry in a detached span for visibility.
+ */
+function checkForUpdateInBackgroundImpl(): void {
+  const { lastChecked } = getVersionCheckInfo();
+
+  if (!shouldCheckForUpdate(lastChecked)) {
+    return;
+  }
```
BYK: I'd bake the `getVersionCheckInfo()` logic into `shouldCheckForUpdate()` so we just call that and it does whatever it needs to to come up with a response.

src/lib/version-check.ts
```diff
@@ -0,0 +1,134 @@
...
+          span.setStatus({ code: 1 }); // OK
+        })
+        .catch((error) => {
+          Sentry.captureException(error);
+          span.setStatus({ code: 2 }); // Error
+        })
+        .finally(() => {
+          span.end();
+        });
+    }
```
BYK: Why not use `async`/`await` along with a traditional `try`/`catch`/`finally` construct here?

src/lib/version-check.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  // Start a detached span for the background update check
+  Sentry.startSpanManual(
+    {
+      name: "version-check",
+      op: "version.check",
+      forceTransaction: true,
+    },
+    (span) => {
+      fetchLatestFromGitHub()
```
BYK: I think we should `.unref()` this to not make it block program exit?

PR #154, opened 2026-02-02
refactor: move feedback and upgrade under `sentry cli` command

BYK:
## Summary

Group CLI meta-commands under a new `cli` subcommand to avoid collision with the existing Sentry Feedback feature (user feedback on issues).

## Changes

**Commands:**
- `sentry feedback` → `sentry cli feedback`
- `sentry upgrade` → `sentry cli upgrade`
- Removed `sentry update` alias

**Files:**
- Created `src/commands/cli/` with `index.ts`, `feedback.ts`, `upgrade.ts`
- Updated `src/app.ts` to use `cliRoute`
- Updated version check notification message
- Updated all documentation (`commands/cli/`, SKILL.md files)

## Breaking Changes

Users will need to use `sentry cli feedback` and `sentry cli upgrade` instead of the old top-level commands.

### DIFF
--- a/test/commands/cli.test.ts
+++ b/test/commands/cli.test.ts
@@ -6,28 +6,8 @@
 
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { feedbackCommand } from "../../src/commands/cli/feedback.js";
-import { cliRoute } from "../../src/commands/cli/index.js";
 import { upgradeCommand } from "../../src/commands/cli/upgrade.js";
 
-describe("cliRoute", () => {
-  test("is exported and defined", () => {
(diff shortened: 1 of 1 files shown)


test/commands/cli.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  test("is exported and defined", () => {
+    expect(feedbackCommand).toBeDefined();
+  });
+});
+
+describe("upgradeCommand", () => {
+  test("is exported and defined", () => {
+    expect(upgradeCommand).toBeDefined();
+  });
+});
```
BYK: These are silly tests. Use e2e tests for these.

PR #160, opened 2026-02-03
feat(commands): add sentry log command

betegon:
## Summary

Adds `sentry log list` command to list and stream logs from Sentry projects. Supports real-time streaming with `--follow` flag using efficient timestamp-based polling.

## Usage

```bash
sentry log list                    # auto-detect from DSN
sentry log list <org>/<project>    # explicit target
sentry log list <project>          # find project across orgs
sentry log list -f                 # stream in real-time
sentry log list --tail 50 -q 'level:error'
```

## Changes

- New `src/commands/log/` with list command
- Log formatters with severity coloring and timestamp formatting
- API client method `listLogs()` with `afterTimestamp` support for follow mode
- Follows existing patterns from `issue list` (parseOrgProjectArg, writeFooter, etc.)

## Test Plan

- `bun test test/lib/formatters/log.test.ts` - formatter unit tests
- Manual testing with real Sentry project using all target modes

### DIFF
--- a/src/commands/issue/utils.ts
+++ b/src/commands/issue/utils.ts
@@ -15,13 +15,10 @@ import { getProjectByAlias } from "../../lib/db/project-aliases.js";
 import { detectAllDsns } from "../../lib/dsn/index.js";
 import { ContextError } from "../../lib/errors.js";
 import { getProgressMessage } from "../../lib/formatters/seer.js";
-import {
-  expandToFullShortId,
-  isNumericId,
-  isShortSuffix,
-} from "../../lib/issue-id.js";
+import { expandToFullShortId, isShortSuffix } from "../../lib/issue-id.js";
 import { poll } from "../../lib/polling.js";
 import { resolveOrgAndProject } from "../../lib/resolve-target.js";
+import { isAllDigits } from "../../lib/utils.js";
 import type { SentryIssue, Writer } from "../../types/index.js";
 import { type AutofixState, isTerminalStatus } from "../../types/seer.js";
 
@@ -51,7 +48,7 @@ export const issueIdPositional = {
  */
 export function buildCommandHint(command: string, issueId: string): string {
   // Numeric IDs always need org context - can't be combined with project
-  if (isNumericId(issueId)) {
+  if (isAllDigits(issueId)) {
     return `sentry issue ${command} <org>/${issueId}`;
   }
   // Short suffixes can be combined with project prefix
--- a/src/lib/arg-parsing.ts
+++ b/src/lib/arg-parsing.ts
@@ -6,7 +6,7 @@
  * project list) and single-item commands (issue view, explain, plan).
  */
 
-import { isNumericId } from "./issue-id.js";
+import { isAllDigits } from "./utils.js";
 
 /** Default span depth when no value is provided */
 const DEFAULT_SPAN_DEPTH = 3;
@@ -205,7 +205,7 @@ function parseAfterSlash(
   org: string,
   rest: string
 ): ParsedIssueArg {
-  if (isNumericId(rest)) {
+  if (isAllDigits(rest)) {
     // "my-org/123456789" → explicit org + numeric ID
     return { type: "explicit-org-numeric", org, numericId: rest };
   }
@@ -288,7 +288,7 @@ function parseWithDash(arg: string): ParsedIssueArg {
 
 export function parseIssueArg(arg: string): ParsedIssueArg {
   // 1. Pure numeric → direct fetch by ID
-  if (isNumericId(arg)) {
+  if (isAllDigits(arg)) {
     return { type: "numeric", id: arg };
   }
 
--- a/src/lib/issue-id.ts
+++ b/src/lib/issue-id.ts
@@ -20,9 +20,6 @@ const SHORT_SUFFIX_PATTERN = /^[a-zA-Z0-9]+$/;
 /** Pattern for alias-suffix format (e.g., "f-g", "fr-a3", "spotlight-e-4y") */
 const ALIAS_SUFFIX_PATTERN = /^(.+)-([a-zA-Z0-9]+)$/i;
 
-/** Pattern for pure numeric IDs (all digits) */
-const NUMERIC_ID_PATTERN = /^\d+$/;
-
 /**
  * Check if a string looks like a short ID (e.g., PROJECT-ABC)
  * vs a numeric ID (e.g., 123456).
@@ -76,15 +73,4 @@ export function expandToFullShortId(
   projectSlug: string
 ): string {
   return `${projectSlug.toUpperCase()}-${suffix.toUpperCase()}`;
-}
-
-/**
- * Check if a string is a pure numeric ID (all digits).
- *
- * Used to distinguish between:
- * - Numeric issue IDs: "123456789" → need org context
- * - Short suffixes: "G", "A3" → can be combined with project
- */
-export function isNumericId(input: string): boolean {
-  return NUMERIC_ID_PATTERN.test(input);
 }
--- a/test/lib/issue-id.property.test.ts
+++ b/test/lib/issue-id.property.test.ts
@@ -17,11 +17,11 @@ import {
 } from "fast-check";
 import {
   expandToFullShortId,
-  isNumericId,
   isShortId,
   isShortSuffix,
   parseAliasSuffix,
 } from "../../src/lib/issue-id.js";
+import { isAllDigits } from "../../src/lib/utils.js";
 import { DEFAULT_NUM_RUNS } from "../model-based/helpers.js";
 
 // Arbitraries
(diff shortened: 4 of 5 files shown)


src/types/sentry.ts
```diff
@@ -678,3 +678,55 @@ export const ProjectKeySchema = z
...
+ * Fields match the Sentry Explore/Events API response for dataset=logs.
+ */
+export const SentryLogSchema = z
+  .object({
+    /** Unique identifier for deduplication */
+    "sentry.item_id": z.string(),
+    /** ISO timestamp of the log entry */
+    timestamp: z.string(),
+    /** Nanosecond-precision timestamp for accurate ordering and filtering */
+    timestamp_precise: z.number(),
```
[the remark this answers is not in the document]
betegon: this would be a risk of: 1 in ~2,000,000 polls could have duplicate/missed log

plugins/sentry-cli/skills/sentry-cli/SKILL.md
```diff
@@ -403,6 +403,21 @@ Update the Sentry CLI to the latest version
...
+### Log
+
+View Sentry logs
+
+#### `sentry log list <target>`
+
+List logs from a project
+
+**Flags:**
+- `-n, --tail <value> - Number of log entries (1-1000) - (default: "100")`
```
BYK: `-n` stands for `--lines` in `--tail`. I don't think `--tail` is a great name.

>-n, --lines=[+]NUM

```suggestion
- `-n, --limit <value> - Number of log entries (1-1000) - (default: "100")`
```

plugins/sentry-cli/skills/sentry-cli/SKILL.md
```diff
@@ -403,6 +403,21 @@ Update the Sentry CLI to the latest version
...
+- `-q, --query <value> - Filter query (Sentry search syntax)`
+- `-f, --follow - Stream logs in real-time`
+- `--pollInterval <value> - Poll interval in seconds (only with --follow) - (default: "2")`
```
BYK: Why not merge this into `-f` as an optional value flag:

If just `-f`, poll with default interval, `-f 10` poll with 10 (seconds?)

src/commands/log/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  docs: {
+    brief: "View Sentry logs",
+    fullDescription:
+      "View and stream logs from your Sentry projects.\n\n" +
+      "Commands:\n" +
+      "  list     List or stream logs from a project\n\n" +
+      "Examples:\n" +
+      "  sentry log list                    # Auto-detect from DSN\n" +
+      "  sentry log list myorg/myproject    # Explicit org/project\n" +
+      "  sentry log list -f                 # Stream logs in real-time\n" +
```
BYK: I thought the SKILL file would get these docs?

src/lib/api-client.ts
```diff
@@ -943,3 +946,81 @@ export function getCurrentUser(): Promise<SentryUser> {
     schema: SentryUserSchema,
   });
 }
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Logs
+// ─────────────────────────────────────────────────────────────────────────────
```
BYK: We should remove these kinds of divider comments, distracting.

src/lib/api-client.ts
```diff
@@ -943,3 +946,81 @@ export function getCurrentUser(): Promise<SentryUser> {
...
+  "sentry.item_id",
+  "trace",
+  "severity",
+  "timestamp",
+  "timestamp_precise",
+  "message",
+];
+
+/** Regex to check if a project identifier is a numeric ID */
+const NUMERIC_PROJECT_REGEX = /^\d+$/;
```
BYK: Pretty sure we have the exact same regex somewhere in the project. We should just name it `isAllDigits` or something and use that everywhere

src/lib/api-client.ts
```diff
@@ -943,3 +946,81 @@ export function getCurrentUser(): Promise<SentryUser> {
...
+export async function listLogs(
+  orgSlug: string,
+  projectSlug: string,
+  options: ListLogsOptions = {}
+): Promise<SentryLog[]> {
+  // API only accepts numeric project IDs as param, slugs go in query
+  const isNumericProject = NUMERIC_PROJECT_REGEX.test(projectSlug);
+
+  // Build query parts
+  const projectFilter = isNumericProject ? "" : `project:${projectSlug}`;
```
BYK, line 1044: Instead of this, can we not infer the numeric id of the project by querying the projects endpoint or something?

src/lib/api-client.ts
```diff
@@ -943,3 +946,81 @@ export function getCurrentUser(): Promise<SentryUser> {
...
+
+  const response = await orgScopedRequest<LogsResponse>(
+    `/organizations/${orgSlug}/events/`,
+    {
+      params: {
+        dataset: "logs",
+        field: LOG_FIELDS,
+        project: isNumericProject ? projectSlug : undefined,
+        query: fullQuery || undefined,
+        per_page: options.limit ?? 100,
```
BYK: Well, we probably shouldn't accept `0` as a limit right?
```suggestion
        per_page: options.limit || 100,
```

src/lib/api-client.ts
```diff
@@ -943,3 +946,81 @@ export function getCurrentUser(): Promise<SentryUser> {
...
+        statsPeriod: options.statsPeriod ?? "90d",
```
BYK: I think 90d is the max and I'd go with lower. Maybe even `1d` or `7d`?
```suggestion
        statsPeriod: options.statsPeriod ?? "30d",
```

src/lib/formatters/log.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // Format as local time: YYYY-MM-DD HH:MM:SS
+  const year = date.getFullYear();
+  const month = String(date.getMonth() + 1).padStart(2, "0");
+  const day = String(date.getDate()).padStart(2, "0");
+  const hours = String(date.getHours()).padStart(2, "0");
+  const minutes = String(date.getMinutes()).padStart(2, "0");
+  const seconds = String(date.getSeconds()).padStart(2, "0");
+  return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`;
```
BYK: Isn't this very close to ISO8601? Can we not use [`.toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) here?

src/lib/api-client.ts
betegon: i don't know why my cleanup skill doesn't filter these

src/lib/api-client.ts
betegon: I see your point, but it's an extra API call that has to happen before executing this request.
i'd leave this as it is, works and doesn't add latency.

src/lib/formatters/log.ts
betegon: this is to maintain the local timezone instead of showing the `UTC` one.

betegon: that said, looks like the Swedish locale `sv-SE` maintains the local timezone.

src/lib/api-client.ts
betegon: you're right, put it on 7d, should be good.

src/commands/log/list.ts
```diff
@@ -0,0 +1,390 @@
...
+  const logs = await listLogs(org, project, {
+    query: flags.query,
+    limit: flags.limit,
+    statsPeriod: "90d",
+  });
+
+  if (flags.json) {
+    // Reverse for chronological order (API returns newest first)
+    writeJson(stdout, [...logs].reverse());
+    return;
```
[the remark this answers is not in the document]
betegon: this is actually intended. `--json` give us normal json for every command but for streaming where it's ndjson

src/lib/formatters/log.ts
```diff
@@ -47,14 +47,8 @@ function formatTimestamp(timestamp: string): string {
...
 
-  // Format as local time: YYYY-MM-DD HH:MM:SS
-  const year = date.getFullYear();
-  const month = String(date.getMonth() + 1).padStart(2, "0");
-  const day = String(date.getDate()).padStart(2, "0");
-  const hours = String(date.getHours()).padStart(2, "0");
-  const minutes = String(date.getMinutes()).padStart(2, "0");
-  const seconds = String(date.getSeconds()).padStart(2, "0");
-  return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`;
+  // Swedish locale naturally uses ISO 8601 format (YYYY-MM-DD HH:MM:SS) in local time
```
BYK, line 50: 🤣

betegon: i had to ask twice about this 😆

PR #155, opened 2026-02-02
feat(commands): use positional args for org/project selection

BYK:
## Summary

Replace verbose `--org` and `--project` flags with GitHub-style positional arguments for `issue list` and `project list` commands, inspired by the `gh` CLI.

**Before:**
```bash
sentry issue list --org sentry --project cli
sentry project list --org sentry
```

**After:**
```bash
sentry issue list sentry/cli     # explicit org and project
sentry issue list sentry/        # all projects in org  
sentry issue list cli            # find project across all orgs
sentry project list sentry       # list projects in org
```

## Changes

- Add `parseOrgProjectArg()` utility for parsing `<org>/<project>` patterns
- Add `findProjectsBySlug()` API function to search projects across orgs
- Refactor `issue list` to use positional argument instead of flags
- Refactor `project list` to use positional argument instead of flag
- Change cross-org collision alias separator from `:` to `/` (e.g., `o1/d` instead of `o1:d`)
- Update help text and error messages
- Add tests for new functionality

## Breaking Changes

- `--org` and `--project` flags removed from `sentry issue list`
- `--org` flag removed from `sentry project list`

PR #161, opened 2026-02-03
feat(issue): replace --org/--project flags with <org>/ID syntax

BYK:
## Summary

Replace explicit `--org` and `--project` flags with a more intuitive positional syntax for issue commands (`view`, `explain`, `plan`).

**Closes CLI-16** - User got unhelpful error when running `sentry issue view EXTENSION-7` without org context.

## New Issue ID Formats

| Format | Example | Description |
|--------|---------|-------------|
| `<org>/ID` | `sentry/EXTENSION-7` | Explicit org prefix |
| `<project>-suffix` | `cli-G` | Searches for project across orgs |
| `suffix` | `G` | Requires DSN context for project |
| `numeric` | `123456789` | Direct fetch by numeric ID |

## Resolution Logic

1. **Alias cache** - Fast local lookup from previous `issue list` results
2. **Project search** - Search for project slug across all accessible orgs
3. **DSN detection** - Use detected DSN for project context (suffix-only)

## Examples

```bash
# Before (required explicit flags)
sentry issue view EXTENSION-7 --org sentry

# After (intuitive prefix syntax)
sentry issue view sentry/EXTENSION-7

# Or let CLI search for project
sentry issue view cli-G
```

## Changes

- `src/commands/issue/{view,explain,plan}.ts` - Updated to use new syntax
- `src/commands/issue/utils.ts` - New resolution logic with `findProjectsBySlug()`
- `src/lib/issue-id.ts` - New parsing functions for issue ID formats
- `test/commands/issue/utils.test.ts` - Updated tests for new API

### DIFF
--- a/src/commands/issue/utils.ts
+++ b/src/commands/issue/utils.ts
@@ -193,30 +193,29 @@ async function resolveSuffixOnly(
 }
 
 /**
- * Resolve explicit-org-suffix type: org provided, need project from DSN.
+ * Resolve explicit-org-suffix type: org provided but only suffix given.
+ *
+ * This format (`org/suffix`) is ambiguous - we have org but no project.
+ * We don't use DSN detection here because mixing explicit org with
+ * DSN-detected project (which belongs to a potentially different org)
+ * would be semantically wrong and confusing.
  *
  * @param org - Explicit organization slug
  * @param suffix - Issue suffix (uppercase)
- * @param cwd - Current working directory
  * @param commandHint - Hint for error messages
  */
-async function resolveExplicitOrgSuffix(
+function resolveExplicitOrgSuffix(
   org: string,
   suffix: string,
-  cwd: string,
   commandHint: string
-): Promise<StrictResolvedIssue> {
-  const target = await resolveOrgAndProject({ cwd });
-  if (target) {
-    const fullShortId = expandToFullShortId(suffix, target.project);
-    const issue = await getIssueByShortId(org, fullShortId);
-    return { org, issue };
-  }
-
+): never {
   throw new ContextError(
     `Cannot resolve suffix '${suffix}' without project context`,
     commandHint,
-    [`Specify the project: sentry issue ... ${org}/<project>-${suffix}`]
+    [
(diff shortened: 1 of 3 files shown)


src/lib/issue-id.ts
```diff
@@ -71,3 +75,82 @@ export function expandToFullShortId(
...
+ * - `project-suffix`: Project slug + suffix (e.g., "cli-G", "spotlight-electron-4Y")
+ * - `suffix`: Short suffix only (e.g., "G", "4Y")
+ * - `numeric`: Numeric issue ID (e.g., "123456789")
+ */
+export type ParsedIssueArg =
+  | { type: "explicit-org"; org: string; rest: string }
+  | { type: "has-dash"; value: string }
+  | { type: "suffix-only"; suffix: string }
+  | { type: "numeric"; id: string };
+
```
BYK: Don't we have very similar functionality that we have in this file in somewhere else that we implemented in https://github.com/getsentry/cli/pull/155 ?

BYK:
Good question! While both functions share the "split on slash" pattern for org prefix, they serve different purposes with different downstream handling:

**`parseOrgProjectArg`** (from PR #155) - For `issue list`, `project list`:
- Parses a **target location**: `org/project`, `org/`, `project`, or auto-detect
- The "rest" after slash is a **project slug**
- Handles cases like `sentry/` (all projects in org)

**`parseIssueArg`** (this PR) - For `issue view`, `explain`, `plan`:
- Parses an **issue identifier**: `org/issue-id`, `project-suffix`, `suffix`, or numeric
- The "rest" after slash is an **issue ID** that needs further parsing (numeric check, dash detection, suffix detection)
- Handles cases like `123456789` (numeric ID), `cli-G` (project-suffix), `G` (suffix only)

The only overlap is the 3-line slash detection. Merging them would mix concerns and require callers to handle cases they don't care about. I think keeping them separate is cleaner.

That said, if you'd prefer, I could extract a tiny `extractOrgPrefix(arg: string): { org: string; rest: string } | null` helper that both functions use. Let me know!

PR #162, opened 2026-02-04
fix: updated the plan command error message when root cause analysis is not done

MathurAditya724:


src/commands/issue/plan.ts
```diff
@@ -74,7 +74,8 @@ function validateAutofixState(
   const causes = extractRootCauses(state);
   if (causes.length === 0) {
     throw new ValidationError(
-      "No root causes identified. Cannot create a plan without a root cause."
+      `Root cause analysis found no causes for issue ${issueId}.\n` +
+        `Try running 'sentry issue explain ${issueId} --force' to retry the analysis.`
```
BYK, line 78: I'd say we should go one step further and offer to run this for the user if this `stdin.isTTY` is true

MathurAditya724:
Closing this in favour of #185

PR #170, opened 2026-02-04
feat(auth): auto-trigger login flow when authentication required

betegon:
## Summary

When a command requires authentication but the user isn't logged in, automatically start the OAuth device flow instead of showing an error. After successful login, the original command retries automatically.

Only triggers in interactive TTY environments - non-TTY sessions (CI, scripts, piped input) continue showing the error message as before.

## Changes

- Extract device flow UI logic into `src/lib/interactive-login.ts` for reuse
- Refactor `auth login` command to use the shared helper
- Add `executeWithAutoAuth()` wrapper in `bin.ts` that catches `AuthError("not_authenticated")` and triggers login

## Test Plan

1. Log out: `sentry auth logout`
2. Run a command that requires auth: `sentry issue list`
3. Should see "Authentication required. Starting login flow..." and the OAuth device flow starts
4. Complete login in browser
5. Original command should automatically retry and succeed

**Non-TTY test:**
```bash
echo "" | sentry issue list
# Should show: "Error: Not authenticated. Run 'sentry auth login' first."
```


src/bin.ts
```diff
@@ -11,6 +13,60 @@ import {
...
+async function executeWithAutoAuth(args: string[]): Promise<void> {
+  try {
+    await withTelemetry(async (span) =>
+      run(app, args, buildContext(process, span))
+    );
+  } catch (err) {
+    // Auto-login for auth errors in interactive TTY environments
+    // Use isatty(0) for reliable stdin TTY detection (process.stdin.isTTY can be undefined in Bun)
+    // Skip for auth commands - they handle their own auth flow (prevents loop on login failure)
+    const isAuthCommand = args[0] === "auth";
```
BYK: Don't do manual argument parsing ever. We need to find a way to get this information from the `err` object itself or stricli.

src/bin.ts
```diff
@@ -11,6 +13,60 @@ import {
...
+        process.stderr,
+        process.stdin
+      );
+
+      if (loginSuccess) {
+        process.stderr.write("\nRetrying command...\n\n");
+        // Retry the original command
+        await withTelemetry(async (span) =>
+          run(app, args, buildContext(process, span))
+        );
```
BYK: nit: This repeated code can/should be abstracted out.

PR #178, opened 2026-02-05
feat(dsn): infer project from directory name when DSN detection fails

BYK:
## Summary

Adds a fallback mechanism that uses the project root directory name to find matching Sentry projects when no DSN is found in the codebase.

- Uses bidirectional word-boundary matching (`\b` regex)
- Directory `cli` matches project `sentry-cli` or `cli-website`
- Directory `sentry-docs` matches project `docs`
- Results are cached using the existing DSN cache with `source: "inferred"`
- Gracefully handles unauthenticated state

## Changes

- **src/lib/api-client.ts**: Add `findProjectsByPattern()` function with `escapeRegex()` and `matchesWordBoundary()` helpers
- **src/lib/dsn/types.ts**: Add `"inferred"` to `DsnSource` type
- **src/lib/resolve-target.ts**: Add `inferFromDirectoryName()` with caching, integrate into `resolveAllTargets()` and `resolveOrgAndProject()`
- **test/lib/resolve-target.test.ts**: Add unit tests for word boundary matching
- **test/lib/word-boundary.property.test.ts**: Add property-based tests using fast-check

## Notes

The implementation plan is available as a git note on the commit.

### DIFF
--- a/src/lib/resolve-target.ts
+++ b/src/lib/resolve-target.ts
@@ -372,51 +372,72 @@ async function inferFromDirectoryName(cwd: string): Promise<ResolvedTargets> {
 
   // Check cache first (reuse DSN cache with source: "inferred")
   const cached = await getCachedDsn(projectRoot);
-  if (cached?.source === "inferred" && cached.resolved) {
-    // Return cached result
+  if (cached?.source === "inferred") {
     const detectedFrom = `directory name "${dirName}"`;
-    return {
-      targets: [
-        {
-          org: cached.resolved.orgSlug,
-          project: cached.resolved.projectSlug,
-          orgDisplay: cached.resolved.orgName,
-          projectDisplay: cached.resolved.projectName,
-          detectedFrom,
-        },
-      ],
-    };
+
+    // Return all cached targets if available
+    if (cached.allResolved && cached.allResolved.length > 0) {
+      const targets = cached.allResolved.map((r) => ({
+        org: r.orgSlug,
+        project: r.projectSlug,
+        orgDisplay: r.orgName,
+        projectDisplay: r.projectName,
+        detectedFrom,
+      }));
+      return {
+        targets,
+        footer:
+          targets.length > 1
+            ? `Found ${targets.length} projects matching directory "${dirName}"`
+            : undefined,
+      };
+    }
+
+    // Fallback to single resolved target (legacy cache entries)
+    if (cached.resolved) {
+      return {
+        targets: [
+          {
+            org: cached.resolved.orgSlug,
+            project: cached.resolved.projectSlug,
+            orgDisplay: cached.resolved.orgName,
+            projectDisplay: cached.resolved.projectName,
+            detectedFrom,
+          },
+        ],
+      };
+    }
   }
 
   // Search for matching projects using word-boundary matching
   let matches: Awaited<ReturnType<typeof findProjectsByPattern>>;
   try {
     matches = await findProjectsByPattern(dirName);
-  } catch (error) {
-    // If not authenticated or API fails, skip inference
-    if (error instanceof AuthError) {
-      return { targets: [] };
-    }
-    throw error;
+  } catch {
+    // If not authenticated or API fails, skip inference silently
+    return { targets: [] };
   }
 
   if (matches.length === 0) {
     return { targets: [] };
   }
 
-  // Cache the first match for faster subsequent lookups
+  // Cache all matches for faster subsequent lookups
   const [primary] = matches;
   if (primary) {
+    const allResolved = matches.map((m) => ({
+      orgSlug: m.orgSlug,
+      orgName: m.organization?.name ?? m.orgSlug,
+      projectSlug: m.slug,
+      projectName: m.name,
+    }));
+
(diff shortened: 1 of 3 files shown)


src/lib/api-client.ts
```diff
@@ -616,6 +616,69 @@ export async function findProjectsBySlug(
...
+ * @example
+ * matchesWordBoundary("cli", "cli-website")  // true: "cli" in "cli-website"
+ * matchesWordBoundary("sentry-docs", "docs") // true: "docs" in "sentry-docs"
+ * matchesWordBoundary("cli", "eclipse")      // false: no word boundary
+ */
+function matchesWordBoundary(a: string, b: string): boolean {
+  const aInB = new RegExp(`\\b${escapeRegex(a)}\\b`, "i");
+  const bInA = new RegExp(`\\b${escapeRegex(b)}\\b`, "i");
+  return aInB.test(b) || bInA.test(a);
+}
```
BYK: Why are these in `api-client.ts`? Don't we have a utils file? Or maybe a new string utils file or something?

src/lib/api-client.ts
```diff
@@ -616,6 +616,69 @@ export async function findProjectsBySlug(
   return searchResults.filter((r): r is ProjectWithOrg => r !== null);
 }
 
+/** Escape special regex characters in a string */
+function escapeRegex(str: string): string {
```
BYK: I wonder if we can use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape

BYK: Done! Replaced `escapeRegex` with the native `RegExp.escape`, which is available in Bun.

src/lib/api-client.ts
BYK: With `escapeRegex` gone (now using native `RegExp.escape`), only `matchesWordBoundary` remains. Since it's only used by `findProjectsByPattern` in this same file, I kept it here rather than creating a separate utils file for a single function. Happy to move it if you prefer!

test/lib/resolve-target.test.ts
```diff
@@ -1,107 +1,108 @@
...
-    const result = parseOrgProjectArg("123/456");
-    expect(result).toEqual({
-      type: "explicit",
-      org: "123",
-      project: "456",
-    });
+  test("hidden directories are valid", () => {
+    expect(isValidDirNameForInference(".env")).toBe(true);
+    expect(isValidDirNameForInference(".git")).toBe(true);
+    expect(isValidDirNameForInference(".config")).toBe(true);
```
BYK: these should NOT be valid

BYK: Fixed! Now rejecting all names starting with `.` (hidden directories). This simplifies the logic since it also covers the dot-only cases like `.` and `..`.

src/lib/resolve-target.ts
```diff
@@ -323,6 +331,113 @@ async function resolveDsnToTarget(
...
+  // Search for matching projects using word-boundary matching
+  let matches: Awaited<ReturnType<typeof findProjectsByPattern>>;
+  try {
+    matches = await findProjectsByPattern(dirName);
+  } catch (error) {
+    // If not authenticated or API fails, skip inference
+    if (error instanceof AuthError) {
+      return { targets: [] };
+    }
+    throw error;
```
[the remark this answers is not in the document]
BYK: Fixed! Changed the catch block to catch all errors (not just `AuthError`) so any API or network failures are handled gracefully by returning an empty array of targets.

src/lib/resolve-target.ts
```diff
@@ -323,6 +331,113 @@ async function resolveDsnToTarget(
...
+  }
+
+  const detectedFrom = `directory name "${dirName}"`;
+  const targets: ResolvedTarget[] = matches.map((m) => ({
+    org: m.orgSlug,
+    project: m.slug,
+    orgDisplay: m.orgSlug,
+    projectDisplay: m.name,
+    detectedFrom,
+  }));
```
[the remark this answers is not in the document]
BYK: Fixed! Now using `m.organization?.name ?? m.orgSlug` for `orgDisplay` in the fresh results path, matching the cached behavior.

src/lib/resolve-target.ts
```diff
@@ -323,6 +331,113 @@ async function resolveDsnToTarget(
...
+      targets: [
+        {
+          org: cached.resolved.orgSlug,
+          project: cached.resolved.projectSlug,
+          orgDisplay: cached.resolved.orgName,
+          projectDisplay: cached.resolved.projectName,
+          detectedFrom,
+        },
+      ],
+    };
```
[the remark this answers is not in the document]
BYK: Fixed! Extended the cache to store all inferred targets:

1. Added `allResolved?: ResolvedProjectInfo[]` to `CachedDsnEntry` type
2. Updated `setCachedDsn` to store all resolved targets in `all_dsns_json`
3. Updated `getCachedDsn` to read all resolved targets from `all_dsns_json`
4. Updated `inferFromDirectoryName` to cache and retrieve all matches

Now both fresh and cached results return the same set of targets.

PR #197, opened 2026-02-05
feat(db): add schema repair and `sentry cli fix` command

BYK:
## Summary

Adds automatic schema repair for SQLite database migration issues. This fixes cases where the schema version gets out of sync with actual table structure during development.

### Changes

**Auto-Repair on SQLiteError**
- When a query fails due to missing columns/tables, the CLI automatically repairs the schema and retries
- Works at both query preparation (`db.query()`) and execution (`stmt.get()`) stages
- Logs repairs to stderr without breaking JSON output
- Opt-out via `SENTRY_CLI_NO_AUTO_REPAIR=1`

**`sentry cli fix` Command**
```bash
sentry cli fix              # Diagnose and repair schema issues
sentry cli fix --dry-run    # Show what would be fixed
```

**Schema Repair Utilities**
- `EXPECTED_TABLES` / `EXPECTED_COLUMNS` - Canonical schema definitions
- `getSchemaIssues()` - Diagnostics for missing tables/columns
- `repairSchema()` - Non-destructive repair (only adds, never deletes)

### Testing
- 14 new tests for schema repair functions
- 5 new tests for fix command
- All 997 tests pass

### DIFF
--- a/src/lib/db/schema.ts
+++ b/src/lib/db/schema.ts
@@ -195,7 +195,6 @@ export const TABLE_SCHEMAS: Record<string, TableSchema> = {
   },
 };
 
-/** Generate CREATE TABLE DDL from a table schema */
 /** Generate CREATE TABLE DDL from column definitions */
 function columnDefsToDDL(
   tableName: string,


src/lib/db/schema.ts
```diff
@@ -48,6 +52,273 @@ const PROJECT_ROOT_CACHE_TABLE = `
   )
 `;
 
+// =============================================================================
+// Schema Definitions for Repair
+// =============================================================================
+
```
BYK: No divider/separator comments

src/lib/db/schema.ts
```diff
@@ -48,6 +52,273 @@ const PROJECT_ROOT_CACHE_TABLE = `
...
+      project_name TEXT NOT NULL,
+      cached_at INTEGER NOT NULL DEFAULT (unixepoch() * 1000),
+      last_accessed INTEGER NOT NULL DEFAULT (unixepoch() * 1000)
+    )
+  `,
+  dsn_cache: `
+    CREATE TABLE IF NOT EXISTS dsn_cache (
+      directory TEXT PRIMARY KEY,
+      dsn TEXT NOT NULL,
+      project_id TEXT NOT NULL,
```
BYK: I feel like generating these from some JS object or a Zod schema would be much better?

src/lib/db/schema.ts
```diff
@@ -48,6 +52,273 @@ const PROJECT_ROOT_CACHE_TABLE = `
...
+/** Schema issue types for diagnostics */
+export type SchemaIssue =
+  | { type: "missing_table"; table: string }
+  | { type: "missing_column"; table: string; column: string };
+
+/**
+ * Check schema and return list of issues.
+ * Used for diagnostics and dry-run mode in `sentry cli fix`.
+ */
+export function getSchemaIssues(db: Database): SchemaIssue[] {
```
BYK, line 328: Since we have this now, do we actually need explicit migrations? We can just fix things on the fly?

src/lib/telemetry.ts
```diff
@@ -357,6 +357,87 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
+    const { fixed } = repairSchema(rawDb);
+
+    if (fixed.length > 0) {
+      // Log to stderr so it doesn't break JSON output
+      console.error(`Auto-repaired database: ${fixed.join(", ")}`);
+      // Retry the operation
+      return operation();
+    }
+  } catch {
+    // Repair failed, will throw original error
```
BYK: Will this really throw the original error?

src/lib/telemetry.ts
```diff
@@ -357,6 +357,87 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
+): T | undefined {
+  // Don't repair if disabled via env var
+  if (process.env[NO_AUTO_REPAIR_ENV] === "1") {
+    return;
+  }
+
+  // Don't repair if not a schema error
+  if (!isSchemaError(error)) {
+    return;
+  }
```
BYK: Shoulg we not just throw the original error instead of silently returning?

src/lib/telemetry.ts
```diff
@@ -400,7 +481,22 @@ function createTracedStatement<T>(stmt: T, sql: string): T {
...
-          () => (value as (...a: unknown[]) => unknown).apply(target, args)
+          () => {
+            const execute = () =>
+              (value as (...a: unknown[]) => unknown).apply(target, args);
+
+            try {
+              return execute();
+            } catch (error) {
+              // Attempt auto-repair for schema errors
+              const result = tryRepairAndRetry(execute, error);
```
BYK: Would be nice if we could separate `telemetry` and `auto-repair` parts so they don't get intermixed.

BYK:
Regarding why migrations are still needed alongside auto-repair:

Auto-repair only handles **additive** schema changes (missing tables/columns). Explicit migrations are still needed for:

- **Data transformations** - e.g., splitting a column, populating new columns from existing data
- **Column renames** - SQLite requires copying data to a new table
- **Complex constraints** - constraints that depend on existing data or need specific ordering

Auto-repair is a safety net for when schema version gets out of sync (e.g., database copied between machines, downgrade then upgrade), not a replacement for migrations.

test/commands/cli/fix.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    const db = new Database(join(testDir, "cli.db"));
+    db.exec(`
+      CREATE TABLE schema_version (version INTEGER PRIMARY KEY);
+      INSERT INTO schema_version (version) VALUES (4);
+      CREATE TABLE auth (id INTEGER PRIMARY KEY);
+      CREATE TABLE defaults (id INTEGER PRIMARY KEY);
+      CREATE TABLE project_cache (cache_key TEXT PRIMARY KEY);
+      CREATE TABLE dsn_cache (
+        directory TEXT PRIMARY KEY,
+        dsn TEXT NOT NULL,
```
BYK: I'd derive these SQL statements from the schema definitions and use the create table helper from there.

PR #208, opened 2026-02-06
fix(upgrade): replace curl pipe with direct binary download

BYK:
## Summary

Fixes `sentry cli upgrade --method curl` failing with `Error: TODO: stream.Readable stdio @ 0`.

The curl upgrade method used `node:child_process.spawn` to pipe curl's stdout to bash's stdin, but Bun's Node.js compatibility layer doesn't support `stream.Readable` in `stdio` options.

Fixes CLI-39.

### Core Fix
- Replace shell-based `curl | bash` with direct binary download via `fetch()`
- Download binary to temp file (`.download`), then atomically rename to target
- Handle Windows "file in use" by renaming running exe to `.old` first
- Clean up `.old` and `.download` files on next CLI startup
- Extend `Bun.write()` polyfill to handle `Response`, `ArrayBuffer`, `Uint8Array`

### Concurrent Upgrade Protection
- Added PID-based lock file mechanism to prevent concurrent upgrades
- `acquireUpgradeLock()` checks if lock file exists, reads PID, verifies if process is still running
- Stale locks from dead processes are automatically cleaned up
- Lock is always released in `finally` block, even on failure

### Test Coverage
- Exported lock functions (`isProcessRunning`, `acquireUpgradeLock`, `releaseUpgradeLock`) for testing
- Exported helper functions (`getBinaryDownloadUrl`, `getCurlInstallPaths`) for testing
- Added comprehensive tests for:
  - Lock acquisition/release mechanics
  - Stale lock detection and cleanup
  - `executeUpgrade("curl", ...)` integration tests
  - Binary download success/failure scenarios
- Function coverage improved from ~65% to ~85%

## Testing

```bash
bun test test/lib/upgrade.test.ts
# 54 pass, 0 fail
```

### DIFF
--- a/src/lib/upgrade.ts
+++ b/src/lib/upgrade.ts
@@ -159,10 +159,14 @@ function handleExistingLock(lockPath: string): void {
   let content: string;
   try {
     content = readFileSync(lockPath, "utf-8").trim();
-  } catch {
-    // Lock file disappeared between our check and read - retry acquisition
-    acquireUpgradeLock(lockPath);
-    return;
+  } catch (error) {
+    // Only retry if file disappeared (ENOENT) - race condition with another process
+    // For other errors (EACCES, etc.), re-throw to avoid infinite recursion
+    if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+      acquireUpgradeLock(lockPath);
+      return;
+    }
+    throw error;
   }
 
   const existingPid = Number.parseInt(content, 10);
@@ -177,8 +181,12 @@ function handleExistingLock(lockPath: string): void {
   // Stale lock from dead process - remove and retry
   try {
     unlinkSync(lockPath);
-  } catch {
-    // Someone else removed it - that's fine
+  } catch (error) {
+    // Only proceed if file already gone (ENOENT) - someone else removed it
+    // For other errors (EACCES, etc.), re-throw to avoid infinite recursion
+    if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+      throw error;
+    }
   }
 
   // Retry acquisition (recursive call handles race with other processes)
--- a/test/lib/upgrade.test.ts
+++ b/test/lib/upgrade.test.ts
@@ -5,9 +5,9 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { chmodSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { unlink } from "node:fs/promises";
-import { homedir } from "node:os";
+import { homedir, platform } from "node:os";
 import { join } from "node:path";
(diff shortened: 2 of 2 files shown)


src/lib/upgrade.ts
```diff
@@ -46,6 +44,62 @@ function getGitHubHeaders() {
...
+}
+
+/**
+ * Clean up old binary from previous upgrade.
+ * Called on CLI startup to remove .old files left from Windows upgrades.
+ */
+export function cleanupOldBinary(): void {
+  const { oldPath } = getCurlInstallPaths();
+  try {
+    unlinkSync(oldPath);
```
BYK: Isn't it better to do `unlink(oldPath).unref()` (from fs/Promises) for truly fire-and-forget and non-blocking?

BYK (commented):
I think we may have weird failures in case somehow we end up having multiple `sentry cli upgrade` calls running at the same time. Also, if the download gets interrupted, what do we do with the left-over file? No clean up? No resume?

src/lib/upgrade.ts
BYK: Good call! Changed to use `unlink()` from `fs/promises` with `.catch()` for fire-and-forget behavior.

Note: Promises don't have `.unref()` (that's for Node.js handles like timers), but the effect is the same - we just don't await the promise and add `.catch()` to suppress errors.

BYK:
## Re: Sentry Bot's Windows Rename Concern

The Sentry bot flagged the Windows upgrade logic at lines 483-502 as a bug, claiming `renameSync` will fail because Windows locks running executables.

**This is a false positive.** Windows file locking behavior is:
- ❌ **Cannot delete** a running executable
- ✅ **Can rename** a running executable

This is a well-documented and widely-used self-update pattern. From [Super User](https://superuser.com/questions/488127):

> "There really is no such thing as renaming a file... Renaming is an operation on the directory entry, which is not affected by the fact that the file is locked for execution."

This pattern is used by:
- **Chrome** - renames running binary during update
- **VS Code** - same approach for self-updates
- **Electron apps** - documented pattern in Electron's auto-updater

The code is correct:
1. Rename running `sentry.exe` → `sentry.exe.old` (allowed)
2. Write new binary to `sentry.exe` (target path is now free)
3. Clean up `.old` on next startup (can now delete since not running)

No changes needed.

src/lib/upgrade.ts
```diff
@@ -46,6 +51,153 @@ function getGitHubHeaders() {
...
+function handleExistingLock(lockPath: string): void {
+  // Lock file exists - read and check if owner process is still alive
+  let content: string;
+  try {
+    content = readFileSync(lockPath, "utf-8").trim();
+  } catch {
+    // Lock file disappeared between our check and read - retry acquisition
+    acquireUpgradeLock(lockPath);
+    return;
+  }
```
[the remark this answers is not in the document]
BYK: Already fixed in commit 50edc9e - the catch blocks now check for `ENOENT` specifically and re-throw other errors. Added a test case that verifies this with an unreadable lock file.

PR #213, opened 2026-02-06
feat(cli): add setup command for shell integration

### LINKED ISSUE #67: Autocomplete commands
https://bloomberg.github.io/stricli/docs/features/shell-autocomplete

BYK:
## Summary

Adds a new `sentry cli setup` command that configures shell integration, replacing the bash-based PATH modification logic with a proper TypeScript implementation.

### New `sentry cli setup` command

- **PATH modification** - Detects shell type, finds config file, adds PATH export
- **Shell completions** - Dynamically generates and installs completions for bash, zsh, and fish
- **Install metadata** - Records method/path/version for future upgrades

```bash
sentry cli setup                    # Auto-detect and configure
sentry cli setup --method curl      # Record install method
sentry cli setup --no-modify-path   # Skip PATH modification
sentry cli setup --no-completions   # Skip shell completions
```

### Simplified install script

The curl install script is now much simpler - it just downloads the binary and calls `sentry cli setup`:

**Install priority order:**
1. `$SENTRY_INSTALL_DIR` (if set)
2. `~/.local/bin` (if exists AND in `$PATH`)
3. `~/bin` (if exists AND in `$PATH`)
4. `~/.sentry/bin` (fallback)

### Dynamic shell completions

Completion scripts are **auto-generated from the Stricli route map** — when commands are added or removed, completions update automatically. No manual maintenance needed.

### Automatic flag telemetry

A `buildCommand` wrapper auto-captures flag values as Sentry tags (e.g., `flag.verbose`, `flag.no-modify-path`) and positional args as Sentry context. Commands just import `buildCommand` from `lib/command.js` instead of `@stricli/core`.

### Kebab-case flag support

Enabled `allow-kebab-for-camel` in Stricli config so all flags accept both `--noModifyPath` and `--no-modify-path`.

### Benefits

| Before | After |
|--------|-------|
| PATH modification in bash (~100 lines) | PATH modification in TypeScript (testable) |
| No shell completions | Tab completion for bash/zsh/fish |
| `record-install` + bash logic | Single `setup` command handles everything |
| Manual download users get no help | Can run `sentry cli setup` to configure |
| No flag telemetry | Automatic flag/arg capture for all commands |
| Static completion scripts | Dynamic generation from route map |

### New files

| File | Purpose |
|------|---------|
| `src/lib/shell.ts` | Shell detection, config file finding, PATH modification |
| `src/lib/completions.ts` | Dynamic shell completion script generation |
| `src/lib/command.ts` | `buildCommand` wrapper with automatic telemetry |
| `src/commands/cli/setup.ts` | The setup command |
| `test/lib/shell.test.ts` | 35 tests |
| `test/lib/completions.test.ts` | 30 tests |
| `test/lib/command.test.ts` | 8 tests |
| `test/commands/cli/setup.test.ts` | 11 tests |

### Removed

 - src/commands/cli/record-install.ts - functionality merged into setup --method

Fixes #67.

### DIFF
--- a/test/lib/completions.property.test.ts
+++ b/test/lib/completions.property.test.ts
@@ -1,0 +1,288 @@
+/**
+ * Property-Based Tests for Shell Completions
+ *
+ * Verifies invariants of the completion generation system:
+ * - Cross-shell consistency (every command appears in all three scripts)
+ * - Binary name parametrization
+ * - Structural invariants of the command tree
+ *
+ * Also includes integration tests using Stricli's proposeCompletions API
+ * and a real bash simulation of the generated completion script.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { join } from "node:path";
+import { proposeCompletions } from "@stricli/core";
+import { constantFrom, assert as fcAssert, property } from "fast-check";
+import { app } from "../../src/app.js";
+import {
+  extractCommandTree,
+  generateBashCompletion,
+  generateFishCompletion,
+  generateZshCompletion,
+} from "../../src/lib/completions.js";
+import { DEFAULT_NUM_RUNS } from "../model-based/helpers.js";
+
+// -- Arbitraries --
+
+/** Generate valid binary names (lowercase alphanumeric + hyphens) */
+const binaryNameArb = constantFrom(
+  "sentry",
+  "my-cli",
+  "test-tool",
+  "acme",
+  "dev-helper",
+  "s"
+);
+
+// -- Helpers --
+
+/** Minimal context for proposeCompletions */
+const completionContext = {
+  process: {
+    stdout: {
+      write: () => true,
+    },
+    stderr: {
+      write: () => true,
+    },
+    stdin: process.stdin,
+  },
+  forCommand: () => ({}),
+};
+
+// -- Tests --
+
+describe("property: extractCommandTree invariants", () => {
+  const tree = extractCommandTree();
+
+  test("every group has at least one subcommand", () => {
+    for (const group of tree.groups) {
+      expect(group.subcommands.length).toBeGreaterThan(0);
+    }
+  });
+
+  test("no duplicate group names", () => {
+    const names = tree.groups.map((g) => g.name);
+    expect(new Set(names).size).toBe(names.length);
+  });
+
+  test("no duplicate subcommand names within a group", () => {
+    for (const group of tree.groups) {
+      const names = group.subcommands.map((s) => s.name);
+      expect(new Set(names).size).toBe(names.length);
+    }
+  });
+
+  test("all names are non-empty strings", () => {
+    for (const group of tree.groups) {
+      expect(group.name.length).toBeGreaterThan(0);
+      expect(group.brief.length).toBeGreaterThan(0);
+      for (const sub of group.subcommands) {
+        expect(sub.name.length).toBeGreaterThan(0);
+        expect(sub.brief.length).toBeGreaterThan(0);
+      }
+    }
+    for (const cmd of tree.standalone) {
+      expect(cmd.name.length).toBeGreaterThan(0);
+      expect(cmd.brief.length).toBeGreaterThan(0);
+    }
+  });
+});
+
+describe("property: cross-shell consistency", () => {
+  const tree = extractCommandTree();
+
+  test("every group name appears in all three shell scripts", () => {
+    fcAssert(
+      property(binaryNameArb, (name) => {
+        const bash = generateBashCompletion(name);
+        const zsh = generateZshCompletion(name);
+        const fish = generateFishCompletion(name);
+
+        for (const group of tree.groups) {
+          expect(bash).toContain(group.name);
+          expect(zsh).toContain(group.name);
+          expect(fish).toContain(group.name);
+        }
+      }),
+      { numRuns: DEFAULT_NUM_RUNS }
+    );
+  });
+
+  test("every subcommand name appears in all three shell scripts", () => {
+    fcAssert(
+      property(binaryNameArb, (name) => {
+        const bash = generateBashCompletion(name);
+        const zsh = generateZshCompletion(name);
+        const fish = generateFishCompletion(name);
+
+        for (const group of tree.groups) {
+          for (const sub of group.subcommands) {
+            expect(bash).toContain(sub.name);
+            expect(zsh).toContain(sub.name);
+            expect(fish).toContain(sub.name);
+          }
+        }
+      }),
+      { numRuns: DEFAULT_NUM_RUNS }
+    );
+  });
+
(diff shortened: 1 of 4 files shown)


install
```diff
@@ -80,22 +96,58 @@ version="${version#v}"
...
+
+# Download binary
 echo -e "${MUTED}Downloading sentry v${version}...${NC}"
 curl -fsSL --progress-bar "$url" -o "$install_path"
 chmod +x "$install_path"
 
+# Run setup command to configure PATH, completions, and record install info
+setup_args="--method curl"
+if [[ "$no_modify_path" == "true" ]]; then
+  setup_args="$setup_args --noModifyPath"
```
BYK: I think our CLI should support `--no-modify-path` to make it align with Posix conventions. Let's fix that.

src/lib/command.ts
```diff
@@ -0,0 +1,95 @@
...
+  const FLAGS extends BaseFlags = NonNullable<unknown>,
+  const ARGS extends BaseArgs = [],
+  const CONTEXT extends CommandContext = CommandContext,
+>(
+  builderArgs: LocalCommandBuilderArguments<FLAGS, ARGS, CONTEXT>
+): Command<CONTEXT> {
+  const originalFunc = builderArgs.func;
+
+  // Wrap the function to capture flags before execution
+  const wrappedFunc = function (
```
BYK, line 77: Can we also record the context and args in Sentry? Also I'm wondering if we still need our `withTelemetry` top-level wrapper since now we have direct access to the commands?

test/lib/telemetry.test.ts
```diff
@@ -149,6 +150,77 @@ describe("setOrgProjectContext", () => {
...
+describe("setFlagContext", () => {
+  test("handles empty flags object", () => {
+    expect(() => setFlagContext({})).not.toThrow();
+  });
+
+  test("handles boolean flags (true sets tag)", () => {
+    expect(() => setFlagContext({ verbose: true, debug: true })).not.toThrow();
+  });
+
+  test("handles boolean flags (false is skipped)", () => {
```
BYK: `.not.toThrow()` is a very low bar. Can we not mock out Sentry API and actually check the correct values are being set (or not set)?

src/lib/completions.ts
```diff
@@ -0,0 +1,311 @@
...
+  /** Path where the completion file was installed */
+  path: string;
+  /** Whether the file was created or already existed */
+  created: boolean;
+};
+
+/**
+ * Generate bash completion script.
+ */
+export function generateBashCompletion(binaryName: string): string {
```
BYK, line 98: I think Stricli has _some_ support for generating this automatically, why are we rolling our own: https://bloomberg.github.io/stricli/docs/features/shell-autocomplete

install
BYK: Fixed! Updated the install script to use `--no-modify-path` and `--no-completions` (kebab-case) which Stricli already supports automatically.

src/lib/command.ts
BYK: Added positional args capture via `setArgsContext()` which stores them as Sentry context (not tags, since args can be long/complex).

Regarding `withTelemetry` - we still need it because it handles:
- **SDK initialization** (`Sentry.init()`)
- **Session lifecycle** (`startSession`/`endSession`)
- **Root span/transaction creation** for the CLI command
- **Error capture** for unhandled exceptions  
- **Flush on exit** to ensure events are sent before process terminates

The `buildCommand` wrapper operates *within* the span created by `withTelemetry` - it just adds flag/arg context to that existing span. Without `withTelemetry`, there would be no telemetry infrastructure at all.

test/lib/telemetry.test.ts
BYK: Good call! Rewrote the tests to use `spyOn(Sentry, "setTag")` and verify:
- Correct tag names (with `flag.` prefix and kebab-case conversion)
- Correct values (booleans as "true", numbers/strings as-is)
- Skipped values (false, undefined, null, empty strings, empty arrays)
- Long value truncation (200 char limit)

Also added tests for `setArgsContext()` that verify `Sentry.setContext()` calls.

src/lib/completions.ts
BYK: Good question! I looked into Stricli's `@stricli/auto-complete` package and there are a few reasons we rolled our own:

1. **Shell support**: Stricli only supports **bash** currently ("Support for other shells is planned for the future"). We support **bash, zsh, and fish**.

2. **Architecture**: Stricli uses a **dynamic approach** where the CLI binary is invoked for every completion request. Our approach generates **static completion scripts** that work without invoking the CLI.

3. **Integration**: Our completions are installed via `sentry cli setup` as part of the install flow, rather than requiring a separate `npx @stricli/auto-complete install` step.

That said, if Stricli adds zsh/fish support in the future, we could consider migrating to it.

src/lib/completions.ts
```diff
@@ -0,0 +1,311 @@
...
+ * Generate fish completion script.
+ */
+export function generateFishCompletion(binaryName: string): string {
+  return `# fish completion for ${binaryName}
+# Install: ${binaryName} cli setup
+
+# Disable file completion by default
+complete -c ${binaryName} -f
+
+# Top-level commands
```
BYK, line 247: Pretty sure we can shorten this with some generation logic as there's a lot of repetition here.

src/lib/completions.ts
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * Get the completion script for a shell type.
+ */
+export function getCompletionScript(
+  shellType: ShellType,
+  binaryName = "sentry"
+): string | null {
+  switch (shellType) {
+    case "bash":
```
BYK, line 261: I wonder if we can also add support for `xonsh`? (https://xon.sh/)

src/lib/completions.ts
BYK: Completely rewrote this. Completion scripts are now **dynamically generated** from the Stricli route map at `app.ts`. The `extractCommandTree()` function walks `routes.getAllEntries()` recursively to discover all commands, subcommands, and aliases.

This means:
- **Zero manual maintenance** — when commands are added/removed, completions update automatically
- **Fixed all missing commands** — `log/list`, `auth/token`, and aliases (`issues`, `orgs`, `projects`, `logs`) are now included
- **Single source of truth** — command names and descriptions come directly from the Stricli route definitions

src/lib/completions.ts
BYK: xonsh is pretty niche — I'd rather wait for user demand before adding it. The good news is the refactor makes adding new shells trivial: just add a new `generateXonshCompletion()` function that consumes the same `extractCommandTree()` output, and add `"xonsh"` to the `ShellType` union. Opening a tracking issue would make sense if we want to remember this.

PR #224, opened 2026-02-10
chore(ci): auto-commit SKILL.md when stale

betegon:
## Summary

When `check-skill` detects SKILL.md is out of date, auto-commit the regenerated file instead of failing CI with instructions.

`check-skill.ts` already runs `generate-skill.ts` internally and leaves the updated file on disk — so we just need to `git add`, `commit`, and `push` when the check fails.

## Changes

Replaced the "Output regeneration instructions" step (which printed manual fix commands and failed the build) with a step that commits and pushes the regenerated SKILL.md using the `github-actions[bot]` identity.

### DIFF
--- a/plugins/sentry-cli/skills/sentry-cli/SKILL.md
+++ b/plugins/sentry-cli/skills/sentry-cli/SKILL.md
@@ -100,7 +100,7 @@ Work with Sentry organizations
 
 #### `sentry org list`
 
-List organizations (test auto-commit)
+List organizations
 
 **Flags:**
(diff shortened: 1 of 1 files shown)


BYK:
We should do this, which will then work: https://github.com/getsentry/sentry/blob/ff5444fbcdca906c03c53346242297120142269b/.github/workflows/pre-commit.yml#L34-L54

BYK (approved):
Token flow now matches release.yml (actions/create-github-app-token with SENTRY_RELEASE_BOT). Auto-commit verified working — CI auto-committed regenerated SKILL.md successfully. Test change reverted.

PR #226, opened 2026-02-10
ref: use @sentry/api client for requests

MathurAditya724:
## Summary

Migrates the Sentry API client from hand-rolled `ky` HTTP calls with Zod-validated responses to the [`@sentry/api`](https://www.npmjs.com/package/@sentry/api) SDK package. This gives us type-safe API calls generated from the OpenAPI spec, while keeping raw requests for internal/undocumented endpoints.

This supersedes #164, which generated the client in-repo using `@hey-api/openapi-ts`. Using the published `@sentry/api` package instead avoids maintaining ~20k lines of generated code in the repo.

### New: `src/lib/sentry-client.ts`
Request configuration layer for `@sentry/api` SDK functions. Provides:
- **Authenticated fetch** with Bearer token injection
- **Automatic retry** on transient errors (408, 429, 5xx) with exponential backoff
- **401 token refresh** — force-refreshes OAuth token and retries once
- **Request timeout** via AbortController (30s)
- Per-region config via `getSdkConfig(regionUrl)`

### Rewritten: `src/lib/api-client.ts`
All API functions now use `@sentry/api` SDK functions where possible:

| Function | SDK function used |
|----------|------------------|
| `listOrganizationsInRegion` | `listYourOrganizations` |
| `getOrganization` | `retrieveAnOrganization` |
| `listProjects` | `listAnOrganization_sProjects` |
| `getProject` | `retrieveAProject` |
| `getProjectKeys` | `listAProject_sClientKeys` |
| `getIssueByShortId` | `resolveAShortId` |
| `getLatestEvent` | `retrieveAnIssueEvent` |
| `getEvent` | `retrieveAnEventForAProject` |
| `triggerRootCauseAnalysis` | `startSeerIssueFix` |
| `getAutofixState` | `retrieveSeerIssueFixState` |
| `listLogs` / `getLog` | `queryExploreEventsInTableFormat` |

Functions that use raw requests (no SDK equivalent):
- `getUserRegions` — `/users/me/regions/` (internal endpoint)
- `getCurrentUser` — `/users/me/` (internal endpoint)
- `findProjectByDsnKey` — `/projects/?query=dsn:...` (internal query param)
- `listIssues` — SDK missing `limit`/`sort` params
- `getIssue` / `updateIssueStatus` — legacy `/issues/{id}/` endpoint (no org slug needed)
- `getDetailedTrace` — `/organizations/{org}/trace/{traceId}/` (internal endpoint)
- `triggerSolutionPlanning` — SDK doesn't support `run_id` + `step` body

### Refactored: `src/types/sentry.ts`
- SDK-backed types now derive from `@sentry/api` response types using `Partial<SdkType> & RequiredCore`
- Keeps Zod schemas only for internal endpoints not covered by the SDK (Region, User, Logs)
- Event entry types (exceptions, breadcrumbs, request, etc.) remain as plain TypeScript interfaces

### Removed
- `ky` dependency (replaced by `@sentry/api` + native fetch)
- Most Zod schemas for SDK-covered types (Organization, Project, Issue, Event, ProjectKey, etc.)
- `SentryOrganizationSchema` and other schema exports from `types/index.ts`


src/lib/api-client.ts
```diff
@@ -1,32 +1,41 @@
...
+ * Falls back to raw requests for internal/undocumented endpoints.
  */
 
-import kyHttpClient, { type KyInstance } from "ky";
-import { z } from "zod";
 import {
-  type DetailedLogsResponse,
+  listAnOrganization_sIssues,
+  listAnOrganization_sProjects,
+  listAProject_sClientKeys,
```
BYK, line 15: These are some weird-ass names lol

src/lib/api-client.ts
```diff
@@ -36,53 +45,20 @@ import {
...
-import { DEFAULT_SENTRY_URL, getUserAgent } from "./constants.js";
-import { refreshToken } from "./db/auth.js";
 import { ApiError, AuthError } from "./errors.js";
+import { resolveOrgRegion } from "./region.js";
+import {
+  getApiBaseUrl,
+  getControlSiloUrl,
+  getDefaultSdkConfig,
+  getSdkConfig,
+} from "./sentry-client.js";
```
BYK, line 59: What's the value of splitting `sentry-client` off? the `api-client` _is_ the Sentry Client and it cannot be anything else anyway?

src/lib/api-client.ts
```diff
@@ -432,9 +329,9 @@ export async function apiRequestToRegion<T>(
  * @returns Array of regions with name and URL
  */
 export async function getUserRegions(): Promise<Region[]> {
-  // Always use control silo for this endpoint
+  // /users/me/regions/ is an internal endpoint - use raw request
   const response = await apiRequestToRegion<UserRegionsResponse>(
```
BYK, line 329: We should automate this logic: if the SDK has the endpoint, use that, if not fallback to our hand-rolled one.

src/lib/api-client.ts
```diff
@@ -673,14 +538,6 @@ export function matchesWordBoundary(a: string, b: string): boolean {
...
  * Used for directory name inference when DSN detection fails.
  *
- * Uses `\b` regex word boundary, which matches:
- * - Start/end of string
- * - Between word char (`\w`) and non-word char (like "-")
- *
- * Matching is bidirectional:
- * - Directory name in project slug: dir "cli" matches project "cli-website"
- * - Project slug in directory name: project "docs" matches dir "sentry-docs"
- *
```
BYK, line 695: Why are we deleting this comment? Restore it?

src/lib/api-client.ts
```diff
@@ -713,8 +569,7 @@ export async function findProjectsByPattern(
  * Find a project by DSN public key.
  *
  * Uses the /api/0/projects/ endpoint with query=dsn:<key> to search
- * across all accessible projects in all regions. This works for both
```
BYK, line 728: Same as above, get my useful comments back.

src/lib/api-client.ts
```diff
@@ -777,10 +629,22 @@ export function getProject(
   orgSlug: string,
   projectSlug: string
 ): Promise<SentryProject> {
-  return orgScopedRequest<SentryProject>(
+  return withHttpSpan(
```
BYK: We should move the `withHttpSpan` wrapper to the `apiRequestToRegion` and other request layer. We cannot afford to wrap every single function with this.

src/types/sentry.ts
BYK, line 1: I'd argue all the types from this file should also come from the API based on the OpenAPI schema

BYK (changes requested):
I guess we lost the retry logic now? (unless it is baked into the API SDK) We should get that back.

src/lib/api-client.ts
MathurAditya724: I have 2 words for you - "Circular dependency", between `sentry-client.ts`, `api-client.ts` and `region.ts`

src/lib/api-client.ts
MathurAditya724: Auto generated, so can't change that

src/lib/api-client.ts
MathurAditya724: These endpoints are not part of the public API Spec, and the reason why they are not in the SDK

src/lib/api-client.ts
MathurAditya724: That was my stupidity

src/types/sentry.ts
MathurAditya724: The current state of the `@sentry/api` OpenAPI schema has significant gaps that make it impractical for many of these types. Either they are too generic, missing fields or internal endpoints

BYK: Let's make a note to fix this then?

src/lib/api-client.ts
BYK, line 921: Since when we can update issue statuses?! @betegon

src/lib/sentry-client.ts
```diff
@@ -177,6 +178,23 @@ function handleFetchError(
...
+/** Extract the URL pathname for span naming */
+function extractUrlPath(input: Request | string | URL): string {
+  let raw: string;
+  if (typeof input === "string") {
+    raw = input;
+  } else if (input instanceof URL) {
+    raw = input.href;
+  } else {
+    raw = input.url;
+  }
```
BYK, line 190: If it is already an instance of URL, why not just do `return input.pathname`??

PR #235, opened 2026-02-11
fix(db): handle readonly database gracefully instead of crashing

betegon:
## Summary

When the CLI's local database (`~/.sentry/cli.db`) is on a read-only filesystem or has wrong permissions, DB writes throw `SQLiteError: attempt to write a readonly database` and crash commands. Now the CLI catches these errors centrally in the traced DB proxy, warns the user once, and continues working — reads still succeed, only caching/persistence is lost.

Closes https://sentry.sentry.io/issues/7258516238/

## Changes

Readonly errors are caught in the traced DB proxy (`createTracedStatement`) alongside the existing schema auto-repair logic. A one-time warning prints to stderr pointing users to `sentry cli fix`. This single change protects all ~14 DB write paths without touching individual modules.

`sentry cli fix` now also checks file permissions on the DB file, WAL/SHM journals, and config directory — and repairs them with `chmod` when possible.

## Test Plan

- `bun run typecheck` — clean
- `bun run lint` — clean
- 88 tests pass across the 3 changed test files (schema, telemetry, fix command)
- New tests: `isReadonlyError()` detection (5 cases), readonly proxy behavior (4 cases: no throw on write, reads work, warns once, message content)

### DIFF
--- a/test/commands/cli/fix.test.ts
+++ b/test/commands/cli/fix.test.ts
@@ -299,4 +299,73 @@ describe("sentry cli fix", () => {
     expect(stdout).toContain("repaired successfully");
     expect(exitCode).toBe(0);
   });
+
+  test("repairs missing columns and reports success", async () => {
+    // Create database with pre-migration tables then repair (non-dry-run)
+    // This exercises the schema repair success path
+    const db = new Database(join(getTestDir(), "cli.db"));
+    createPreMigrationDatabase(db);
+    db.close();
+
+    const { stdout, exitCode } = await runFix(false);
+
+    expect(stdout).toContain("schema issue(s)");
+    expect(stdout).toContain("Missing column");
+    expect(stdout).toContain("Repairing schema");
+    expect(stdout).toContain("Added column");
+    expect(stdout).toContain("repaired successfully");
+    expect(exitCode).toBe(0);
+  });
+
+  test("sets exitCode=1 when schema check throws with no permission issues", async () => {
+    // Create a DB file that cannot be opened by getRawDatabase.
+    // Write garbage so SQLite cannot parse it — getRawDatabase will throw.
+    const dbPath = join(getTestDir(), "cli.db");
+    closeDatabase();
+    await Bun.write(dbPath, "not a sqlite database");
+    chmodSync(dbPath, 0o600);
+    chmodSync(getTestDir(), 0o700);
+
+    const { stdout, stderr, exitCode } = await runFix(false);
+
+    // No permission issues found, so schema failure should be reported
+    expect(stderr).toContain("Could not open database to check schema");
+    expect(stderr).toContain("Try deleting the database");
+    expect(exitCode).toBe(1);
+    // Should NOT say "No issues found"
+    expect(stdout).not.toContain("No issues found");
+  });
+
+  test("dry-run sets exitCode=1 when schema check throws", async () => {
+    // Same corrupt DB scenario, but in dry-run mode
+    const dbPath = join(getTestDir(), "cli.db");
+    closeDatabase();
+    await Bun.write(dbPath, "not a sqlite database");
+    chmodSync(dbPath, 0o600);
+    chmodSync(getTestDir(), 0o700);
+
+    const { stdout, stderr, exitCode } = await runFix(true);
+
+    expect(stderr).toContain("Could not open database to check schema");
+    expect(exitCode).toBe(1);
+    // Should NOT suggest running fix (no fixable issues found)
+    expect(stdout).not.toContain("Run 'sentry cli fix' to apply fixes");
+  });
+
+  test("schema check failure with permission issues does not print schema error", async () => {
+    // When permissions are broken AND schema can't be opened, the schema error
+    // is suppressed because permission issues are the likely root cause.
+    getDatabase();
+    const dbPath = join(getTestDir(), "cli.db");
+
+    // Make DB readonly — will cause permission issue AND potentially schema failure
+    chmodSync(dbPath, 0o444);
+
+    // The schema catch block should suppress the error message when perm.found > 0
+    const { stdout, stderr } = await runFix(true);
+
+    expect(stdout).toContain("permission issue(s)");
+    // Should NOT print "Could not open database" since permission issues explain it
+    expect(stderr).not.toContain("Could not open database");
+  });
 });
--- a/test/lib/telemetry.test.ts
+++ b/test/lib/telemetry.test.ts
@@ -869,5 +869,40 @@ describe("createTracedDatabase", () => {
 
       db.close();
     });
+
+    test("shows readonly warning when auto-repair fails", () => {
+      // Mock chmodSync to always throw, simulating a file owned by another user.
+      // This makes tryRepairReadonly fail and fall through to warnReadonlyDatabaseOnce.
+      const { mock: mockFn } = require("bun:test");
+      mockFn.module("node:fs", () => {
+        const realFs = require("node:fs");
+        return {
+          ...realFs,
+          chmodSync: () => {
+            throw Object.assign(new Error("EPERM: operation not permitted"), {
+              code: "EPERM",
+            });
+          },
+        };
+      });
+
+      const db = new Database(dbPath);
+      const tracedDb = createTracedDatabase(db);
+
+      const stderrSpy = spyOn(process.stderr, "write");
+
+      tracedDb.query("INSERT INTO test (id, name) VALUES (?, ?)").run(2, "Bob");
+
+      const output = stderrSpy.mock.calls.map((c) => String(c[0])).join("");
+      // When repair fails, the warning message should appear instead
+      expect(output).toContain("read-only");
+      expect(output).toContain("sentry cli fix");
+
+      stderrSpy.mockRestore();
+      db.close();
+
+      // Restore mock
+      mockFn.module("node:fs", () => require("node:fs"));
+    });
   });
 });


src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+ * @param expectedMode - Bitmask of required permission bits (e.g., 0o700)
+ * @returns Object with the actual mode if permissions are insufficient, or null if OK.
+ *          Returns null if the path doesn't exist (missing files aren't a permission problem).
+ */
+function checkMode(
+  path: string,
+  expectedMode: number
+): { actualMode: number } | null {
+  try {
+    const st = statSync(path);
```
BYK: Always prefer async versions.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+    // biome-ignore lint/suspicious/noBitwiseOperators: extracting permission bits with bitmask
+    const mode = st.mode & 0o777;
+    // biome-ignore lint/suspicious/noBitwiseOperators: checking required permission bits are set
+    if ((mode & expectedMode) !== expectedMode) {
+      return { actualMode: mode };
+    }
+  } catch {
+    // File/dir doesn't exist — not a permission issue
```
BYK: You don't know this unless you explicitly check for `ENOENT`. You should handle that and report the other errors (both to sentry and the user)

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+    const fileCheck = checkMode(path, EXPECTED_FILE_MODE);
+    if (fileCheck) {
+      issues.push({
+        path,
+        kind,
+        currentMode: fileCheck.actualMode,
+        expectedMode: EXPECTED_FILE_MODE,
+      });
+    }
+  }
```
BYK: If you switch to async above, you can run all these in parallel with `Promise.all()`

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+      chmodSync(issue.path, issue.expectedMode);
+      fixed.push(
+        `${issue.kind} ${issue.path}: ${formatMode(issue.currentMode)} -> ${formatMode(issue.expectedMode)}`
+      );
+    } catch (error) {
+      const reason =
+        error instanceof Error ? error.message : "permission denied";
+      failed.push(`${issue.kind} ${issue.path}: ${reason}`);
+    }
+  }
```
BYK: Again, all async and parallel please.

src/commands/cli/fix.ts
```diff
@@ -47,51 +324,5 @@ export const fixCommand = buildCommand({
...
-      stderr.write(
-        `\nTry deleting the database and restarting: rm ${dbPath}\n`
-      );
-      proc.exitCode = 1;
-      return;
-    }
-
-    stdout.write("\nDatabase repaired successfully.\n");
-  },
+  func: fixFunc,
```
BYK: Inline `fixFunc` here, it looks weird and useless to take it out above.

src/lib/db/schema.ts
```diff
@@ -438,6 +438,22 @@ function isSchemaError(error: unknown): boolean {
...
+ *
+ * This happens when the CLI's local database file or its containing directory
+ * lacks write permissions (e.g., installed globally in a protected path,
+ * read-only filesystem, or changed permissions).
+ */
+export function isReadonlyError(error: unknown): boolean {
+  if (error instanceof Error && error.name === "SQLiteError") {
+    return error.message
+      .toLowerCase()
+      .includes("attempt to write a readonly database");
```
BYK, line 452: I think we are being a bit too specific with the message, just checking for `readonly` should suffice. But let's roll with this until this becomes an issue.

src/lib/telemetry.ts
```diff
@@ -522,9 +456,63 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
+let readonlyWarningShown = false;
+
+/**
+ * Print a one-time warning to stderr when the local database is read-only.
+ * Does nothing if the warning was already shown this process.
+ *
+ * Uses lazy require for db/index.js to avoid a circular dependency
+ * (db/index.ts imports createTracedDatabase from this module).
+ */
+function warnReadonlyDatabaseOnce(): void {
```
BYK: Trick: you can define `warnReadonlyDatabase` with `let` and replace its own reference with a `noop` function on the first run.

src/lib/telemetry.ts
```diff
@@ -522,9 +456,63 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
 /** Methods on SQLite Statement that execute queries and should be traced */
 const TRACED_STATEMENT_METHODS = ["get", "run", "all", "values"] as const;
 
+/**
+ * Handle a readonly database error by warning the user and returning a
+ * type-appropriate no-op value. Returns `undefined` for run/get (void / no-row)
+ * and `[]` for all/values (empty result set).
+ */
+function handleReadonlyError(method: string | symbol): unknown {
+  warnReadonlyDatabaseOnce();
```
BYK: We actually catch schema errors and run the fix command on the fly. I'd argue we should do the same: try to fix before the warning and show the warning if the fix fails.

src/commands/cli/fix.ts
BYK: Done. Converted `statSync` → `stat` from `node:fs/promises`, `chmodSync` → `chmod` from `node:fs/promises`. All file operations in fix.ts are now async.

src/commands/cli/fix.ts
BYK: Done. Now checks `error.code === 'ENOENT'` explicitly and re-throws unexpected errors.

src/commands/cli/fix.ts
BYK: Done. `checkPermissions` is now async and uses `Promise.all()` for parallel checks.

src/commands/cli/fix.ts
BYK: Done. `repairPermissions` is now async and uses `Promise.allSettled()` for parallel repairs.

src/commands/cli/fix.ts
BYK: Done. `fixFunc` inlined directly into `buildCommand({ async func() {...} })`.

src/lib/db/schema.ts
BYK: Agreed — it's more specific than strictly necessary, but it's also more resistant to false positives if SQLite ever changes its error wording for other readonly conditions. Let's keep it for now and revisit if it causes issues.

src/lib/telemetry.ts
BYK: Done. `warnReadonlyDatabaseOnce` is now `let`-bound and self-replaces with a `noop` on first call. A shared `const noop` avoids creating new empty closures. `resetReadonlyWarning()` restores the original function for tests.

src/lib/telemetry.ts
BYK: Done. Added `tryRepairReadonly()` that runs `chmodSync` on db/wal/shm files before showing any warning. One subtlety: SQLite caches readonly state at connection open time, so even after successful chmod the *current* connection stays readonly — only a new `Database()` connection picks up the fixed permissions. So the function repairs for FUTURE commands and prints a differentiated message:

- Repair succeeded: "Database permissions were auto-repaired. Caching will resume on next command."
- Repair already attempted (guard prevents retry): falls through to `warnReadonlyDatabaseOnce` with the `sentry cli fix` message.

Both are one-shot via the noop pattern. `resetReadonlyWarning()` resets all state.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+function checkMode(
+  path: string,
+  expectedMode: number
+): { actualMode: number } | null {
+  try {
+    const st = statSync(path);
+    // biome-ignore lint/suspicious/noBitwiseOperators: extracting permission bits with bitmask
+    const mode = st.mode & 0o777;
+    // biome-ignore lint/suspicious/noBitwiseOperators: checking required permission bits are set
+    if ((mode & expectedMode) !== expectedMode) {
```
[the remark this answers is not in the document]
BYK: Fixed in 2dcf852. Changed `checkMode` to use exact mode comparison (`mode !== expectedMode`) instead of bitmask. Tests updated to `chmodSync(dbPath, 0o600)` after DB creation to match production `setDbPermissions` behavior.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,298 @@ type FixFlags = {
...
+  const { stdout, process: proc } = this;
+  const dbPath = getDbPath();
+  const dryRun = flags["dry-run"];
+  const out = { stdout, stderr: this.stderr };
+
+  stdout.write(`Database: ${dbPath}\n`);
+  stdout.write(`Expected schema version: ${CURRENT_SCHEMA_VERSION}\n\n`);
+
+  const perm = handlePermissionIssues(dbPath, dryRun, out);
+  const schema = handleSchemaIssues(dbPath, dryRun, out);
```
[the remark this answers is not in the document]
BYK: Fixed in 2dcf852. Wrapped `handleSchemaIssues` call in a try/catch inside `func()` — if it throws (e.g. readonly DB during `--dry-run`), we log a warning and continue with the rest of the checks instead of crashing.

src/lib/telemetry.ts
```diff
@@ -586,9 +587,129 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
+    }
+    try {
+      chmodSync(`${dbPath}-shm`, 0o600);
+    } catch {
+      // SHM file may not exist
+    }
+
+    // Disable the fallback warning — repair succeeded
+    warnReadonlyDatabaseOnce = noop;
+
```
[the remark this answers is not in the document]
BYK: Fixed in 2dcf852. `tryRepairReadonly()` now also `chmodSync`s the config directory (`dirname(dbPath)`) to `0o700` before repairing individual DB files.

src/lib/telemetry.ts
```diff
@@ -586,9 +587,129 @@ export function withDbSpan<T>(operation: string, fn: () => T): T {
...
+    chmodSync(dbPath, 0o600);
+    try {
+      chmodSync(`${dbPath}-wal`, 0o600);
+    } catch {
+      // WAL file may not exist
+    }
+    try {
+      chmodSync(`${dbPath}-shm`, 0o600);
+    } catch {
+      // SHM file may not exist
```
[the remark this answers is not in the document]
BYK: Fixed in 2dcf852. Extracted a `chmodIfExists()` helper that only catches `ENOENT` (file not found) and re-throws all other errors. WAL/SHM repair now uses this helper instead of a bare catch-all.

src/commands/cli/fix.ts
```diff
@@ -47,51 +311,54 @@ export const fixCommand = buildCommand({
...
-    }
-    stdout.write("\n");
+    const totalFound = perm.found + schema.found;
 
-    if (flags["dry-run"]) {
-      stdout.write("Run 'sentry cli fix' to apply fixes.\n");
+    if (totalFound === 0) {
+      stdout.write(
+        "No issues found. Database schema and permissions are correct.\n"
+      );
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in 587d8fd. The `totalFound === 0` check now also gates on `!anyFailed`, so when `handleSchemaIssues` throws (setting `repairFailed: true`), we no longer print "No issues found". Instead `exitCode` is set to 1. Also improved the dry-run path to only suggest "Run sentry cli fix" when there are actual fixable issues, not just failures.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,285 @@ type FixFlags = {
...
+ */
+async function checkMode(
+  path: string,
+  expectedMode: number
+): Promise<{ actualMode: number } | null> {
+  try {
+    const st = await stat(path);
+    // biome-ignore lint/suspicious/noBitwiseOperators: extracting permission bits with bitmask
+    const mode = st.mode & 0o777;
+    if (mode !== expectedMode) {
```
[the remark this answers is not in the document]
BYK: This CLI targets macOS and Linux only — Windows is not a supported platform. The POSIX permission model is used throughout the codebase (see `setDbPermissions` in `db/index.ts`). Adding Windows-specific permission handling is out of scope for this PR.

src/commands/cli/fix.ts
```diff
@@ -47,51 +311,60 @@ export const fixCommand = buildCommand({
...
 
-    if (issues.length === 0) {
-      stdout.write("No issues found. Database schema is up to date.\n");
-      return;
+    // Schema check opens the database, which can throw if the DB or config
+    // directory is readonly. Guard with try/catch so --dry-run can finish
+    // diagnostics even when the filesystem is broken.
+    let schema: DiagnoseResult;
+    try {
+      schema = handleSchemaIssues(dbPath, dryRun, out);
```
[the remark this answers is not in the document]
BYK: This is a known limitation documented in the codebase. SQLite caches readonly state at connection open time — `chmod` + retry on the same `Database` instance still fails because `bun:sqlite` opened the file descriptor as readonly. `closeDatabase()` + reopen would fix it, but the connection is shared and closing mid-operation would break other queries.

The design is intentional: `sentry cli fix` repairs permissions and schema for **future commands**. If both permission and schema issues exist, the first run repairs permissions, and the next invocation (with a fresh connection) can repair schema. The catch block we added ensures this doesn't crash.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,285 @@ type FixFlags = {
...
+    // Missing files aren't a permission problem (WAL/SHM created on demand)
+    if (
+      error instanceof Error &&
+      (error as NodeJS.ErrnoException).code === "ENOENT"
+    ) {
+      return null;
+    }
+    // Unexpected filesystem error — re-throw so it surfaces to the user
+    // and gets captured by the top-level Sentry error handler in bin.ts
+    throw error;
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in 6887a60. `checkMode()` now treats `EACCES` the same as `ENOENT` — the file is skipped since the directory permission check (which runs in parallel) will catch the root cause. This prevents `Promise.all` from rejecting when the config directory lacks execute permission.

src/commands/cli/fix.ts
```diff
@@ -18,22 +19,286 @@ type FixFlags = {
...
+ * @returns Separate lists of human-readable repair successes and failures
+ */
+async function repairPermissions(issues: PermissionIssue[]): Promise<{
+  fixed: string[];
+  failed: string[];
+}> {
+  const results = await Promise.allSettled(
+    issues.map(async (issue) => {
+      await chmod(issue.path, issue.expectedMode);
+      return `${issue.kind} ${issue.path}: ${formatMode(issue.currentMode)} -> ${formatMode(issue.expectedMode)}`;
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in c245be5. `repairPermissions()` now repairs directory issues first (sequentially), then file issues in parallel. This prevents EACCES failures when the parent directory lacks execute permission at the time child file chmods run.

PR #237, opened 2026-02-12
feat(project): add `project create` command

betegon:
## Summary

Adds `sentry project create <name> <platform>` — the first write command in the CLI. Follows `gh repo create` conventions with org/name syntax, auto-detection of org and team, and actionable errors at every step.

## Changes

- New `SentryTeam` Zod schema and type
- `listTeams()` and `createProject()` API functions with region-aware routing (`TEAM_ENDPOINT_REGEX`)
- `project create` command with two required positionals, `--team`/`-t` flag, and `--json` output
- Team auto-selection when org has exactly one team; lists available teams otherwise
- Org fallback: on 404 from bad auto-detected org, shows user's actual organizations
- DSN fetched after creation (best-effort) so the user can start sending events immediately
- 19 tests covering happy path, missing args, team resolution, conflict handling, and JSON output

## Test Plan

- `bun test test/commands/project/create.test.ts` — 19 pass, 0 fail
- `bun run typecheck` — clean
- `bun run lint` — clean
- Manually tested against live API: create with auto-team, explicit team, bad team, 409 conflict, missing args, JSON mode

### DIFF
--- a/src/lib/resolve-team.ts
+++ b/src/lib/resolve-team.ts
@@ -202,7 +202,7 @@ export async function buildOrgNotFoundError(
     throw new ResolutionError(
       `Organization '${orgSlug}'`,
       `not found (did you mean '${effectiveOrg}'?)`,
-      `${usageHint} --team <team-slug>`,
+      usageHint,
       [`Try using '${effectiveOrg}' as the org slug instead of '${orgSlug}'`]
     );
   }
@@ -220,7 +220,7 @@ export async function buildOrgNotFoundError(
   throw new ResolutionError(
     `Organization '${orgSlug}'`,
     "not found",
-    `${usageHint} --team <team-slug>`,
+    usageHint,
     suggestions
   );
 }


src/commands/project/create.ts
```diff
@@ -0,0 +1,271 @@
...
+      }
+      if (error.status === 400 && isPlatformError(error)) {
+        throw new CliError(buildPlatformError(`${orgSlug}/${name}`, platform));
+      }
+      if (error.status === 404) {
+        throw new CliError(
+          `Team '${teamSlug}' not found in ${orgSlug}.\n\n` +
+            "Check the team slug and try again:\n" +
+            `  sentry project create ${orgSlug}/${name} ${platform} --team <team-slug>`
+        );
```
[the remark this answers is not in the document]
betegon: weird edge case. added a differentiation to be more explicit: 
Screenshot 2026-02-13 at 19 31 43

src/lib/arg-parsing.ts
```diff
@@ -210,6 +210,57 @@ export function parseOrgProjectArg(arg: string | undefined): ParsedOrgProject {
...
+ * @throws {ContextError} If slash is present but org or name is empty
+ *
+ * @example
+ * parseOrgPrefixedArg("my-app", "Project name", "sentry project create <org>/<name>")
+ * // { name: "my-app" }
+ *
+ * parseOrgPrefixedArg("acme/my-app", "Project name", "sentry project create <org>/<name>")
+ * // { org: "acme", name: "my-app" }
+ */
+export function parseOrgPrefixedArg(
```
BYK: I'm doing a massive refactor which may make this part obsolete.

src/lib/resolve-team.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * When listTeams fails (e.g., bad org slug from auto-detection), the error
+ * includes the user's actual organizations so they can fix the command.
+ *
+ * @param orgSlug - Organization to list teams from
+ * @param options - Resolution options (team flag, usage hint, detection source)
+ * @returns Team slug to use
+ * @throws {ContextError} When team cannot be resolved
+ */
+export async function resolveTeam(
```
BYK: Same as above

test/commands/project/create.test.ts
```diff
@@ -0,0 +1,472 @@
...
+
+    // listTeams should NOT be called when --team is explicit
+    expect(listTeamsSpy).not.toHaveBeenCalled();
+    expect(createProjectSpy).toHaveBeenCalledWith("acme-corp", "mobile", {
+      name: "my-app",
+      platform: "go",
+    });
+  });
+
+  test("errors when multiple teams exist without --team", async () => {
```
BYK: I think we select _some_ team automatically based on the teams the user is a member of or _something_ instead of requiring a manual selection. I wonder if we can do this here too. Otherwise we cannot use `project create` without passing an explicit `--team` flag in any non-trivial org as they will 100% have 1+ teams

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+ *
+ * @example slugify("My Cool App") // "my-cool-app"
+ * @example slugify("my-app")      // "my-app"
+ */
+function slugify(name: string): string {
+  return name
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, "-")
+    .replace(/^-|-$/g, "");
+}
```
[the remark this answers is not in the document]
betegon: - both uses of slugify are advisory hints, not functional logic. The `409` path shows a `project view` command the user can copy, worst case if the slug diverges from Sentry's, it `404`s and they run project list instead.
- the divergence note always shows the actual assigned slug; only the "already taken" explanation could be slightly off for rare unicode names. 

No project gets "mislinked", a wrong guess just doesn't resolve. I dont think we should be adding complexity for this.

src/lib/arg-parsing.ts
betegon: finally looked through the current infra after merging main, `parseOrgProjectArg` and `dispatchOrgScopedList` are for list/query commands. `parseOrgPrefixedArg` parses a new resource name, not an existing slug, and `resolveTeam` handles team auto-selection for write operations.

So i don't think they have an equivalent in your changes, maybe we could consolidate later with more `create` commands?

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+/** Usage hint template — base command without positionals */
+const USAGE_HINT = "sentry project create <org>/<name> <platform>";
+
+type CreateFlags = {
+  readonly team?: string;
+  readonly json: boolean;
+};
+
+/** Common Sentry platform strings, shown when platform arg is missing or invalid */
+const PLATFORMS = [
```
BYK, line 63: This list is too narrow. We should _at least_ be using https://github.com/getsentry/sentry/blob/411b4c5754e608c06754a3695a2630a23e6a6851/src/sentry/models/project.py#L64-L173

But ideally we fetch it at runtime from https://release-registry.services.sentry.io/sdks

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+  "elixir",
+] as const;
+
+/**
+ * Convert a project name to its expected Sentry slug.
+ * Sentry slugs are lowercase, with non-alphanumeric runs replaced by hyphens.
+ *
+ * @example slugify("My Cool App") // "my-cool-app"
+ * @example slugify("my-app")      // "my-app"
+ */
```
BYK, line 101: Do we have this logic coded somewhere in the Sentry app? We should and if we do, this comment should refer to that specific location.

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+  const heading = platform
+    ? `Invalid platform '${platform}'.`
+    : "Platform is required.";
+
+  return (
+    `${heading}\n\n` +
+    "Usage:\n" +
+    `  sentry project create ${nameArg} <platform>\n\n` +
+    `Available platforms:\n\n${list}\n\n` +
+    "Full list: https://docs.sentry.io/platforms/"
```
BYK: See my earlier comment: we should list the ones from the release regists.

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+        `Team '${teamSlug}' not found in ${orgSlug}.\n\n` +
+          `Available teams:\n\n${teamList}\n\n` +
+          "Try:\n" +
+          `  sentry project create ${orgSlug}/${name} ${platform} --team <team-slug>`
+      );
+    }
+    throw new CliError(
+      `No teams found in ${orgSlug}.\n\n` +
+        "Create a team first, then try again."
+    );
```
BYK, line 195: Can we offer creating a new, default team on the fly? Or actually "just do it"?

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+    }
+    throw error;
+  }
+}
+
+/**
+ * Write key-value pairs with aligned columns.
+ * Used for human-readable output after resource creation.
+ */
+function writeKeyValue(
```
BYK: This should be a shared utility across all commands

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+}
+
+export const createCommand = buildCommand({
+  docs: {
+    brief: "Create a new project",
+    fullDescription:
+      "Create a new Sentry project in an organization.\n\n" +
+      "The name supports org/name syntax to specify the organization explicitly.\n" +
+      "If omitted, the org is auto-detected from config defaults or DSN.\n\n" +
```
BYK: It's a bit confusing to me to detect anything form the DSN under this context as if we are creating a new project, that implies we don't typically have a DSN around?

src/commands/project/create.ts
```diff
@@ -0,0 +1,368 @@
...
+      "Project name",
+      USAGE_HINT
+    );
+
+    // Resolve organization
+    const resolved = await resolveOrg({ org: explicitOrg, cwd });
+    if (!resolved) {
+      throw new ContextError("Organization", USAGE_HINT, [
+        `Include org in name: ${USAGE_HINT}`,
+        "Set a default: sentry org view <org>",
```
BYK: This is not how we set a default org. In fact we don't have the notion of a default org but I like it. Good follow up PR.

src/lib/api-client.ts
```diff
@@ -700,6 +710,37 @@ export async function listTeams(orgSlug: string): Promise<SentryTeam[]> {
...
+ */
+export async function createProject(
+  orgSlug: string,
+  teamSlug: string,
+  body: CreateProjectBody
+): Promise<SentryProject> {
+  const regionUrl = await resolveOrgRegion(orgSlug);
+  const { data } = await apiRequestToRegion<SentryProject>(
+    regionUrl,
+    `/teams/${orgSlug}/${teamSlug}/projects/`,
```
BYK: Is the reason to not use `@sentry/api` things because they are not region-aware? If that's the case I think we should add the region-aware version to that package @MathurAditya724

src/lib/arg-parsing.ts
```diff
@@ -235,6 +235,59 @@ export function parseOrgProjectArg(arg: string | undefined): ParsedOrgProject {
   return { type: "project-search", projectSlug: trimmed };
 }
 
+/** Parsed result from an `org/name` positional argument */
+export type ParsedOrgPrefixed = {
```
BYK: I'm 100% sure we have something better than this introduced in https://github.com/getsentry/cli/pull/262/changes (or already existing)

src/commands/project/create.ts
betegon: my thinking here was to show just a small list of the most used/popular ones If we show the 140, that would be way harder to read and fills all the terminal screen

src/commands/project/create.ts
betegon: found it 🅰️ 🅰️ 🅰️ 🅰️

src/commands/project/create.ts
betegon: created an issue for this https://github.com/getsentry/cli/issues/303

src/lib/arg-parsing.ts
betegon: you're right. started this branch before we merged that one 😬 . just added it.

src/lib/api-client.ts
betegon: @MathurAditya724 solved this, thanks!

src/commands/project/create.ts
betegon: https://github.com/getsentry/cli/issues/304

src/lib/api-client.ts
```diff
@@ -779,6 +825,52 @@ export function listRepositoriesPaginated(
...
+ *
+ * @param orgSlug - The organization slug
+ * @returns Array of teams in the organization
+ */
+export function listTeams(orgSlug: string): Promise<SentryTeam[]> {
+  return orgScopedRequest<SentryTeam[]>(`/organizations/${orgSlug}/teams/`, {
+    params: { detailed: "0" },
+    schema: z.array(SentryTeamSchema),
+  });
+}
```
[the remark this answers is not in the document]
betegon: I KNOW

src/lib/api-client.ts
```diff
@@ -779,6 +825,52 @@ export function listRepositoriesPaginated(
...
+): Promise<SentryProject> {
+  return orgScopedRequest<SentryProject>(
+    `/teams/${orgSlug}/${teamSlug}/projects/`,
+    {
+      method: "POST",
+      body,
+      schema: SentryProjectSchema,
+    }
+  );
+}
```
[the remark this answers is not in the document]
betegon: I KNOW

src/commands/project/create.ts
```diff
@@ -0,0 +1,387 @@
...
+  readonly json: boolean;
+};
+
+/** Common Sentry platform strings, shown when platform arg is missing or invalid */
+const PLATFORMS = [
+  "javascript",
+  "javascript-react",
+  "javascript-nextjs",
+  "javascript-vue",
+  "javascript-angular",
```
BYK, line 68: Where are we getting these platform names from? I thought it would match the names from https://release-registry.services.sentry.io/sdks which seem to be in the format `javascript.react` not `javascript-react`?

src/commands/project/create.ts
```diff
@@ -0,0 +1,387 @@
...
+  const heading = platform
+    ? `Invalid platform '${platform}'.`
+    : "Platform is required.";
+
+  return (
+    `${heading}\n\n` +
+    "Usage:\n" +
+    `  sentry project create ${nameArg} <platform>\n\n` +
+    `Available platforms:\n\n${list}\n\n` +
+    "Full list: https://docs.sentry.io/platforms/"
```
BYK: This link is not helpful as it doesn't show you the actual values you can use in the CLI?

src/commands/project/create.ts
```diff
@@ -0,0 +1,387 @@
...
+  );
+}
+
+/**
+ * Disambiguate a 404 from the create project endpoint.
+ *
+ * The `/teams/{org}/{team}/projects/` endpoint returns 404 for both
+ * a bad org and a bad team. This helper calls `listTeams` to determine
```
BYK, line 145: I added quick and mostly offline org validation here: https://github.com/getsentry/cli/pull/316

We should probably just leverage that here

src/commands/project/create.ts
```diff
@@ -0,0 +1,387 @@
...
+          "Try:\n" +
+          `  sentry project create ${orgSlug}/${name} ${platform} --team <team-slug>`
+      );
+    }
+    throw new CliError(
+      `No teams found in ${orgSlug}.\n\n` +
+        "Create a team first, then try again."
+    );
+  }
+
```
BYK, line 197: We should document our expected flow in the comments here. And I'd say instead of requiring the user to find a relevant team upfront we should help them get started with minimum friction if possible. That means:

1. If there's only one team in the org, just pick that
2. If there are no teams in the org, just create one with the project name, add the author and move on
3. For the case of multiple teams existing but the user not passing `--team` flag, we should do what the existing Sentry UI does and let the user know at the end which team we used and how they can change that.

src/commands/project/create.ts
```diff
@@ -376,6 +399,19 @@ export const createCommand = buildCommand({
       );
     }
 
+    // Inform user which team was used when not explicitly specified
+    if (team.source === "auto-created") {
+      stdout.write(
+        `Note: Created team '${team.slug}' (org had no teams). ` +
+          `Change with: sentry project create ${nameArg} ${platformArg} --team <slug>\n`
```
BYK: This would not change the team but would create a new project. I'd just omit this message from auto-created flow.

src/commands/project/create.ts
```diff
@@ -376,6 +399,19 @@ export const createCommand = buildCommand({
...
+      );
+    } else if (team.source === "auto-selected") {
+      stdout.write(
+        `Note: Using team '${team.slug}'. ` +
+          `Change with: sentry project create ${nameArg} ${platformArg} --team <slug>\n`
```
BYK: You cannot change with that command. Maybe we can hint at `sentry team list` command to help them see existing teams?

src/commands/project/create.ts
```diff
@@ -112,7 +134,7 @@ function buildPlatformError(nameArg: string, platform?: string): string {
     "Usage:\n" +
     `  sentry project create ${nameArg} <platform>\n\n` +
     `Available platforms:\n\n${list}\n\n` +
-    "Full list: https://docs.sentry.io/platforms/"
+    "Run 'sentry project create <name> <platform>' with any valid Sentry platform identifier."
```
BYK, line 137: I wonder if we could get the platforms we support from https://release-registry.services.sentry.io/sdks, "slugify" them (strip the `sentry.` prefix then convert `.` to `-` I guess? Check the code in Sentry repo) and provide an auto-complete list in the shell if it is interactive? or show a tree-view if non interactive?

src/lib/api-client.ts
```diff
@@ -710,6 +711,27 @@ export async function createProject(
...
+ *
+ * @param orgSlug - The organization slug
+ * @param slug - Team slug (also used as display name)
+ * @returns The created team
+ */
+export async function createTeam(
+  orgSlug: string,
+  slug: string
+): Promise<SentryTeam> {
+  const config = await getOrgSdkConfig(orgSlug);
```
BYK, line 733: Where are the members of the team?

src/lib/resolve-team.ts
```diff
@@ -38,31 +54,36 @@ export type ResolveTeamOptions = {
...
- * includes the user's actual organizations so they can fix the command.
- *
  * @param orgSlug - Organization to list teams from
  * @param options - Resolution options (team flag, usage hint, detection source)
- * @returns Team slug to use
+ * @returns Resolved team slug with source info
  * @throws {ContextError} When team cannot be resolved
  * @throws {ResolutionError} When org slug returns 404
  */
 export async function resolveTeam(
```
BYK: The name is a bit misleading. It should be `resolveOrCreateTeam()` I'd say?

src/lib/resolve-team.ts
```diff
@@ -83,20 +104,32 @@ export async function resolveTeam(
     throw error;
   }
 
+  // No teams — auto-create one if a slug was provided
   if (teams.length === 0) {
+    if (options.autoCreateSlug) {
+      return await autoCreateTeam(orgSlug, options.autoCreateSlug);
```
BYK, line 120: We should add the current user as a member of the newly created team

src/lib/resolve-team.ts
```diff
@@ -83,20 +104,32 @@ export async function resolveTeam(
...
 
   if (candidates.length === 1) {
-    return (candidates[0] as SentryTeam).slug;
+    return {
+      slug: (candidates[0] as SentryTeam).slug,
+      source: "auto-selected",
+    };
   }
 
   // Multiple candidates — user must specify
```
BYK, line 145: Is this the experience we have on the Sentry SaaS UI?

src/lib/arg-parsing.ts
BYK: Agreed — `parseOrgPrefixedArg` serves write operations (creating new resources) while `parseOrgProjectArg`/`dispatchOrgScopedList` handle read/list operations. Happy to consolidate later when more `create` commands land.

src/commands/team/view.ts
```diff
@@ -0,0 +1,236 @@
...
+    const { targets } = await resolveAllTargets({ cwd });
+    if (targets.length > 0 && targets[0]) {
+      return targets[0].org;
+    }
+  } catch {
+    // Fall through — DSN detection is best-effort
+  }
+
+  return null;
+}
```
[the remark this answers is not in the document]
BYK: Valid finding. `resolveOrgForTeam` in `team/view.ts` is pre-existing code not changed in this PR — filing a follow-up to refactor it to use the shared `resolveOrg` from `resolve-target.ts`.

src/commands/project/create.ts
BYK: These come from Sentry's internal platform registry (`sentry/utils/platform_categories.py`), which uses hyphen-separated format (`javascript-react`). The release registry (`sentry.javascript.react`) uses a different namespace convention. The API endpoint `POST /teams/{org}/{team}/projects/` validates against the internal platform list, so we must use the hyphen format.

Added a comment in the code documenting this (see the JSDoc on `PLATFORMS`). For a follow-up, we can look at fetching from the release registry and converting formats — see Thread 12 reply.

src/commands/project/create.ts
BYK: Fixed in `59ce9a0` — replaced the docs.sentry.io/platforms link with an inline usage hint that shows the actual command syntax:

```
Run 'sentry project create <name> <platform>' with any valid Sentry platform identifier.
```

The `buildPlatformError` function now shows the curated platform list inline.

src/commands/project/create.ts
BYK: Done in `59ce9a0` — `resolve-team.ts` now uses `resolveEffectiveOrg` from `region.ts` (which is the same utility from PR #316) for offline org validation when `listTeams` returns 404. See the `buildOrgFailureError` function.

src/commands/project/create.ts
BYK: All three cases implemented in `59ce9a0` + `9e927d0`:

1. **Single team** → auto-select it (step 4 in `resolveOrCreateTeam`)
2. **Zero teams** → auto-create team with project slug, add current user as member via `addMemberToTeam("me")` (step 3)
3. **Multiple teams** → filter to `isMember` teams first; if still multiple, error with team list and `--team` hint (steps 5-6)

Auto-selected/auto-created teams show a note in the output:
- Auto-created: `Note: Created team 'my-app' (org had no teams).`
- Auto-selected: `Note: Using team 'engineering'. See all teams: sentry team list`

The full flow is documented in the JSDoc at the top of `resolve-team.ts`.

src/commands/project/create.ts
```diff
@@ -0,0 +1,423 @@
...
+        throw new CliError(
+          `A project named '${name}' already exists in ${orgSlug}.\n\n` +
+            `View it: sentry project view ${orgSlug}/${slug}`
+        );
+      }
+      if (error.status === 400 && isPlatformError(error)) {
+        throw new CliError(buildPlatformError(`${orgSlug}/${name}`, platform));
+      }
+      if (error.status === 404) {
+        return handleCreateProject404(orgSlug, teamSlug, name, platform);
```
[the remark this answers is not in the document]
BYK: Fixed in `9e927d0` — added `await` before `return handleCreateProject404(...)` to preserve the stack frame per the project's lore entry.

src/lib/resolve-team.ts
```diff
@@ -0,0 +1,210 @@
...
+  options: ResolveTeamOptions
+): Promise<never> {
+  // Try resolving DSN-style org IDs (e.g., o1081365 → actual slug)
+  const effectiveOrg = await resolveEffectiveOrg(orgSlug);
+  if (effectiveOrg !== orgSlug) {
+    throw new ResolutionError(
+      `Organization '${orgSlug}'`,
+      `not found (did you mean '${effectiveOrg}'?)`,
+      `${options.usageHint} --team <team-slug>`,
+      [`Try: ${options.usageHint.replace(orgSlug, effectiveOrg)}`]
```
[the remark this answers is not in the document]
BYK: Fixed in `9e927d0` — replaced the `.replace(orgSlug, effectiveOrg)` no-op with direct string interpolation:

```typescript
[`Try using '${effectiveOrg}' as the org slug instead of '${orgSlug}'`]
```

src/lib/resolve-team.ts
```diff
@@ -0,0 +1,210 @@
...
+  } catch (error) {
+    // If team creation fails (permissions, etc.), fall back to error
+    throw new CliError(
+      `No teams found in org '${orgSlug}' and automatic team creation failed.\n\n` +
+        `Create a team manually at ${getSentryBaseUrl()}/settings/${orgSlug}/teams/` +
+        (error instanceof ApiError
+          ? `\n\nAPI error (${error.status}): ${error.detail ?? error.message}`
+          : "")
+    );
+  }
```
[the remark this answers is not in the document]
BYK: Fixed in `9e927d0` — added `AuthError` re-throw before the generic catch in `autoCreateTeam`:

```typescript
if (error instanceof AuthError) {
  throw error;
}
```

This preserves the auto-login flow from the central error handler.

src/commands/project/create.ts
BYK: Fixed in `9e927d0` — the auto-created team note now just says:

```
Note: Created team 'my-app' (org had no teams).
```

No misleading "Change with" command.

src/commands/project/create.ts
BYK: Fixed in `9e927d0` — the auto-selected team note now says:

```
Note: Using team 'engineering'. See all teams: sentry team list
```

Instead of the incorrect "Change with" command.

src/commands/project/create.ts
BYK: Great idea! This would be a nice follow-up. The plan:

1. Fetch from `release-registry.services.sentry.io/sdks`
2. Strip `sentry.` prefix, convert `.` → `-` to match the internal platform format
3. Cache the result (SQLite or in-memory)
4. Interactive: auto-complete / fuzzy search
5. Non-interactive: tree-view grouped by language family

For now, keeping the curated `PLATFORMS` list as a starting point — it covers the most common platforms and the help text mentions this is a subset. I'll file a follow-up issue for the release-registry integration.

src/lib/api-client.ts
BYK: Fixed in `9e927d0` — `createTeam` now calls `addMemberToTeam(orgSlug, team.slug, "me")` after creating the team to add the current user as a member. The member-add is best-effort (caught errors don't fail the team creation).

New `addMemberToTeam()` function uses the `addAnOrganizationMemberToATeam` SDK endpoint with `"me"` as the member_id.

src/lib/resolve-team.ts
BYK: Done in `9e927d0` — renamed to `resolveOrCreateTeam()`. Updated import in `create.ts` and test comments.

src/lib/resolve-team.ts
BYK: Done in `9e927d0` — `createTeam` in `api-client.ts` now calls `addMemberToTeam(orgSlug, team.slug, "me")` after team creation. Uses the `addAnOrganizationMemberToATeam` SDK endpoint with `"me"` as member_id (Sentry API shorthand for the authenticated user). Best-effort — errors don't fail the team creation.

src/lib/resolve-team.ts
BYK: Yes — the Sentry SaaS project creation wizard always requires a team. If the org has no teams, the UI auto-creates one (typically named after the first user or the org). Our flow mirrors this:

1. Zero teams → auto-create with the project name as slug, add the current user
2. Single team → auto-select
3. Multiple teams → ask user to specify (SaaS shows a dropdown)

The key difference is we use the project name as the auto-created team slug (since we don't have an interactive team name prompt), while the SaaS UI may use a different default. The end result is the same: the user gets a project without needing to manually create a team first.

src/lib/api-client.ts
```diff
@@ -729,7 +735,40 @@ export async function createTeam(
...
   });
   const data = unwrapResult(result, "Failed to create team");
-  return data as unknown as SentryTeam;
+  const team = data as unknown as SentryTeam;
+
+  // Best-effort: add the current user to the team
+  try {
+    await addMemberToTeam(orgSlug, team.slug, "me");
+  } catch {
+    // Non-fatal — team was created, user just won't be a member
```
BYK: We should capture this with Sentry and possibly let the user know.

src/commands/project/create.ts
BYK: File a follow up issue for this.

src/commands/project/create.ts
```diff
@@ -0,0 +1,420 @@
...
+    throw new CliError(`Organization '${orgSlug}' not found.\n\n${orgHint}`);
+  }
+
+  // listTeams failed for other reasons (403, 5xx, network) — can't disambiguate
+  throw new CliError(
+    `Failed to create project '${name}' in ${orgSlug}.\n\n` +
+      "The organization or team may not exist, or you may lack access.\n\n" +
+      "Try:\n" +
+      `  sentry project create ${orgSlug}/${name} ${platform} --team <team-slug>`
+  );
```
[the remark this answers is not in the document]
BYK: Fixed in 4811560. Extracted `buildOrgNotFoundError` from `resolve-team.ts` as a shared export, and `handleCreateProject404` now delegates to it instead of duplicating the logic. Both paths now get `resolveEffectiveOrg` DSN resolution, `detectedFrom` context, and throw `ResolutionError` consistently.

src/commands/project/create.ts
```diff
@@ -0,0 +1,420 @@
...
+      name,
+      platformArg
+    );
+
+    // Fetch DSN (best-effort)
+    const dsn = await tryGetPrimaryDsn(orgSlug, project.slug);
+
+    // JSON output
+    if (flags.json) {
+      writeJson(stdout, { ...project, dsn, team: team.slug });
```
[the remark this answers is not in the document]
BYK: Fixed in 4811560. Renamed `team: team.slug` to `teamSlug: team.slug` so spreading the project response preserves the API's `team` object (which has `id`/`name`/`slug` properties). Added test assertion for `parsed.teamSlug`.

src/lib/api-client.ts
BYK: Fixed in 4811560. The catch block now calls `Sentry.captureException(error, { extra: { orgSlug, teamSlug, context } })` and writes a warning to stderr: `Warning: Team '<slug>' was created but you could not be added as a member.`

src/commands/project/create.ts
```diff
@@ -0,0 +1,418 @@
...
+  }
+
+  // listTeams failed for other reasons (403, 5xx, network) — can't disambiguate
+  throw new CliError(
+    `Failed to create project '${name}' in ${orgSlug}.\n\n` +
+      "The organization or team may not exist, or you may lack access.\n\n" +
+      "Try:\n" +
+      `  sentry project create ${orgSlug}/${name} ${platform} --team <team-slug>`
+  );
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 87e7e44. `resolved.detectedFrom` is now threaded through `createProjectWithErrors` → `handleCreateProject404` → `buildOrgNotFoundError`. Both functions refactored to use options objects to stay within the 4-parameter lint limit.

src/lib/resolve-team.ts
```diff
@@ -0,0 +1,226 @@
...
+  // Try resolving DSN-style org IDs (e.g., o1081365 → actual slug)
+  const effectiveOrg = await resolveEffectiveOrg(orgSlug);
+  if (effectiveOrg !== orgSlug) {
+    throw new ResolutionError(
+      `Organization '${orgSlug}'`,
+      `not found (did you mean '${effectiveOrg}'?)`,
+      `${usageHint} --team <team-slug>`,
+      [`Try using '${effectiveOrg}' as the org slug instead of '${orgSlug}'`]
+    );
+  }
```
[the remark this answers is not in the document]
BYK: Fixed in a22bf9e. Removed the `--team <team-slug>` suffix from the usage hint in `buildOrgNotFoundError` — when the org is wrong, suggesting `--team` is misleading. The base usage hint (e.g., `sentry project create <org>/<name> <platform>`) is now shown as-is.

src/commands/project/create.ts
```diff
@@ -0,0 +1,423 @@
...
+
+/**
+ * Check whether an API error is about an invalid platform value.
+ * Relies on Sentry's error message wording — may need updating if the API changes.
+ */
+function isPlatformError(error: ApiError): boolean {
+  const detail = error.detail ?? error.message;
+  return detail.includes("platform") && detail.includes("Invalid");
+}
```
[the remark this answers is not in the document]
BYK: Acknowledged — the string matching is intentionally pragmatic here. The Sentry API returns `"Invalid platform"` for bad platform values, and there's no structured error code to match on. The risk of a false positive (another 400 error containing both `"platform"` AND `"Invalid"` as substrings) is very low in practice, and the cost is just showing a platform suggestion list instead of a raw API error. If the API adds structured error codes in the future, we can upgrade. Resolving as accepted risk.

Issue #255, opened 2026-02-17
Publish via NuGet.org as a .NET tool


Flash0ver:
### Summary

Publish the [Sentry CLI](https://cli.sentry.dev/) via [NuGet.org](NuGet.org) as a [.NET tool](https://learn.microsoft.com/dotnet/core/tools/global-tools).

See also #87.

### Remarks

We want to decouple the release of the Sentry CLI for NuGet from the release of the Sentry .NET SDK,
so we can't reuse the name `Sentry`, which is the "base" .NET SDK.
A lot of tools in the ecosystem prefix their tools with `dotnet-`, so let's use `dotnet-sentry` for the ID of the package.

Support .NET SDK 10.0+: The .NET 10 SDK has introduced `<DotNetCliTool Version="2">` which also allows platform-specific .NET tools incompatible with .NET SDK 9.0-.

The Sentry CLI is currently released as these self-contained platform-specific executables:
| [Release Asset](https://github.com/getsentry/cli/releases/tag/0.10.2) | [.NET Runtime Identifier](https://learn.microsoft.com/dotnet/core/rid-catalog) |
| --- | --- |
| sentry-darwin-arm64 | osx-arm64 |
| sentry-darwin-x64 | osx-x64 |
| sentry-linux-arm64 | linux-arm64 |
| sentry-linux-x64 | linux-x64 |
| sentry-windows-x64.exe | win-x64 |

The "root" package serves as an "entry point" for installing the respective RID-specific tool.
The framework-dependent "any" RID just prints an error that the `{Current-Platform}` is not supported.
Each RID-specific tool contains the binary of the Sentry-CLI for the respective platform, which is launched by a .NET tool.
These RID-specific tools should be Native-AOT compiled for best startup-performance and lower memory footprints of short-lived processes.

### Context
- [What's new in .NET 10 SDK](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/sdk)
- [Create RID-specific, self-contained, and AOT .NET tools](https://learn.microsoft.com/en-us/dotnet/core/tools/rid-specific-tools)
- [Supporting platform-specific .NET tools on old .NET SDKs](https://andrewlock.net/exploring-dotnet-10-preview-features-8-supporting-platform-specific-dotnet-tools-on-old-sdks/)

PR #263, opened 2026-02-19
feat: Pack CLI as .NET tool in NuGet packages

Flash0ver:
Requires .NET SDK 10.0+.

See #255.


BYK:
Closing this for now as I don't think we can support this properly. Hopefully we can come back to this later in the future, especially when/if there's more demand.

Thank you so much @Flash0ver for giving this a shot and sorry we couldn't merge it :(

PR #262, opened 2026-02-18
feat(list): add pagination and consistent target parsing to all list commands

BYK:
## Summary

Extends the pagination and flexible target parsing pattern introduced in #221 for \`project list\` to all remaining list commands: \`issue list\`, \`team list\`, \`repo list\`, \`trace list\`, and \`log list\`. Introduces a shared \`dispatchOrgScopedList\` infrastructure with \`HandlerContext\` that all list commands now use.

### Shared infrastructure

- **\`src/lib/arg-parsing.ts\`**: Added \`validateLimit(value, min, max)\` — shared limit validation, replaces duplicate implementations in \`trace/list.ts\` and \`log/list.ts\`
- **\`src/lib/resolve-target.ts\`**: Added \`resolveOrgsForListing()\`, \`resolveOrgProjectTarget()\`, and \`resolveOrgProjectFromArg()\` — deduplicates logic previously copied across commands
- **\`src/lib/api-client.ts\`**: Added \`listIssuesPaginated()\`, \`listTeamsPaginated()\`, \`listRepositoriesPaginated()\`, \`listProjectTeams()\` using \`orgScopedRequestPaginated()\` for cursor-based pagination with Link header support
- **\`src/lib/org-list.ts\`**: Shared dispatch infrastructure — \`ListCommandMeta\`, \`OrgListConfig\`, \`HandlerContext\`, \`ModeHandler\`, \`ModeHandlerMap\`, \`dispatchOrgScopedList\`. Handlers receive a \`HandlerContext\` with \`{ parsed, stdout, cwd, flags }\` so they don't need to close over those values; commands add extras via spread: \`(ctx) => handle({ ...ctx, extra })\`
- **\`src/lib/list-command.ts\`**: Shared Stricli parameter constants and \`buildOrgListCommand\` factory for simple org-scoped commands (team, repo)
- **\`src/lib/db/pagination.ts\`**: Added \`escapeContextKeyValue()\` for safe pipe-delimiter escaping in context keys, and \`resolveOrgCursor()\` shared across all list commands

### Command changes

| Command | What changed |
|---|---|
| \`issue list\` | Added \`--cursor/-c\` flag; org-all mode now uses \`listIssuesPaginated\` and returns \`{ data, hasMore, nextCursor? }\` in JSON output |
| \`team list\` | Full rewrite: 4-mode handler via \`dispatchOrgScopedList\`, \`--cursor/-c\` flag, project-scoped team listing via \`listProjectTeams\` for \`org/project\` mode |
[...]

PR #271, opened 2026-02-23
fix(help): document target patterns and trailing-slash significance (#267)

### LINKED ISSUE #267: issue list help text shows opaque <target> — should document supported patterns
## Problem

The usage line for `sentry issue list` currently reads:

```
USAGE
  sentry issue list [--query value] [--limit value] [--sort value] [--json] [--cursor value] [<target>]
```

`<target>` is opaque. Users don't know that multiple patterns are supported, or that the trailing `/` on `org/` is significant syntax that changes behavior (enables cursor pagination, switches to org-wide listing).

A user recently tried `sentry issue list sentry --cursor 100` expecting org-scoped pagination, but `sentry` (without trailing slash) is parsed as a project-search, which rejects `--cursor`. The error message mentions `<org>/` but the significance of the trailing slash doesn't register.

## Supported patterns today

| Input | Mode | Behavior |
|---|---|---|
| _(none)_ | auto-detect | DSN detection / config defaults |
| `org/project` | explicit | Single org and project |
| `org/` | org-all | All projects in org (only mode supporting `--cursor`) |
| `/project` | project-search | Find project across all orgs |
| `project` | project-search | Find project across all orgs |

## Ideas

- Show `[<org>/<project>]` in the usage line instead of `[<target>]`
- Document the wildcard patterns somewhere visible — a table in `fullDescription` that `--help` renders, or similar
- Make the trailing `/` significance more obvious in error messages when `--cursor` is rejected
- Consider whether a bare string + `--cursor` should be interpreted as org-all mode rather than erroring

## Context

Ref: user report where all three of `sentry issue list sentry`, `sentry issue list 1`, and `sentry issue list javascript` were tried with `--cursor` and all produced the same confusing error.

BYK:
Fixes #267.

## Problem

The `[<target>]` placeholder in usage lines was opaque — users had no indication that multiple patterns are supported or that the trailing `/` on `org/` is significant syntax that changes behaviour (switches from project-search to org-all, the only mode that accepts `--cursor`).

## Changes

**Five commands updated** (`issue/list`, `project/list`, `trace/list`, `log/list`, `project/view`):

- **Placeholder rename**: `<target>` → `<org/project>` so the usage line is self-documenting without reading `--help`
- **Trailing-slash explanation**: `issue/list` and `project/list` get a prose paragraph with a concrete example (`sentry` searches for a project named "sentry"; `sentry/` lists all projects in the "sentry" org). The other three commands get a shorter note.
- **`project/list` cursor error improvement**: when `--cursor` is rejected in project-search mode, the error now includes a contextual hint — *"Did you mean `sentry project list sentry/`? A bare name searches for a project — add a trailing slash to list an org's projects."*
- Section heading renamed from "Target specification" to "Target patterns"

### DIFF
--- a/src/commands/issue/list.ts
+++ b/src/commands/issue/list.ts
@@ -385,11 +385,15 @@ export const listCommand = buildCommand({
     brief: "List issues in a project",
     fullDescription:
       "List issues from Sentry projects.\n\n" +
-      "Target specification:\n" +
+      "Target patterns:\n" +
       "  sentry issue list               # auto-detect from DSN or config\n" +
       "  sentry issue list <org>/<proj>  # explicit org and project\n" +
-      "  sentry issue list <org>/        # all projects in org\n" +
+      "  sentry issue list <org>/        # all projects in org (trailing / required)\n" +
       "  sentry issue list <project>     # find project across all orgs\n\n" +
+      "The trailing slash on <org>/ is significant — without it, the argument\n" +
+      "is treated as a project name search (e.g., 'sentry' searches for a\n" +
+      "project named 'sentry', while 'sentry/' lists all projects in the\n" +
+      "'sentry' org).\n\n" +
       "In monorepos with multiple Sentry projects, shows issues from all detected projects.",
   },
   parameters: {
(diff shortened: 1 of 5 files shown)


BYK:
Superseded by #272 — rebased cleanly on main after the #262 infrastructure refactor.

PR #277, opened 2026-02-23
feat(release): add Homebrew install support

### LINKED ISSUE #228: Add homebrew install option

BYK:
Closes #228

## What

Adds `brew install getsentry/tools/sentry` as a supported install method.

## Changes

**`.craft.yml`** — new `brew` target that publishes `Formula/sentry.rb` to the `getsentry/homebrew-tools` tap on each stable release. Uses `includeNames` to filter to the 4 macOS/Linux binaries only (no `.gz`, no Windows, no npm tarball). Checksum template keys are the raw filenames since they contain no dots or version strings.

**`src/lib/upgrade.ts`** — adds `"brew"` as an `InstallationMethod`. Detection checks for `/Cellar/` in `process.execPath` (works for both Apple Silicon and Intel Homebrew prefixes). Upgrade runs `brew upgrade getsentry/tools/sentry`. Version pinning is intentionally unsupported since Homebrew manages versioning through the formula.

**Docs** — `getting-started.mdx` and `commands/cli/upgrade.md` updated to document the new install and upgrade paths.

### DIFF
--- a/src/commands/cli/upgrade.ts
+++ b/src/commands/cli/upgrade.ts
@@ -222,10 +222,9 @@ export const upgradeCommand = buildCommand({
       } finally {
         releaseLock(downloadResult.lockPath);
       }
-    } else {
+    } else if (method !== "brew") {
       // Package manager: binary already in place, just run setup.
-      // Always use execPath — storedInfo?.path could reference a stale
-      // binary from a different installation method.
+      // Skip brew — Homebrew's post_install hook already runs setup.
       await runSetupOnNewBinary(this.process.execPath, method, false);
     }
 


.craft.yml
```diff
@@ -17,3 +17,43 @@ targets:
...
+            sha256 "{{checksums.sentry-darwin-x64}}"
+          end
+        elsif OS.linux?
+          if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
+            url "https://github.com/getsentry/cli/releases/download/{{version}}/sentry-linux-arm64"
+            sha256 "{{checksums.sentry-linux-arm64}}"
+          elsif Hardware::CPU.intel? && Hardware::CPU.is_64_bit?
+            url "https://github.com/getsentry/cli/releases/download/{{version}}/sentry-linux-x64"
+            sha256 "{{checksums.sentry-linux-x64}}"
+          end
```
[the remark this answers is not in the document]
BYK: False positive. Mustache uses `.` as the **only** path separator — hyphens are treated as literal characters in key names, not operators. So `{{checksums.sentry-darwin-arm64}}` resolves `checksums["sentry-darwin-arm64"]` correctly.

Verified two ways:
1. Craft source (`src/utils/strings.ts`): `sanitizeObject` only transforms dots to `__`, leaving hyphens untouched. Checksum keys in `src/targets/brew.ts` are the raw filenames.
2. Live test against the exact Mustache version in Craft's `node_modules`:
```js
mustache.render('{{checksums.sentry-darwin-arm64}}', { checksums: { 'sentry-darwin-arm64': 'abc123' } })
// => 'abc123'
```

src/lib/upgrade.ts
```diff
@@ -298,7 +324,9 @@ export async function fetchLatestFromNpm(): Promise<string> {
 export function fetchLatestVersion(
   method: InstallationMethod
 ): Promise<string> {
-  return method === "curl" ? fetchLatestFromGitHub() : fetchLatestFromNpm();
+  return method === "curl" || method === "brew"
+    ? fetchLatestFromGitHub()
+    : fetchLatestFromNpm();
```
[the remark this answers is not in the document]
BYK: False positive. `fetchLatestVersion("brew")` querying GitHub releases is intentional and consistent with the curl install path — both use the same GitHub releases API. The function is used for `--check` mode and "already up to date" detection only; the actual upgrade is always delegated to `brew upgrade`, which installs whatever the formula specifies regardless of what version number we printed. There is no code path where we'd install a specific version via Homebrew (version pinning is explicitly blocked with an `unsupported_operation` error). The GitHub version is the best available signal for 'is there a newer release?' even if the formula update lags by a few minutes.

.craft.yml
```diff
@@ -17,3 +17,47 @@ targets:
...
+          else
+            raise "Unsupported Linux CPU architecture: #{Hardware::CPU.type} (only 64-bit arm and x86_64 are supported)"
+          end
+        else
+          raise "Unsupported operating system"
+        end
+
+        def install
+          bin.install Dir["sentry-*"].first => "sentry"
+        end
```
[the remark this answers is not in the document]
BYK: False positive. Homebrew's `bin.install` DSL method automatically sets the executable bit — it's not a plain file copy. From the Homebrew source: `bin.install` calls `install_p(src, dst)` which includes a `chmod +x` on the destination. This is standard Homebrew formula practice; virtually every formula that installs a pre-built binary uses `bin.install` without explicitly calling `chmod`. An explicit `chmod 0755` would be redundant.

src/commands/cli/upgrade.ts
```diff
@@ -157,7 +157,7 @@ export const upgradeCommand = buildCommand({
       method: {
         kind: "parsed",
         parse: parseInstallationMethod,
-        brief: "Installation method to use (curl, npm, pnpm, bun, yarn)",
+        brief: "Installation method to use (curl, brew, npm, pnpm, bun, yarn)",
```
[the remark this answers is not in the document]
BYK: False positive — this is the same concern as the 'Brew check uses GitHub latest' thread (already resolved). The GitHub release version is the best available signal for 'what is the latest version?', and it is also what the Homebrew tap formula tracks (Craft publishes the formula on each GitHub release). The messaging is intentional: we report the GitHub latest version as the target, and `brew upgrade` will install that version. In the rare case of a formula lag of a few minutes, the user's binary will match after the formula catches up. The alternative — showing no version info at all for brew upgrades — would be a worse UX.

src/lib/upgrade.ts
```diff
@@ -199,7 +225,14 @@ async function detectLegacyInstallationMethod(): Promise<InstallationMethod> {
...
  */
 export async function detectInstallationMethod(): Promise<InstallationMethod> {
-  // 1. Check stored info (fast path)
+  // Always check for Homebrew first — the stored install info may be stale
+  // (e.g. user previously had a curl install recorded, then switched to
+  // Homebrew). The realpath check is cheap and authoritative.
+  if (isHomebrewInstall()) {
+    return "brew";
+  }
+
```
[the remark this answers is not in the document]
BYK: False positive. This CLI is compiled to a native binary with Bun (`bun build --compile`). In a Bun-compiled executable, `process.execPath` resolves to the path of the compiled `sentry` binary itself — not a Node.js interpreter. This is explicitly documented in Bun's docs: https://bun.sh/docs/bundler/executables#standalone-executables. Using `process.execPath` here is intentional and correct; the comment on the same line explains the rationale ('storedInfo?.path could reference a stale binary from a different installation method').

src/commands/cli/upgrade.ts
```diff
@@ -177,6 +177,15 @@ export const upgradeCommand = buildCommand({
       throw new UpgradeError("unknown_method");
     }
 
+    // Homebrew manages versioning through the formula in the tap — the installed
+    // version is always whatever the formula specifies, not an arbitrary release.
+    if (method === "brew" && version) {
```
BYK, line 182: But do we run  the relevant `brew` command ourselves if a version is not passed in?

src/lib/upgrade.ts
```diff
@@ -162,13 +163,38 @@ async function isInstalledWith(pm: PackageManager): Promise<boolean> {
...
  * Legacy detection for existing installs that don't have stored install info.
  * Checks known curl install paths and package managers.
  *
  * @returns Detected installation method, or "unknown" if unable to determine
  */
 async function detectLegacyInstallationMethod(): Promise<InstallationMethod> {
+  // Check for Homebrew Cellar path before curl — both may match known bin dirs
+  if (isHomebrewInstall()) {
+    return "brew";
+  }
```
BYK: We never supported installing via `brew` up to this point. I don't think this change makes sense.

src/commands/cli/upgrade.ts
BYK: Yes. When no version is passed, `executeUpgrade("brew", target)` calls `executeUpgradeHomebrew()` which spawns `brew upgrade getsentry/tools/sentry` with `stdio: "inherit"` so the user sees brew's output directly. The version guard here only blocks the explicit version-pinning case (`sentry cli upgrade 1.2.3`) since Homebrew doesn't support that — the formula always installs whatever version is published in the tap.

src/lib/upgrade.ts
BYK: Good catch. Removed the `isHomebrewInstall()` check from `detectLegacyInstallationMethod` in 8e4166c — it was both conceptually wrong (legacy detection is for pre-existing installs, and brew was never supported before) and technically dead code (the parent `detectInstallationMethod` already checks `isHomebrewInstall()` first and returns `"brew"` before the legacy function is ever called).

.craft.yml
```diff
@@ -17,3 +17,47 @@ targets:
...
+        elsif OS.linux?
+          if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
+            url "https://github.com/getsentry/cli/releases/download/{{version}}/sentry-linux-arm64"
+            sha256 "{{checksums.sentry-linux-arm64}}"
+          elsif Hardware::CPU.intel? && Hardware::CPU.is_64_bit?
+            url "https://github.com/getsentry/cli/releases/download/{{version}}/sentry-linux-x64"
+            sha256 "{{checksums.sentry-linux-x64}}"
+          else
+            raise "Unsupported Linux CPU architecture: #{Hardware::CPU.type} (only 64-bit arm and x86_64 are supported)"
+          end
```
BYK, line 51: Is it not possible to tell brew to download the `.gz` versions and then expand them?

.craft.yml
```diff
@@ -17,3 +17,47 @@ targets:
...
+        else
+          raise "Unsupported operating system"
+        end
+
+        def install
+          bin.install Dir["sentry-*"].first => "sentry"
+        end
+
+        test do
+          assert_match version.to_s, shell_output("#{bin}/sentry --version").chomp
```
BYK, line 65: Shall we not run `sentry setup` after install?

.craft.yml
BYK: Good call! Switched to `.gz` artifacts in d292f6c. Homebrew auto-decompresses `.gz` URLs natively, so the formula now downloads compressed binaries (~62% smaller — e.g. linux-x64 drops from 103MB to 30MB). Updated `includeNames` to match `.gz` filenames, URLs to point to the compressed versions, and checksum keys to use the `__gz` suffix (Craft's `sanitizeObject` replaces `.` with `__` for Mustache compatibility).

.craft.yml
BYK: Added a `post_install` hook in d292f6c that runs `sentry cli setup --method brew --no-modify-path`. This sets up shell completions, agent skills, and records the install method in the DB. Using `--no-modify-path` since Homebrew manages PATH itself (the bin directory is already symlinked into the prefix). `post_install` runs after both initial install and upgrades, so it covers both cases.

PR #283, opened 2026-02-23
feat(init): add init command for guided Sentry project setup

betegon:
## Summary

Adds `sentry init` — an AI-powered wizard that walks users through adding Sentry to their project. It detects the platform, installs the SDK, instruments the code, and configures error monitoring, tracing, and session replay.

### Core wizard
- New `init` command backed by a Mastra AI workflow (hosted at [getsentry/cli-init-api](https://github.com/getsentry/cli-init-api/)) that handles platform detection, SDK installation, and code instrumentation
- ASCII banner, AI transparency note, and review reminder in the wizard UX
- Tracing: unique trace IDs per wizard run with flattened span hierarchy
- Python platforms use venv for isolated dependency installation
- Magic values extracted into named constants (`constants.ts`)
- Docs page added to cli.sentry.dev

### Security hardening
- Shell metacharacter blocklist: `()` subshell bypass, `>` `<` `&` redirection/background, `$` `'` `"` `\` expansion/escaping, `{` `}` `*` `?` glob/brace expansion, `#` shell comment
- Environment variable injection blocking (`VAR=value cmd` pattern)
- Remote-supplied `cwd` validation against project directory
- Dangerous executable blocklist and path-traversal prevention
- File descriptor cleanup on `readSync` failure
- Cross-platform shell execution (`{ shell: true }` instead of hardcoded `sh`)

### Performance
- Pre-computed directory listing sent with first API call to skip initial `list-dir` round-trip
- `_prevPhases` cross-phase caching to reuse results across workflow steps (#307)

### Testing & CI
- Wizard-runner unit tests (coverage 5.94% → 99.42%)
- Banner extracted to `src/lib/banner.ts` to break circular import
- Mock isolation fixes (`spyOn` instead of `mock.module` where possible)
- Simplified coverage pipeline (removed merge-lcov workaround)

## Eval suite

The eval suite validates that the wizard produces correct, buildable Sentry instrumentation for each supported platform. It uses a **3-phase test architecture**:

### Phase 1: Wizard run
Each test scaffolds a fresh project from a platform template, then runs the full `sentry init` wizard against it. The wizard output (exit code, stdout/stderr, git diff, new files) is captured for the next phases.

### Phase 2: Hard assertions (deterministic)
Five code-based pass/fail checks that run without any LLM:
1. **exit-code** — wizard exits 0
2. **sdk-installed** — the Sentry SDK package appears in the dependency file (`package.json` / `requirements.txt`)
3. **init-present** — `Sentry.init` (or `sentry_sdk.init`) appears in changed or new files
4. **no-placeholder-dsn** — no leftover placeholder DSNs (`___PUBLIC_DSN___`, `YOUR_DSN_HERE`, etc.)
5. **build-succeeds** — `npm run build` / equivalent passes after the wizard's changes

### Phase 3: LLM judge (per-feature)
For each feature (errors, tracing, replay, logs, profiling, etc.), an LLM judge scores correctness:
- Official Sentry docs are fetched as ground truth (URLs mapped in `feature-docs.json`)
- GPT-4o evaluates the wizard's diff + new files against the docs on 4 criteria: feature-initialized, correct-imports, no-syntax-errors, follows-docs
- Each criterion is scored pass/fail/unknown; the overall feature score must be >= 0.5

### Platforms
6 platform templates are covered:
| Platform | Template | SDK |
|----------|----------|-----|
| Express | `express/` | `@sentry/node` |
| Next.js | `nextjs/` | `@sentry/nextjs` |
| SvelteKit | `sveltekit/` | `@sentry/sveltekit` |
| React + Vite | `react-vite/` | `@sentry/react` |
| Flask | `python-flask/` | `sentry-sdk` |
| FastAPI | `python-fastapi/` | `sentry-sdk` |

### Running
```bash
bun run test:init-eval          # all platforms
```
Requires `SENTRY_AUTH_TOKEN`, `SENTRY_ORG`, `SENTRY_PROJECT`, and optionally `OPENAI_API_KEY` (LLM judge is skipped without it).

## Test Plan

- [x] All init unit tests pass (124 tests across 7 files)
- [x] `bun run lint` and `bun run typecheck` pass
- [x] CI passes (unit tests, e2e, lint, typecheck, build)
- CI workflow for eval is tracked separately in #290


src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,535 @@
...
+
+      if (
+        recursive &&
+        entry.isDirectory() &&
+        !entry.name.startsWith(".") &&
+        entry.name !== "node_modules"
+      ) {
+        walk(path.join(dir, entry.name), depth + 1);
+      }
+    }
```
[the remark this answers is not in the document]
betegon: will never follow a symlink into an outside directory because the `entry.isDirectory()` guard blocks it

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    const history = stepHistory.get(stepId) ?? [];
+    history.push(localResult);
+    stepHistory.set(stepId, history);
+
+    return {
+      ...localResult,
+      _phase: nextPhase(stepPhases, stepId, ["read-files", "analyze", "done"]),
+      _prevPhases: history.slice(0, -1),
+    };
```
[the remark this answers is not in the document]
betegon: This is by design. Key points:

  1. Per-step isolation: `_prevPhases` only accumulates within a single step. Different steps have separate histories. There's no cross-step accumulation.
  2. Bounded by phase count: typical steps have 2-3 phases (read-files → analyze → done), so _prevPhases holds at most 1-2 prior results.
  3. Size-capped content:  files are capped at 256KB (`MAX_FILE_BYTES`), commands at 64KB (`MAX_OUTPUT_BYTES`).
  4. Intentional trade-off: sending prior phase results lets the server skip re-reading files, saving API round trips. The payload growth (a few hundred KB at worst) is negligible compared to eliminating redundant file reads.

src/lib/sentry-urls.ts
```diff
@@ -15,6 +15,27 @@ export function getSentryBaseUrl(): string {
...
+ */
+export function getOrgBaseUrl(orgSlug: string): string {
+  const base = getSentryBaseUrl();
+  if (!isSentrySaasUrl(base)) {
+    return base;
+  }
+  const parsed = new URL(base);
+  parsed.hostname = `${orgSlug}.${parsed.hostname}`;
+  return parsed.origin;
+}
```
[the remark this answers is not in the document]
betegon: what? we're testing it. that's why we export it.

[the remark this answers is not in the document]
betegon: The only way to hit this would be a user manually setting `SENTRY_URL=https://us.sentry.io` in their
  environment, which the documentation explicitly says not to do (`configuration.md` says _"SaaS users should not set this"_)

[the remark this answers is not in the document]
betegon: this is intended.

test/init-eval/helpers/judge.ts
```diff
@@ -0,0 +1,143 @@
...
+      jsonMatch[0].slice(0, 200)
+    );
+    return null;
+  }
+
+  // Score: ignore "unknown" criteria, only count pass/fail
+  const gradable = parsed.criteria.filter((c) => c.pass !== "unknown");
+  const passing = gradable.filter((c) => c.pass === true).length;
+  const total = gradable.length;
+  const score = total > 0 ? passing / total : 0;
```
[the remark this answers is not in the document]
betegon: on purpose too. i'm not putting an unknown to 1.

plugins/sentry-cli/skills/sentry-cli/SKILL.md
```diff
@@ -630,6 +630,20 @@ View logs associated with a trace
...
+### Init
+
+Initialize Sentry in your project
+
+#### `sentry init <directory>`
+
+Initialize Sentry in your project
+
+**Flags:**
+- `--force - Continue even if Sentry is already installed`
```
BYK, line 642: This is a bit weird as we may want to extend existing Sentry installation (only errors to errors + tracing etc)

src/lib/sentry-urls.ts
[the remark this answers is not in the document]
betegon: `SENTRY_URL` is for selfhosting. No one should put `SENTRY_URL=https://us.sentry.io/`

src/commands/init.ts
```diff
@@ -0,0 +1,84 @@
...
+  readonly features?: string;
+};
+
+export const initCommand = buildCommand<InitFlags, [string?], SentryContext>({
+  docs: {
+    brief: "Initialize Sentry in your project",
+    fullDescription:
+      "Runs the Sentry setup wizard to detect your project's framework, " +
+      "install the SDK, and configure error monitoring. Uses a remote " +
+      "workflow that coordinates local file operations through the CLI.",
```
BYK, line 27: >Uses a remote workflow that coordinates local file operations through the CLI.

Do we really need to call this out in the help text?

src/commands/init.ts
```diff
@@ -0,0 +1,84 @@
...
+          optional: true,
+        },
+      ],
+    },
+    flags: {
+      force: {
+        kind: "boolean",
+        brief: "Continue even if Sentry is already installed",
+        default: false,
+      },
```
BYK, line 46: As mentioned earlier, I vote to remove this.

src/commands/init.ts
```diff
@@ -0,0 +1,84 @@
...
+      "dry-run": {
+        kind: "boolean",
+        brief: "Preview changes without applying them",
+        default: false,
+      },
+      features: {
+        kind: "parsed",
+        parse: String,
+        brief: "Comma-separated features: errors,tracing,logs,replay,metrics",
```
BYK, line 60: I'd also allow `+` and ` ` (for quoted args) for defensive parsing. We may also consider allowing multiple `--features` or `--feature` flags to join them. Similar to what we do with `--field` in the `api` command

src/lib/init/clack-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+export class WizardCancelledError extends Error {
+  constructor() {
+    super("Setup cancelled.");
+    this.name = "WizardCancelledError";
+  }
+}
+
+export function abortIfCancelled<T>(value: T | symbol): T {
+  if (isCancel(value)) {
+    cancel(`Setup cancelled. Visit ${SENTRY_DOCS_URL} to set up manually.`);
```
BYK, line 19:
```suggestion
    cancel(`Setup cancelled. You can visit ${SENTRY_DOCS_URL} to set up manually.`);
```

src/lib/init/clack-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    throw new WizardCancelledError();
+  }
+  return value as T;
+}
+
+const FEATURE_INFO: Record<string, { label: string; hint: string }> = {
+  errorMonitoring: {
+    label: "Error Monitoring",
+    hint: "Automatic error and crash reporting",
```
BYK, line 28: Why mention "automatic"? :)
```suggestion
    hint: "Error and crash reporting",
```

src/lib/init/clack-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  performanceMonitoring: {
+    label: "Performance Monitoring",
```
BYK, line 31:
```suggestion
    label: "Performance Monitoring (Tracing)",
```

src/lib/init/clack-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  sessionReplay: {
+    label: "Session Replay",
+    hint: "Visual replay of user sessions",
+  },
+  profiling: {
+    label: "Profiling",
+    hint: "Code-level performance insights",
+  },
+  logs: { label: "Logging", hint: "Structured log ingestion" },
```
BYK, line 42: Not sure if this ordering is reflected on the UI but if it is, I'd order it as follows:

1. Errors
2. Replay (if UI project)
3. Tracing
4. Logs
5. Metrics
6. Profiling
7. Source maps / Debug IDs (this doesn't apply to Python btw)

src/lib/init/constants.ts
```diff
@@ -0,0 +1,24 @@
...
+export const MAX_FILE_BYTES = 262_144; // 256KB per file
+export const MAX_OUTPUT_BYTES = 65_536; // 64KB stdout/stderr truncation
+export const DEFAULT_COMMAND_TIMEOUT_MS = 120_000; // 2 minutes
+export const API_TIMEOUT_MS = 120_000; // 2 minutes timeout for Mastra API calls
+
+// Exit codes returned by the remote workflow
+export const EXIT_SENTRY_ALREADY_INSTALLED = 10;
+export const EXIT_PLATFORM_NOT_DETECTED = 20;
+export const EXIT_DEPENDENCY_INSTALL_FAILED = 30;
+export const EXIT_VERIFICATION_FAILED = 50;
```
BYK, line 18: The error code idea is great, we should consider spreading this around the entire code base and clearly documenting them on the docs site.

src/lib/init/formatters.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (action === "create") {
+    return "+";
+  }
+  if (action === "delete") {
+    return "-";
+  }
+  return "~";
+}
+
+function buildSummaryLines(output: WizardOutput): string[] {
```
BYK, line 27: We should use the markdown tables (I think we have a specialized key-value table helper already) for this output

src/lib/init/formatters.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  const changedFiles = output.changedFiles;
+  if (changedFiles?.length) {
+    lines.push("");
+    lines.push("Changed files:");
+    for (const f of changedFiles) {
+      lines.push(`  ${fileActionIcon(f.action)} ${f.path}`);
+    }
+  }
```
BYK, line 58: We should use markdown lists and also colorize this output

Should we not have a JSON version of this? Or do we already and I missed it?

src/lib/init/formatters.ts
```diff
@@ -0,0 +1,113 @@
...
+    note(lines.join("\n"), "Setup complete");
+  }
+
+  if (output.warnings?.length) {
+    for (const w of output.warnings) {
+      log.warn(w);
+    }
+  }
+
+  log.info("Please review the changes above before committing.");
```
BYK, line 77: I'd bake this in too with a built-in diff viewer and auto commit if the user used `--yes`

src/lib/init/interactive.ts
```diff
@@ -0,0 +1,145 @@
...
+  switch (payload.kind) {
+    case "select":
+      return await handleSelect(payload, options);
+    case "multi-select":
+      return await handleMultiSelect(payload, options);
+    case "confirm":
+      return await handleConfirm(payload, options);
+    default:
+      return { cancelled: true };
+  }
```
BYK, line 34: Looks like this would be better as a dispatch-map?

```ts
const HANDLERS = {
  "select": handleSelect,
  "multi-select": handleMultiSelect,
  "confirm": handleConfirm,
}

export async function handleInteractive(....) {
  const handler = HANDLERS[payload.kind];
  return handler ? await handler(payload, options) : {cancelled: true};
}
```

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+    { pattern: "?", label: "glob expansion (?)" },
+    { pattern: "#", label: "shell comment (#)" },
+  ];
+
+const WHITESPACE_RE = /\s+/;
+
+/**
+ * Executables that should never appear in a package install command.
+ */
+const BLOCKED_EXECUTABLES = new Set([
```
BYK, line 82: I feel like it would be safer to go with an allow-list instead of a block-list here

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+          error: `Unknown operation: ${
+            // biome-ignore lint/suspicious/noExplicitAny: payload is of type LocalOpPayload
+            (payload as any).operation
+          }`,
+        };
+    }
+  } catch (error) {
+    return {
+      ok: false,
+      error: error instanceof Error ? error.message : String(error),
```
BYK, line 279: Same as earlier, I recommend a dispatch-map

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+    };
+  }
+}
+
+function listDir(payload: ListDirPayload): LocalOpResult {
```
BYK, line 284: Two concerns in this function:

1. Directory may have symlinks that point outside the directory. Not clear to me if we check them
2. Using `opendir` would probably be more efficient and speedy: `await opendir(INPUT_DIR, {recursive: true, bufferSize: 1024});`

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+          fs.readSync(fd, buffer, 0, maxBytes, 0);
+        } finally {
+          fs.closeSync(fd);
+        }
+        files[filePath] = buffer.toString("utf-8");
+      } else {
+        files[filePath] = fs.readFileSync(absPath, "utf-8");
+      }
+    } catch {
+      files[filePath] = null;
```
BYK, line 353: We should def use `async` here. Having `maxBytes` is great, we may need cursor/offset based reading for the agent in the future to get a specific part of the file extracted. Pretty common.

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+    const result = await runSingleCommand(command, cwd, timeoutMs);
+    results.push(result);
+    if (result.exitCode !== 0) {
+      return {
+        ok: false,
+        error: `Command "${command}" failed with exit code ${result.exitCode}: ${result.stderr}`,
+        data: { results },
+      };
+    }
+  }
```
BYK, line 419: If these are not necessarily oredered commands, would be nice to do async and in parallel. Or if this becomes a bottleneck, we may expose a generic `parallelRun` tool in the future.

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+  timeoutMs: number
+): Promise<{
+  command: string;
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}> {
+  return new Promise((resolve) => {
+    const child = spawn(command, [], {
+      shell: true,
```
BYK, line 438: This is quite dangerous, I wonder if we should avoid shell commands and just run binaries directly instead.

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+
+  return { ok: true, data: { applied } };
+}
+
+/**
+ * Resolve the org slug from local config, env vars, or by listing the user's
+ * organizations from the API as a fallback.
+ * Returns the slug on success, or a LocalOpResult error to return early.
+ */
+async function resolveOrgSlug(
```
BYK, line 573: We should already have this somewhere around the codebase including offline checking first.

src/lib/init/local-ops.ts
```diff
@@ -0,0 +1,690 @@
...
+      hint: o.slug,
+    })),
+  });
+  if (isCancel(selected)) {
+    return { ok: false, error: "Organization selection cancelled." };
+  }
+  return selected;
+}
+
+async function createSentryProject(
```
BYK, line 616: Are we duplicating the `sentry project create` logic here or am I missing something? We should abstract out the core part of that and use it directly from here.

package.json
```diff
@@ -27,6 +30,7 @@
     "http-cache-semantics": "^4.2.0",
     "ignore": "^7.0.5",
     "marked": "^15",
+    "openai": "^6.22.0",
```
BYK, line 33: Why do we need this?

package.json
```diff
@@ -6,7 +6,10 @@
     "url": "git+https://github.com/getsentry/cli.git"
   },
   "devDependencies": {
+    "@anthropic-ai/sdk": "^0.39.0",
```
BYK, line 9: You know I'm not a fan but why not Vercel AI SDK as I assume that's more established?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+  ctx: StepContext,
+  stepPhases: Map<string, number>,
+  stepHistory: Map<string, Record<string, unknown>[]>
+): Promise<Record<string, unknown>> {
+  const { payload, stepId, spin, spinState, options } = ctx;
+  const label = STEP_LABELS[stepId] ?? stepId;
+
+  if (payload.type === "local-op") {
+    const detail = payload.operation ? ` (${payload.operation})` : "";
+    spin.message(`${label}${detail}...`);
```
BYK, line 64: We should be using our markdown pipeline if we don't already

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+    // (the server skips apply-patchset). Auto-continue since this is expected.
+    if (options.dryRun && stepId === VERIFY_CHANGES_STEP) {
+      return {
+        action: "continue",
+        _phase: nextPhase(stepPhases, stepId, ["apply"]),
+      };
+    }
+
+    spin.stop(label);
+    spinState.running = false;
```
BYK, line 90: Why do we have a separate `spinState` thing? Why not read it directly from `spin`?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    const interactiveResult = await handleInteractive(payload, options);
+
+    spin.start("Processing...");
```
BYK, line 94: For future: we should have fun random things to say here like our UI loader (reticulating splines, scortching the Earth with AI tokens etc.)

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  // Unreachable: assertSuspendPayload validates the type before we get here.
+  // Kept as a defensive fallback.
+  spin.stop("Error", 1);
+  spinState.running = false;
+  log.error(
+    `Unknown suspend payload type "${(payload as { type: string }).type}"`
+  );
+  cancel("Setup failed");
+  throw new WizardCancelledError();
```
BYK, line 111: This error should have the message from line 107 above and whatever handles the error should call `log.error()` there not here.

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+}
+
+function errorMessage(err: unknown): string {
+  return err instanceof Error ? err.message : String(err);
+}
+
+function assertWorkflowResult(raw: unknown): WorkflowRunResult {
+  if (!raw || typeof raw !== "object") {
+    throw new Error("Invalid workflow response: expected object");
```
BYK, line 120: No custom errors here?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+}
+
+function assertSuspendPayload(raw: unknown): SuspendPayload {
+  if (!raw || typeof raw !== "object") {
+    throw new Error("Invalid suspend payload: expected object");
+  }
+  const obj = raw as Record<string, unknown>;
+  if (
+    typeof obj.type !== "string" ||
+    !["local-op", "interactive"].includes(obj.type)
```
BYK, line 139: This should 100% be a `const` `Set` defined outside of this function.

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+/**
+ * Race a promise against a timeout. Rejects with a descriptive error
+ * if the promise doesn't settle within `ms` milliseconds.
+ */
+function withTimeout<T>(
+  promise: Promise<T>,
+  ms: number,
+  label: string
+): Promise<T> {
+  return new Promise((resolve, reject) => {
```
BYK, line 155: The problem with this approach is it never actually stops the underlying promise. At the very least, I think we should call `promise.unref()` before rejecting on line 157. And if this is an HTTP request, we should do the proper `AbortHandler` dance to abort it, however annoying it might be.

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+        reject(err);
+      }
+    );
+  });
+}
+
+export async function runWizard(options: WizardOptions): Promise<void> {
+  const { directory, force, yes, dryRun, features } = options;
+
+  if (!(yes || process.stdin.isTTY)) {
```
BYK, line 176: The intent is clearer this way:
```suggestion
  if (!process.stdin.isTTY && !yes) {
```

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+    process.stderr.write(
```
BYK, line 177: use `log.error()`

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+      "Error: Interactive mode requires a terminal. Use --yes for non-interactive mode.\n"
+    );
+    process.exitCode = 1;
+    return;
+  }
+
+  process.stderr.write(`\n${formatBanner()}\n\n`);
```
BYK, line 184: Why to `stderr`?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+  intro("sentry init");
```
BYK, line 185: Why is the banner printing part not part of the `intro` helper already?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  log.info(
+    "This wizard uses AI to analyze your project and configure Sentry." +
+      `\nFor manual setup: ${SENTRY_DOCS_URL}`
+  );
+
+  const tracingOptions = {
+    traceId: randomBytes(16).toString("hex"),
+    tags: ["sentry-cli", "init-wizard"],
```
BYK, line 198: Okay, what? What is the `sentry-cli` tag? What are we doing here with a custom trace id etc?

src/lib/init/wizard-runner.ts
```diff
@@ -0,0 +1,353 @@
...
+      if (!extracted) {
+        spin.stop("Error", 1);
+        spinState.running = false;
+        log.error(`No suspend payload found for step "${stepId}"`);
+        cancel("Setup failed");
+        process.exitCode = 1;
+        return;
+      }
+
+      const resumeData = await handleSuspendedStep(
```
BYK, line 264: I'd say if we can, we should support resuming from suspension from information stored on disk in case of an unexpected process termination

src/lib/sentry-urls.ts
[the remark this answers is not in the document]
BYK: @betegon yes BUT bots may do it accidentally. My approach to these situations is to recover if it is safe and then emit a warning to educate still

src/lib/init/interactive.ts
betegon: the issue is TS's "correlated unions" problem. When you index into a dispatch map with a union key, TS can't correlate the handler type with the payload type. You'd need a `PayloadKindMap` + mapped `HandlerMap` type and still end up with an as never cast at the call site to make it compile. For 3 cases the switch feels simpler and gives us type narrowing + exhaustiveness for free. Happy to refactor if this grows though

PR #288, opened 2026-02-24
fix(brew): handle root-owned config dir from sudo installs

BYK:
## Problem

Two errors were reported during Homebrew post-install of v0.12.0:

1. `SQLiteError: unable to open database file` (SQLITE_CANTOPEN) — setup aborts
2. `Warning: read-only database` + `EPERM` on zsh completions — setup aborts

Root cause: `sudo brew install` creates root-owned `~/.sentry/` and `~/.local/share/zsh/` files. The setup command had no error handling, so any failure aborted the entire post-install script and Homebrew showed a scary error even though the binary installed fine.

The same root cause explains the recurring "attempt to write a readonly database" Sentry telemetry issues (8 issues, 19 events, 100% macOS). Moving the config directory would not help — the same permission problems would occur at any path if created by root. macOS TCC does not restrict `~/.sentry/`.

### Non-fatal setup steps (`setup.ts`)
Added a `bestEffort()` wrapper around each post-install configuration step. Permission failures now log a warning instead of aborting. The binary is already installed — these steps are nice-to-have side effects.

### Root-owned file detection in `tryRepairReadonly` (`telemetry.ts`)
Before attempting `chmod` (which fails silently on root-owned files), now checks `stat().uid`. If the file is owned by root, emits a targeted actionable message with the actual username:
```
Warning: Sentry CLI config directory is owned by root.
  Path:  /Users/am/.sentry
  Fix:   sudo chown -R am "/Users/am/.sentry"
  Or:    sudo sentry cli fix
```
Username is inferred from `SUDO_USER` → `USER` → `USERNAME` → `os.userInfo()`.

### Ownership repair in `sentry cli fix` (`fix.ts`)
- Ownership check runs **before** permissions (chmod fails on root-owned files anyway)
- **Not root**: prints `sudo chown -R <user> <configDir>` and `sudo sentry cli fix` instructions, exits with code 1
- **Running as root** (`sudo sentry cli fix`): resolves the real user's UID via `id -u <username>`, then performs `chown` to transfer ownership back

Fixes CLI-7Q, CLI-7K, CLI-6N, CLI-6D, CLI-4Z, CLI-51, CLI-4E, CLI-2E

### DIFF
--- a/src/commands/cli/fix.ts
+++ b/src/commands/cli/fix.ts
@@ -1,10 +1,11 @@
 /**
  * sentry cli fix
  *
- * Diagnose and repair CLI database issues (schema and permissions).
+ * Diagnose and repair CLI database issues (schema, permissions, and ownership).
  */
 
-import { chmod, stat } from "node:fs/promises";
+import { execFileSync } from "node:child_process";
+import { chmod, chown, stat } from "node:fs/promises";
 import type { SentryContext } from "../../context.js";
 import { buildCommand } from "../../lib/command.js";
 import { getConfigDir, getDbPath, getRawDatabase } from "../../lib/db/index.js";
@@ -14,6 +15,7 @@ import {
   repairSchema,
   type SchemaIssue,
 } from "../../lib/db/schema.js";
+import { getRealUsername } from "../../lib/utils.js";
 
 type FixFlags = {
   readonly "dry-run": boolean;
@@ -41,6 +43,17 @@ type PermissionIssue = {
   kind: "directory" | "database" | "journal";
   currentMode: number;
   expectedMode: number;
+};
+
+/**
+ * A file or directory that is owned by a different user (typically root),
+ * preventing the current process from writing to it.
+ */
+type OwnershipIssue = {
+  path: string;
+  kind: "directory" | "database" | "journal";
+  /** UID of the file's current owner */
+  ownerUid: number;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


src/commands/cli/fix.ts
```diff
@@ -132,6 +160,208 @@ async function checkPermissions(dbPath: string): Promise<PermissionIssue[]> {
...
+            : undefined;
+        // Missing files are fine (WAL/SHM created on demand).
+        if (code === "ENOENT" || code === "EACCES") {
+          return null;
+        }
+        throw error;
+      }
+      return null;
+    })
+  );
```
BYK: Can this be simplified via `Promise.allSettled`?

src/commands/cli/fix.ts
```diff
@@ -132,6 +160,208 @@ async function checkPermissions(dbPath: string): Promise<PermissionIssue[]> {
...
+  return results.filter((r): r is OwnershipIssue => r !== null);
+}
+
+/**
+ * Resolve the numeric UID for a username by running `id -u <username>`.
+ * Returns null if the lookup fails or returns a non-numeric result.
+ */
+async function resolveUid(username: string): Promise<number | null> {
+  try {
+    const result = await Bun.$`id -u ${username}`.text();
```
BYK: Will this work on our `node` builds? Do we have a proper shim?

PR #289, opened 2026-02-24
feat(issue-list): add --period flag, pagination progress, and count abbreviation

BYK:
## Summary

Three UX improvements to `sentry issue list` following the auto-pagination work in #274:

- **Fix implicit time filtering** — add `--period` / `-t` flag (default `90d`) so the CLI matches the UI's default time window instead of the Sentry API's implicit 14d cutoff, which was causing the count discrepancy (e.g. 114 vs 240 issues)
- **Pagination progress spinner** — animated braille spinner on stderr while fetching multiple pages (`Fetching issues... 200/500`); reuses `withProgress<T>()` in `polling.ts` so the animation is consistent with `sentry issue explain`
- **Fix COUNT column layout shift** — abbreviate event counts ≥10k with K/M/B/T/P/E suffixes (`12.3K`, `150K`, `1.5M`) so the column stays exactly 5 chars wide and never overflows

## Details

- `--period` alias is `-t` (time period), not `-p` (avoids confusion with `--platform` in other commands)
- `onPage` callback added to `listIssuesAllPages` for per-page progress hooks
- Animation interval changed from 80ms to 50ms (20fps / 500ms full braille cycle), matching the ora/inquirer standard — applies to Seer commands too
- Decimal shown only when `scaled < 100` (e.g. `12.3K` and `1.5M` but not `100.0M`)

### DIFF
--- a/src/lib/formatters/human.ts
+++ b/src/lib/formatters/human.ts
@@ -335,13 +335,10 @@ function abbreviateCount(raw: string): string {
   if (Number.isNaN(n)) {
     // Non-numeric input: use a placeholder rather than passing through an
     // arbitrarily wide string that would break column alignment
-    Sentry.captureMessage(
-      `Unexpected non-numeric issue count: ${raw}`,
-      "warning"
-    );
+    Sentry.logger.warn(`Unexpected non-numeric issue count: ${raw}`);
     return "?".padStart(COL_COUNT);
   }
-  if (n < 10_000) {
+  if (raw.length <= COL_COUNT) {
     return raw.padStart(COL_COUNT);
   }
   const tier = Math.min(Math.floor(Math.log10(n) / 3), QUANTIFIERS.length - 1);
--- a/src/lib/telemetry.ts
+++ b/src/lib/telemetry.ts
@@ -263,6 +263,8 @@ export function initSentry(enabled: boolean): Sentry.BunClient | undefined {
         (integration) => !EXCLUDED_INTEGRATIONS.has(integration.name)
       ),
     environment,
+    // Enable Sentry structured logs for non-exception telemetry (e.g., unexpected input warnings)
+    enableLogs: true,
     // Sample all events for CLI telemetry (low volume)
     tracesSampleRate: 1,
     sampleRate: 1,


src/commands/issue/list.ts
```diff
@@ -483,7 +499,22 @@ async function handleOrgAllIssues(options: OrgAllIssuesOptions): Promise<void> {
 
   setContext([org], []);
 
-  const { issues, nextCursor } = await fetchOrgAllIssues(org, flags, cursor);
+  const shouldShowProgress = flags.limit > API_MAX_PER_PAGE && !flags.json;
+  const { issues, nextCursor } = await withProgress(
+    {
+      stderr,
+      message: "Fetching issues...",
+      json: !shouldShowProgress,
```
BYK: This usage feels a bit backwards (why decide JSON based on `shouldShowProgress`? Seems incidental and unsafe). I expect `withProgress` to determine this. Also, I wonder if we can still emit the progress on `stderr` for JSON output?

src/commands/issue/list.ts
```diff
@@ -574,14 +605,31 @@ async function handleResolvedTargets(
...
+          fetchIssuesForTarget(t, {
+            query: flags.query,
+            limit: flags.limit,
+            sort: flags.sort,
+            statsPeriod: flags.period,
+            onPage,
+          })
+        )
+      );
+    }
```
BYK, line 624: Let's be more concise?

```suggestion
    (setMessage) => Promise.all(
        targets.map((t) =>
          fetchIssuesForTarget(t, {
            query: flags.query,
            limit: flags.limit,
            sort: flags.sort,
            statsPeriod: flags.period,
            onPage: shouldShowProgress
              ? (fetched: number, limit: number) => setMessage(`Fetching issues... ${fetched}/${limit}`) : undefined,
          })
        )
      )
```

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,28 @@ const COL_SEEN = 10;
...
+    return raw.padStart(COL_COUNT);
+  }
+  let abbreviated: string;
+  if (n >= 1_000_000_000) {
+    abbreviated = `${(n / 1_000_000_000).toFixed(n >= 10_000_000_000 ? 0 : 1)}B`;
+  } else if (n >= 1_000_000) {
+    abbreviated = `${(n / 1_000_000).toFixed(n >= 10_000_000 ? 0 : 1)}M`;
+  } else {
+    abbreviated = `${(n / 1000).toFixed(n >= 100_000 ? 0 : 1)}K`;
+  }
```
BYK, line 358: Is it costlier to do a `Math.floor(Math.log10(n) / 3)` and map that to something like `const QUANTIFIERS = ['', 'K', 'M', 'B'];` ?

src/lib/formatters/human.ts
```diff
@@ -582,7 +604,7 @@ export function formatIssueRow(
   const rawLen = getShortIdDisplayLength(issue.shortId);
   const shortIdPadding = " ".repeat(Math.max(0, COL_SHORT_ID - rawLen));
   const shortId = `${formattedShortId}${shortIdPadding}`;
-  const count = `${issue.count}`.padStart(COL_COUNT);
+  const count = abbreviateCount(`${issue.count}`);
```
BYK, line 633: We still don't seem to ensure we never exceed the column width here

src/lib/polling.ts
```diff
@@ -121,3 +121,71 @@ export async function poll<T>(options: PollOptions<T>): Promise<T> {
...
+  options: WithProgressOptions,
+  fn: (setMessage: (msg: string) => void) => Promise<T>
+): Promise<T> {
+  const { stderr, json = false } = options;
+  let currentMessage = options.message;
+  let tick = 0;
+
+  let animationTimer: Timer | undefined;
+  if (!json) {
+    animationTimer = setInterval(() => {
```
BYK: Not sure if `setInterval()` is safe as there's this general concern about it trying to execute regardless of how long the callback function takes to run. So scheduling a next tic with `setTimeout` might be safer? We should also 100% `.unref()` this.

src/commands/issue/list.ts
BYK: Fixed — removed the `json` option from `withProgress` entirely. Progress always goes to stderr which doesn't contaminate stdout JSON output, so the spinner now runs unconditionally. Callers just pass `{ stderr, message }` and always get the animated spinner.

src/commands/issue/list.ts
BYK: Done — inlined `onPage` directly into `fetchIssuesForTarget` and removed the `shouldShowPageCount` conditional. Both call sites are now ~7-15 lines instead of 15-25.

src/lib/formatters/human.ts
BYK: Adopted — rewrote with `Math.floor(Math.log10(n) / 3)` indexing into a `QUANTIFIERS = ["", "K", "M", "B", "T"]` array. Also added T (trillions) for free.

src/lib/formatters/human.ts
BYK: Fixed — the result is now guaranteed to fit: tries `toFixed(1)` first, falls back to `Math.round()` if too wide, then a final `.slice(0, COL_COUNT).padStart(COL_COUNT)` ensures the output is always exactly COL_COUNT characters.

src/lib/polling.ts
BYK: Fixed — replaced `setInterval` with a `setTimeout` chain (each frame schedules the next via `setTimeout(scheduleFrame, ANIMATION_INTERVAL_MS).unref()`). A `stopped` flag halts the chain in `finally`. Applied to both `withProgress` and `poll`. The `.unref()` ensures the timer doesn't keep the process alive.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,35 @@ const COL_SEEN = 10;
...
+  const tier = Math.min(Math.floor(Math.log10(n) / 3), QUANTIFIERS.length - 1);
+  const suffix = QUANTIFIERS[tier] ?? "";
+  const scaled = n / 10 ** (tier * 3);
+  // Use 1 decimal when it fits within COL_COUNT, otherwise 0
+  const withDecimal = `${scaled.toFixed(1)}${suffix}`;
+  const abbreviated =
+    withDecimal.length <= COL_COUNT
+      ? withDecimal
+      : `${Math.round(scaled)}${suffix}`;
+  return abbreviated.slice(0, COL_COUNT).padStart(COL_COUNT);
```
BYK: Won't this arbitrarily cut the count making it confusing or even outright incorrect without any warning?

src/lib/polling.ts
```diff
@@ -115,9 +120,77 @@ export async function poll<T>(options: PollOptions<T>): Promise<T> {
...
+  const scheduleFrame = () => {
+    if (stopped) {
+      return;
+    }
+    const display = truncateProgressMessage(currentMessage);
+    stderr.write(`\r\x1b[K${formatProgressLine(display, tick)}`);
+    tick += 1;
+    setTimeout(scheduleFrame, ANIMATION_INTERVAL_MS).unref();
+  };
+  setTimeout(scheduleFrame, ANIMATION_INTERVAL_MS).unref();
```
BYK: Why not just call `scheduleFrame()` to start the animation immediately? The code would also look simpler.

src/lib/formatters/human.ts
BYK: You're right — removed the `.slice()`. The `toFixed(1)` → `Math.round()` fallback already guarantees the result fits within COL_COUNT (worst case is `"1000K"` = 5 chars at the K→M boundary), so the slice was dead code that would only silently corrupt the value if my analysis were wrong.

src/lib/polling.ts
BYK: Good call — changed both `poll` and `withProgress` to call `scheduleFrame()` directly. First frame now renders immediately instead of after a 50ms delay.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,34 @@ const COL_SEEN = 10;
...
+  const tier = Math.min(Math.floor(Math.log10(n) / 3), QUANTIFIERS.length - 1);
+  const suffix = QUANTIFIERS[tier] ?? "";
+  const scaled = n / 10 ** (tier * 3);
+  // Only show decimal when it adds information (scaled < 100), e.g. "12.3K" not "100.0K"
+  const abbreviated =
+    scaled < 100
+      ? `${scaled.toFixed(1)}${suffix}`
+      : `${Math.round(scaled)}${suffix}`;
+  return abbreviated.padStart(COL_COUNT);
+}
```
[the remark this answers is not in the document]
BYK: Fixed — the bug was that we checked `scaled < 100` but `toFixed(1)` can round up past the boundary (e.g. `scaled = 99.95` → `"100.0K"`). Now we pre-compute `rounded1dp = Number(scaled.toFixed(1))` and compare *that* against `< 100`, so the rounding and the guard are consistent.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,34 @@ const COL_SEEN = 10;
...
+    return raw.padStart(COL_COUNT);
+  }
+  const tier = Math.min(Math.floor(Math.log10(n) / 3), QUANTIFIERS.length - 1);
+  const suffix = QUANTIFIERS[tier] ?? "";
+  const scaled = n / 10 ** (tier * 3);
+  // Only show decimal when it adds information (scaled < 100), e.g. "12.3K" not "100.0K"
+  const abbreviated =
+    scaled < 100
+      ? `${scaled.toFixed(1)}${suffix}`
+      : `${Math.round(scaled)}${suffix}`;
```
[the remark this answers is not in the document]
BYK: Fixed — same root cause as above. `rounded1dp = Number(scaled.toFixed(1))` is now used for the `< 100` check, so `scaled = 99.95` correctly falls through to `Math.round()` → `"100K"` (4 chars).

src/commands/issue/list.ts
```diff
@@ -473,7 +489,7 @@ type OrgAllIssuesOptions = {
  * never accidentally reused.
  */
 async function handleOrgAllIssues(options: OrgAllIssuesOptions): Promise<void> {
-  const { stdout, org, flags, setContext } = options;
+  const { stdout, stderr, org, flags, setContext } = options;
```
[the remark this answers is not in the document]
BYK: Fixed — added `|period:${flags.period}` to the `contextKey`. A cursor stored from a `--period 14d` query will no longer be reused for a `--period 90d` query on the same org/sort/filter.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,36 @@ const COL_SEEN = 10;
...
+  const scaled = n / 10 ** (tier * 3);
+  // Only show decimal when it adds information — check the rounded value to avoid
+  // "100.0K" when scaled is e.g. 99.95 (toFixed(1) rounds up to "100.0")
+  const rounded1dp = Number(scaled.toFixed(1));
+  const abbreviated =
+    rounded1dp < 100
+      ? `${rounded1dp.toFixed(1)}${suffix}`
+      : `${Math.round(scaled)}${suffix}`;
+  return abbreviated.padStart(COL_COUNT);
+}
```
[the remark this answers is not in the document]
BYK: Fixed — when `Math.round(scaled) >= 1000` and a higher tier exists, we now promote: divide by 1000 and use the next suffix. So `999_950` → scaled=`999.95` → rounded=`1000` → `"1.0M"` instead of `"1000K"`.

src/lib/api-client.ts
```diff
@@ -1095,6 +1097,7 @@ export async function listIssuesAllPages(
     });
 
     allResults.push(...response.data);
+    options.onPage?.(allResults.length, options.limit);
```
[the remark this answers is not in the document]
BYK: Fixed — capped the reported count at `options.limit` via `Math.min(allResults.length, options.limit)`, so the progress line shows `100/100` rather than `120/100` when the last page overshoots.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,41 @@ const COL_SEEN = 10;
...
+    return `${rounded1dp.toFixed(1)}${suffix}`.padStart(COL_COUNT);
+  }
+  const rounded = Math.round(scaled);
+  // Promote to next tier if rounding produces >= 1000 (e.g. 999.95 → "1000K" → "1.0M")
+  if (rounded >= 1000 && tier < QUANTIFIERS.length - 1) {
+    const nextSuffix = QUANTIFIERS[tier + 1] ?? "";
+    return `${(rounded / 1000).toFixed(1)}${nextSuffix}`.padStart(COL_COUNT);
+  }
+  return `${rounded}${suffix}`.padStart(COL_COUNT);
+}
```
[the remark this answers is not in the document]
BYK: Addressed both points:

1. **Above E tier**: added `Math.min(rounded, 999)` as a final cap so values at the max tier that round to >= 1000 (unreachable in practice for Sentry counts) produce `"999E"` (5 chars) rather than overflowing.
2. **`Number.MAX_SAFE_INTEGER` precision loss**: added a JSDoc note documenting this (~9P / 9×10^15), which is well beyond any realistic Sentry event count.

src/lib/polling.ts
```diff
@@ -115,9 +120,77 @@ export async function poll<T>(options: PollOptions<T>): Promise<T> {
...
+
+  try {
+    return await fn((msg) => {
+      currentMessage = msg;
+    });
+  } finally {
+    stopped = true;
+    stderr.write("\r\x1b[K");
+  }
+}
```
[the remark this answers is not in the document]
BYK: Not a bug — this is intentional. The spinner writes to **stderr**, not stdout, so it never contaminates JSON output on stdout. We explicitly removed the `json` guard in a previous review round for this reason: `stderr` is the correct channel for progress/diagnostic output regardless of the output format.

src/commands/issue/list.ts
```diff
@@ -574,14 +596,21 @@ async function handleResolvedTargets(
...
+          fetchIssuesForTarget(t, {
+            query: flags.query,
+            limit: flags.limit,
+            sort: flags.sort,
+            statsPeriod: flags.period,
+            onPage: (fetched, limit) =>
+              setMessage(`Fetching issues... ${fetched}/${limit}`),
+          })
+        )
+      )
```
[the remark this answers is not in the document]
BYK: Not a bug — intentional design. The spinner writes exclusively to **stderr**; JSON output goes to **stdout**. These are independent streams, so the spinner never contaminates the JSON. This was specifically discussed and confirmed in an earlier review round (see thread on the `json` option removal). Consumers that merge stdout+stderr are doing so incorrectly for a CLI that emits structured JSON on stdout.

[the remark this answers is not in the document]
BYK: Same as the other spinner thread — not a bug. Progress goes to **stderr**, JSON goes to **stdout**. They are separate streams. The `json` guard was deliberately removed in a previous review round precisely because stderr output is always safe alongside JSON stdout.

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,46 @@ const COL_SEEN = 10;
...
+  const rounded = Math.round(scaled);
+  // Promote to next tier if rounding produces >= 1000 (e.g. 999.95K → "1.0M")
+  if (rounded >= 1000 && tier < QUANTIFIERS.length - 1) {
+    const nextSuffix = QUANTIFIERS[tier + 1] ?? "";
+    return `${(rounded / 1000).toFixed(1)}${nextSuffix}`.padStart(COL_COUNT);
+  }
+  // At max tier with no promotion available: cap at 999 to guarantee COL_COUNT width
+  // (numbers > 10^21 are unreachable in practice for Sentry event counts)
+  return `${Math.min(rounded, 999)}${suffix}`.padStart(COL_COUNT);
+}
```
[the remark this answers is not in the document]
BYK: Not a real concern in practice: the `999E` cap is only reached for values ≥ 10^21 (one sextillion events). `Number.MAX_SAFE_INTEGER` is ~9×10^15, so floating-point precision breaks long before we reach that tier boundary anyway — both are documented in the JSDoc. No Sentry project will have event counts anywhere near this range.

src/commands/issue/list.ts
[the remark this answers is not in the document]
BYK: Fixed — replaced the shared `setMessage` call with a `fetchedCounts` array indexed by target. Each project writes its own fetched count to `fetchedCounts[i]`, and the callback sums the array before updating the message. This gives a monotonically-increasing combined total (e.g. `150/200` across two projects) instead of racing between individual project values.

src/commands/issue/list.ts
```diff
@@ -473,17 +489,23 @@ type OrgAllIssuesOptions = {
...
  */
 async function handleOrgAllIssues(options: OrgAllIssuesOptions): Promise<void> {
-  const { stdout, org, flags, setContext } = options;
+  const { stdout, stderr, org, flags, setContext } = options;
   // Encode sort + query in context key so cursors from different searches don't collide.
   const escapedQuery = flags.query
     ? escapeContextKeyValue(flags.query)
     : undefined;
-  const contextKey = `host:${getApiBaseUrl()}|type:org:${org}|sort:${flags.sort}${escapedQuery ? `|q:${escapedQuery}` : ""}`;
+  const contextKey = `host:${getApiBaseUrl()}|type:org:${org}|sort:${flags.sort}|period:${flags.period}${escapedQuery ? `|q:${escapedQuery}` : ""}`;
```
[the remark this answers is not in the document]
BYK: Fixed — `flags.period` is now wrapped with `escapeContextKeyValue(flags.period ?? "90d")`, matching the same treatment applied to `flags.query`. A pipe in a user-supplied period value (e.g. `--period "14d|injected"`) no longer corrupts the key structure.

src/commands/issue/list.ts
```diff
@@ -574,14 +597,27 @@ async function handleResolvedTargets(
...
-      fetchIssuesForTarget(t, {
-        query: flags.query,
-        limit: flags.limit,
-        sort: flags.sort,
-      })
-    )
+  const results = await withProgress(
+    { stderr, message: "Fetching issues..." },
+    (setMessage) => {
+      const fetchedCounts = new Array<number>(targets.length).fill(0);
```
BYK: Why are we keeping this as an array instead of a running tally? (either convert or add inline comments explaining why)

src/lib/formatters/human.ts
```diff
@@ -313,6 +313,51 @@ const COL_SEEN = 10;
...
+ * (~9P / 9×10^15), which is far beyond any realistic Sentry event count.
+ *
+ * Examples: 999 → "  999", 12345 → "12.3K", 150000 → " 150K", 1500000 → "1.5M"
+ */
+function abbreviateCount(raw: string): string {
+  const n = Number(raw);
+  if (Number.isNaN(n)) {
+    // Non-numeric input: use a placeholder rather than passing through an
+    // arbitrarily wide string that would break column alignment
+    return "    ?";
```
BYK: Use `COL_COUNT` instead of manual spacing as it can change.

We should also report this to Sentry (unexpected case)

src/lib/polling.ts
```diff
@@ -115,9 +120,77 @@ export async function poll<T>(options: PollOptions<T>): Promise<T> {
     throw new Error(timeoutMessage);
   } finally {
     // Clean up animation timer
-    if (animationTimer) {
-      clearInterval(animationTimer);
+    if (!json) {
+      stopped = true;
```
BYK: Shouldn't we set `stopped = true` no matter what (regardless of `json`'s value?)

src/lib/polling.ts
BYK: Looks like most of this code is duplicated from above??

src/commands/issue/list.ts
BYK: Converted to a running total with delta tracking. The array is still needed because `onPage` reports the *cumulative* count per target, not a delta — so we need the previous value to compute `fetched - prevFetched[i]`. But now we maintain a `totalFetched` running sum and add deltas instead of re-summing via `reduce()` on every callback.

src/lib/formatters/human.ts
BYK: Fixed both — placeholder is now `"?".padStart(COL_COUNT)` so it adapts if COL_COUNT changes, and we report the unexpected input to Sentry via `captureMessage(..., "warning")`.

src/lib/polling.ts
BYK: Good catch — `stopped = true` is now set unconditionally. The `stderr.write("\n")` stays guarded by `!json` since there's nothing to clear in JSON mode, but the flag itself is always set for safety.

src/lib/polling.ts
BYK: Extracted a shared `startSpinner()` helper that returns `{ setMessage, stop }`. Both `poll()` and `withProgress()` now use it, eliminating ~30 lines of duplication. The spinner setup (scheduleFrame, stopped flag, tick counter, setTimeout chain) lives in one place.

src/lib/formatters/human.ts
```diff
@@ -333,7 +335,11 @@ function abbreviateCount(raw: string): string {
   if (Number.isNaN(n)) {
     // Non-numeric input: use a placeholder rather than passing through an
     // arbitrarily wide string that would break column alignment
-    return "    ?";
+    Sentry.captureMessage(
```
BYK: This should be an error? Otherwise we should use the new sentry logs feature instead of `captureMessage`

src/lib/formatters/human.ts
```diff
@@ -333,7 +335,11 @@ function abbreviateCount(raw: string): string {
...
+      `Unexpected non-numeric issue count: ${raw}`,
+      "warning"
+    );
+    return "?".padStart(COL_COUNT);
   }
   if (n < 10_000) {
```
BYK: Instead of hard-coding `10_000` we should go with the digit count and `COL_COUNT` comparison here

src/lib/formatters/human.ts
BYK: Switched to `Sentry.logger.warn()` and added `enableLogs: true` to `Sentry.init()` in telemetry.ts. This fits better — it's an unexpected-input warning, not an exception.

src/lib/formatters/human.ts
BYK: Replaced with `raw.length <= COL_COUNT` — since `raw` is `String(count)`, its length is exactly the digit count. This ties the threshold directly to column width and is simpler than `Math.floor(Math.log10(n)) + 1`. Numbers up to 99999 (5 digits = COL_COUNT) now display as-is instead of abbreviating at 10000.

src/lib/formatters/human.ts
```diff
@@ -313,6 +315,52 @@ const COL_SEEN = 10;
...
+  const n = Number(raw);
+  if (Number.isNaN(n)) {
+    // Non-numeric input: use a placeholder rather than passing through an
+    // arbitrarily wide string that would break column alignment
+    Sentry.logger.warn(`Unexpected non-numeric issue count: ${raw}`);
+    return "?".padStart(COL_COUNT);
+  }
+  if (raw.length <= COL_COUNT) {
+    return raw.padStart(COL_COUNT);
+  }
```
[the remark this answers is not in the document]
BYK: This is intentional, not a bug. The check `raw.length <= COL_COUNT` means: if the raw number string fits within the column width (COL_COUNT=5), display it as-is without abbreviation. "10000" is 5 characters, which fits in 5 columns — no abbreviation needed, no formatting inconsistency.

This approach was explicitly requested by the reviewer in place of a hardcoded `10_000` threshold, because it adapts automatically if `COL_COUNT` ever changes. The "intent" is not to abbreviate counts >= 10k — it is to abbreviate counts that would overflow the column.

PR #292, opened 2026-02-24
feat(upgrade): add nightly release channel

BYK:
## Summary

- Add a **nightly release channel** that tracks rolling builds from \`main\`, versioned as \`0.x.y-dev.<unix_seconds>\`
- Users opt in via \`sentry cli upgrade nightly\` or \`curl ... | bash -s -- --version nightly\`; the channel is persisted so future \`sentry cli upgrade\` calls track the same channel automatically
- Add \`--force\` flag to re-download even when already up to date

## Changes

**CI** — three new/modified jobs in \`ci.yml\`:
- \`nightly-version\`: computes the nightly version string on \`main\` pushes (skipped on PRs — skipped ≠ failed)
- \`build-binary\`: injects the nightly version into \`package.json\` before building when on \`main\`
- \`publish-nightly\`: after build+E2E pass on \`main\`, uploads \`.gz\` binaries + \`version.json\` to a rolling GitHub prerelease tagged \`nightly\` via \`gh release upload --clobber\`

**Install script** — \`--version nightly\` downloads from the \`nightly\` tag, fetches \`version.json\` for the display version, passes \`--channel nightly\` to \`cli setup\` to persist the preference

**DB** — new \`src/lib/db/release-channel.ts\`: \`getReleaseChannel()\` / \`setReleaseChannel()\` / \`parseReleaseChannel()\` using the existing \`metadata\` table (no migration needed)

**\`cli setup\`** — new \`--channel\` flag persists the channel on install (used by the install script and upgrade auto-migration)

**\`cli upgrade\`** — \`nightly\`/\`stable\` as positional version args switch channels (no separate \`--channel\` flag needed); new \`--force\` flag; auto-migrates brew/npm/pnpm/bun/yarn installs to a standalone binary when switching to nightly (with a warning about the old install)

**Version check** — reads persisted channel; fetches \`version.json\` from the nightly release for nightly users; shows "New nightly available:" in the update notification

## Testing

- 11 new unit tests in \`test/lib/db/release-channel.test.ts\`
- \`bun run typecheck\` passes clean
- \`bun run lint\` passes clean
- \`bun test test/lib/db/ test/lib/version-check.test.ts\` — 156 tests passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,11 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  # Commit timestamp used for deterministic nightly version strings.
+  # Defined at workflow level so build-binary and publish-nightly always agree.
+  COMMIT_TIMESTAMP: ${{ github.event.head_commit.timestamp }}
+
 jobs:
   changes:
     name: Detect Changes
@@ -186,16 +191,12 @@ jobs:
         # string for a given commit.
         if: github.ref == 'refs/heads/main' && github.event_name == 'push'
         shell: bash
-        env:
-          COMMIT_TIMESTAMP: ${{ github.event.head_commit.timestamp }}
         run: |
-          node -e "
-            const fs = require('fs');
-            const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
-            const ts = Math.floor(new Date(process.env.COMMIT_TIMESTAMP).getTime() / 1000);
-            pkg.version = pkg.version.replace(/-dev\.\d+$/, '-dev.' + ts);
-            fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n');
-          "
+          TS=$(date -d "$COMMIT_TIMESTAMP" +%s 2>/dev/null || date -jf "%Y-%m-%dT%H:%M:%SZ" "$COMMIT_TIMESTAMP" +%s)
+          CURRENT=$(jq -r .version package.json)
+          NIGHTLY=$(echo "$CURRENT" | sed "s/-dev\.[0-9]*$/-dev.${TS}/")
+          jq --arg v "$NIGHTLY" '.version = $v' package.json > package.json.tmp
+          mv package.json.tmp package.json
       - name: Build
         env:
           SENTRY_CLIENT_ID: ${{ vars.SENTRY_CLIENT_ID }}
@@ -317,14 +318,10 @@ jobs:
         # Uses the commit timestamp — the same value as build-binary — so
         # version.json exactly matches the version baked into the binaries.
         id: version
-        env:
-          COMMIT_TIMESTAMP: ${{ github.event.head_commit.timestamp }}
         run: |
-          VERSION=$(node -e "
-            const pkg = JSON.parse(require('fs').readFileSync('package.json', 'utf8'));
-            const ts = Math.floor(new Date(process.env.COMMIT_TIMESTAMP).getTime() / 1000);
-            console.log(pkg.version.replace(/-dev\.\d+$/, '-dev.' + ts));
-          ")
+          TS=$(date -d "$COMMIT_TIMESTAMP" +%s)
+          CURRENT=$(jq -r .version package.json)
+          VERSION=$(echo "$CURRENT" | sed "s/-dev\.[0-9]*$/-dev.${TS}/")
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
       - name: Download all binary artifacts
         uses: actions/download-artifact@v4
@@ -357,8 +354,13 @@ jobs:
           **Version:** \`${{ steps.version.outputs.version }}\`
           **Commit:** ${{ github.sha }}"
 
-          # Upload/replace all .gz binaries and the version manifest
-          gh release upload nightly --clobber \
+          # Delete all existing assets first so removed/renamed files don't linger
+          gh release view nightly --json assets --jq '.assets[].name' | while read -r name; do
+            gh release delete-asset nightly "$name" --yes
+          done
+
+          # Upload the new .gz binaries and the version manifest
+          gh release upload nightly \
             artifacts/dist-bin/*.gz \
             version.json
 


src/commands/cli/upgrade.ts
```diff
@@ -124,24 +207,101 @@ async function runSetupOnNewBinary(
...
       "  sentry cli upgrade --method npm # Force using npm to upgrade",
   },
   parameters: {
     positional: {
       kind: "tuple",
       parameters: [
         {
-          brief: "Target version to install (defaults to latest)",
+          brief:
+            'Target version, or "nightly"/"stable" to switch channel (defaults to latest)',
```
BYK: The "defaults to latest" bit is now a bit confusing

install
```diff
@@ -80,21 +81,44 @@ if [[ "$os" == "windows" ]]; then
...
   version=$(curl -fsSL https://api.github.com/repos/getsentry/cli/releases/latest | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
   if [[ -z "$version" ]]; then
     echo -e "${RED}Failed to fetch latest version${NC}"
     exit 1
   fi
+  download_tag="$version"
+elif [[ "$requested_version" == "nightly" ]]; then
+  channel="nightly"
+  download_tag="nightly"
+  version_json=$(curl -fsSL "https://github.com/getsentry/cli/releases/download/nightly/version.json")
```
BYK: We don't need the nightly version for anything, right? This only looks like a sanity check but it's an extra request per install so I'd remove it if not needed.

src/commands/cli/upgrade.ts
BYK: Fixed in 709cd50: the positional brief was already updated to: 'Specific version (e.g. 0.5.0), or "nightly"/"stable" to switch channel; omit to update within current channel'.

install
BYK: Done in 80e198f: `version="nightly"` is set directly without fetching version.json. The install script no longer does any network request for nightly other than the binary download itself.

src/commands/cli/upgrade.ts
```diff
@@ -4,68 +4,117 @@
...
     stdout.write("\nAlready up to date.\n");
     return null;
   }
 
-  if (version) {
+  // Validate that a specific pinned version actually exists on GitHub/npm
+  if (versionArg && !CHANNEL_VERSIONS.has(versionArg)) {
     const exists = await versionExists(method, target);
     if (!exists) {
       throw new UpgradeError(
```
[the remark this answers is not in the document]
BYK: Fixed in 709cd50: when `channel === 'nightly'`, `resolveTargetVersion` now passes `'curl'` (GitHub lookup) to `versionExists` regardless of the installed method. This ensures nightly-only versions are always validated against GitHub releases, not the npm registry.

src/commands/cli/upgrade.ts
```diff
@@ -4,68 +4,117 @@
...
+ * parameter. "nightly" and "stable" are treated as channel selectors, not
+ * literal version strings.
+ *
+ * @returns `{ channel, versionArg }` where versionArg is undefined when the
+ *   positional was a channel name (so we resolve to latest) or was omitted.
+ */
+function resolveChannelAndVersion(positional: string | undefined): {
+  channel: ReleaseChannel;
+  versionArg: string | undefined;
+} {
```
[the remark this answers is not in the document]
BYK: Fixed in 709cd50: `resolveChannelAndVersion` now normalizes the positional to lowercase before comparing, so `'Nightly'`, `'STABLE'`, etc. all work as channel selectors.

.github/workflows/ci.yml
```diff
@@ -282,17 +300,79 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: package.json
+      - name: Compute nightly version
+        # Uses the commit timestamp — the same value as build-binary — so
+        # version.json exactly matches the version baked into the binaries.
+        id: version
+        env:
+          COMMIT_TIMESTAMP: ${{ github.event.head_commit.timestamp }}
```
BYK: I think we can define this at workflow level?

.github/workflows/ci.yml
```diff
@@ -178,6 +178,24 @@ jobs:
...
+        shell: bash
+        env:
+          COMMIT_TIMESTAMP: ${{ github.event.head_commit.timestamp }}
+        run: |
+          node -e "
+            const fs = require('fs');
+            const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+            const ts = Math.floor(new Date(process.env.COMMIT_TIMESTAMP).getTime() / 1000);
+            pkg.version = pkg.version.replace(/-dev\.\d+$/, '-dev.' + ts);
+            fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n');
```
BYK: I wonder if this would be simpler if we used `jq` and `awk` or something

.github/workflows/ci.yml
```diff
@@ -282,17 +300,79 @@ jobs:
...
+            --prerelease \
+            --notes "Latest nightly build from the \`main\` branch.
+
+          **Version:** \`${{ steps.version.outputs.version }}\`
+          **Commit:** ${{ github.sha }}"
+
+          # Upload/replace all .gz binaries and the version manifest
+          gh release upload nightly --clobber \
+            artifacts/dist-bin/*.gz \
+            version.json
```
BYK: I'd argue deleting existing assets first in case we change asset names in the future or remove certain ones

.github/workflows/ci.yml
BYK: Done in eb7bdd7: moved `COMMIT_TIMESTAMP` to workflow-level `env:` block so both `build-binary` and `publish-nightly` inherit it without repetition.

.github/workflows/ci.yml
BYK: Done in eb7bdd7: replaced `node -e` with `jq` + `sed`. Both the build-binary and publish-nightly steps now use:
```bash
TS=$(date -d "$COMMIT_TIMESTAMP" +%s)
CURRENT=$(jq -r .version package.json)
VERSION=$(echo "$CURRENT" | sed "s/-dev\\.[0-9]*$/-dev.${TS}/")
```
The build-binary step includes a BSD date fallback for macOS runners.

.github/workflows/ci.yml
BYK: Done in eb7bdd7: now deletes all existing nightly release assets before uploading:
```bash
gh release view nightly --json assets --jq '.assets[].name' | while read -r name; do
  gh release delete-asset nightly "$name" --yes
done
```
This ensures removed or renamed assets don't linger.

PR #297, opened 2026-02-26
feat(formatters): render all terminal output as markdown

BYK:
## Summary

- Replace all ad-hoc ANSI/chalk formatting with a **markdown-first pipeline**: formatters build CommonMark strings internally, which are rendered through a custom marked-terminal renderer in TTY mode or emitted as raw markdown in non-TTY/piped mode
- **Plain output mode**: auto-detected via `process.stdout.isTTY`; overridable with `SENTRY_PLAIN_OUTPUT` or `NO_COLOR`. TTY → styled Unicode tables + ANSI colors. Non-TTY → raw CommonMark (great for piping, CI, AI agents)
- **Issue list SHORT IDs** are markdown links rendered as OSC 8 terminal hyperlinks (clickable in iTerm2, VS Code, Windows Terminal, etc.) with alias characters bold-highlighted (e.g. `CLI-`**`W`**`EBSITE-`**`4`**)
- **Seer output** (root cause analysis, fix plans) renders fenced code blocks with syntax highlighting and structured heading hierarchy

image

## Key changes

- Add `src/lib/formatters/markdown.ts` — `renderMarkdown()`, `renderInlineMarkdown()`, `isPlainOutput()`, `escapeMarkdownCell()`, `mdRow()`, `mdKvTable()`, `mdTableHeader()`, `divider()` helpers
- All detail formatters (`formatIssueDetails`, `formatEventDetails`, `formatOrgDetails`, `formatProjectDetails`, `formatLogDetails`, `formatTraceSummary`, `formatRootCauseList`, `formatSolution`) return rendered `string` instead of `string[]`
- `writeTable()` processes cell values through `renderInlineMarkdown()` so `**bold**` and `[text](url)` in cells render as styled/clickable text; markdown links become OSC 8 hyperlinks
- `formatShortId()` returns markdown bold (`**text**`) instead of ANSI for alias highlighting — composable with link syntax
- Streaming row formatters (`formatLogRow`, `formatTraceRow`) are dual-mode: padded ANSI text in TTY, raw markdown table rows in plain mode
- Fix SDK name/exception value/org name escaping so special characters (`_`, `` ` ``, `|`, `\`) don't break markdown rendering

### DIFF
--- a/src/commands/log/list.ts
+++ b/src/commands/log/list.ts
@@ -132,7 +132,7 @@ async function executeSingleFetch(
   // Reverse for chronological order (API returns newest first, tail shows oldest first)
   const chronological = [...logs].reverse();
 
-  stdout.write(`${formatLogTable(chronological)}\n`);
+  stdout.write(formatLogTable(chronological));
 
   // Show footer with tip if we hit the limit
   const hasMore = logs.length >= flags.limit;
--- a/src/commands/trace/list.ts
+++ b/src/commands/trace/list.ts
@@ -151,7 +151,7 @@ export const listCommand = buildListCommand("trace", {
     }
 
     stdout.write(`Recent traces in ${org}/${project}:\n\n`);
-    stdout.write(`${formatTraceTable(traces)}\n`);
+    stdout.write(formatTraceTable(traces));
 
     // Show footer with tip
     const hasMore = traces.length >= flags.limit;


src/lib/formatters/log.ts
```diff
@@ -55,29 +60,72 @@ function formatTimestamp(timestamp: string): string {
...
+      : "";
+    return `| ${timestamp} | ${severity} | ${message}${trace} |\n`;
+  }
+
   const timestamp = formatTimestamp(log.timestamp);
   const severity = formatSeverity(log.severity);
   const message = log.message ?? "";
   const trace = log.trace ? muted(` [${log.trace.slice(0, 8)}]`) : "";
-
   return `${timestamp}  ${severity}  ${message}${trace}\n`;
```
BYK: I think there should be more code reuse in this section. I expect `isPlainOutput()` check to only change whether we use `renderInlineMarkdown()` or not once we prepare the plain output or one step before that. Like we can have an array of cell values in plain (markdown) format, then if `!isPlainOutput()` we do `array.map(renderInlineMarkdown)` or something like that.

src/lib/formatters/log.ts
```diff
@@ -55,29 +60,72 @@ function formatTimestamp(timestamp: string): string {
...
+ * Build a markdown table for a list of log entries.
+ *
+ * Pre-rendered ANSI codes in cell values (e.g. colored severity) are preserved.
+ *
+ * @param logs - Log entries to display
+ * @returns Rendered terminal string with Unicode-bordered table
+ */
+export function formatLogTable(logs: SentryLog[]): string {
+  const header = "| Timestamp | Level | Message |";
+  const separator = "| --- | --- | --- |";
```
BYK: Would probably be better if we extract `const header` into a module-level constant array: `const LOG_TABLE_COLS = ["Timestamp", "Level", "Message"]` and then build the separator dynamically based on that.

src/lib/formatters/log.ts
```diff
@@ -92,127 +140,140 @@ function formatSeverityLabel(severity: string | null | undefined): string {
...
 
-  // Header
-  const headerText = `Log ${logId.slice(0, 12)}...`;
-  const separatorWidth = Math.max(
-    MIN_HEADER_WIDTH,
-    Math.min(80, headerText.length)
-  );
-  lines.push(headerText);
-  lines.push(muted("═".repeat(separatorWidth)));
+  lines.push(`## Log \`${logId.slice(0, 12)}...\``);
```
BYK: Would probably be better if we `...`'d the middle portion instead of the end like `abc...123` instead of `abcdef...`. I think we do this with token masking in `auth status` command

src/lib/formatters/log.ts
```diff
@@ -92,127 +140,140 @@ function formatSeverityLabel(severity: string | null | undefined): string {
...
-  lines.push(`Timestamp:  ${formatTimestamp(log.timestamp)}`);
-  lines.push(`Severity:   ${formatSeverityLabel(log.severity)}`);
+  // Core fields table
+  const rows: string[] = [];
+  rows.push(`| **ID** | \`${logId}\` |`);
+  rows.push(`| **Timestamp** | ${formatTimestamp(log.timestamp)} |`);
+  rows.push(`| **Severity** | ${formatSeverityLabel(log.severity)} |`);
+
+  lines.push("| | |");
+  lines.push("|---|---|");
```
BYK: We really should switch to array based cell/row construction

src/lib/formatters/markdown.ts
BYK, line 1: I feel like `human` and `markdown` formatters should be the same?

src/lib/formatters/trace.ts
```diff
@@ -41,15 +47,21 @@ export function formatTraceDuration(ms: number): string {
...
  *
- * @returns Header line with column titles and divider
+ * In plain mode: emits a markdown table header + separator row.
+ * In rendered mode: emits an ANSI-muted text header with a rule separator.
+ *
+ * @returns Header string (includes trailing newline)
  */
 export function formatTracesHeader(): string {
+  if (isPlainOutput()) {
+    return "| Trace ID | Transaction | Duration | When |\n| --- | --- | ---: | --- |\n";
```
BYK: Again, dynamic, array based columnts/cells

src/lib/formatters/trace.ts
```diff
@@ -41,15 +47,21 @@ export function formatTraceDuration(ms: number): string {
...
+  }
   const header = muted(
     "TRACE ID                          TRANSACTION                    DURATION    WHEN"
   );
-  return `${header}\n${divider(96)}\n`;
+  return `${header}\n${muted("─".repeat(96))}\n`;
```
BYK: I think the `divider(n)` helper was better than writing `"-".repeat(n)`. More descriptive and also if we wanted to change a common thing, it's a single entry point

src/lib/formatters/trace.ts
```diff
@@ -64,10 +76,22 @@ const DURATION_WIDTH = 10;
...
  * @param item - Transaction list item from the API
  * @returns Formatted row string with newline
  */
 export function formatTraceRow(item: TransactionListItem): string {
+  if (isPlainOutput()) {
+    const traceId = renderInlineMarkdown(`\`${item.trace}\``);
+    const transaction = escapeMarkdownCell(item.transaction || "unknown");
+    const duration = formatTraceDuration(item["transaction.duration"]);
+    const when = formatRelativeTime(item.timestamp).trim();
+    return `| ${traceId} | ${transaction} | ${duration} | ${when} |\n`;
```
BYK: Same concern as earlier: I'd prefer we build everything in markdown and check `isPlainOutput()` at the end to decide whether to render or not

package.json
```diff
@@ -1,29 +1,9 @@
...
-    "test": "bun run test:unit && bun run test:isolated",
-    "test:unit": "bun test test/lib test/commands test/types --coverage --coverage-reporter=lcov",
-    "test:isolated": "bun test test/isolated",
-    "test:e2e": "bun test test/e2e",
-    "generate:skill": "bun run script/generate-skill.ts",
-    "check:skill": "bun run script/check-skill.ts",
-    "check:deps": "bun run script/check-no-deps.ts"
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/getsentry/cli.git"
```
BYK:
```suggestion
    "url": "git+https://github.com/getsentry/cli.git"
```

src/lib/formatters/log.ts
```diff
@@ -68,23 +73,17 @@ function formatTimestamp(timestamp: string): string {
...
-    const message = escapeMarkdownCell(log.message ?? "");
-    const trace = log.trace
-      ? ` ${renderInlineMarkdown(`\`[${log.trace.slice(0, 8)}]\``)}`
-      : "";
-    return `| ${timestamp} | ${severity} | ${message}${trace} |\n`;
+    return `| ${cells.join(" | ")} |\n`;
   }
 
-  const timestamp = formatTimestamp(log.timestamp);
-  const severity = formatSeverity(log.severity);
```
BYK, line 66: Even here you can do:

```ts
const row = isPlainOutput ? cells : cells.map(renderInlineMarkdown);
return `| ${row.join(" | ")} |\n`;
```

src/lib/formatters/log.ts
```diff
@@ -97,10 +96,12 @@ export function formatLogRow(log: SentryLog): string {
...
   if (isPlainOutput()) {
-    return "| Timestamp | Level | Message |\n| --- | --- | --- |\n";
+    return `${mdTableHeader(LOG_TABLE_COLS)}\n`;
   }
-  const header = muted("TIMESTAMP            LEVEL    MESSAGE");
-  return `${header}\n${muted("─".repeat(80))}\n`;
+  const header = renderInlineMarkdown(
+    LOG_TABLE_COLS.map((c) => `**${c}**`).join("  ")
+  );
+  return `${header}\n${divider(80)}\n`;
```
BYK: This is a bit confusing: so if it is not plain output, we actually don't show a regular table?

src/lib/formatters/log.ts
BYK: Cell array based row construction here for the values too?

src/lib/formatters/markdown.ts
```diff
@@ -127,6 +128,34 @@ export function escapeMarkdownCell(value: string): string {
...
+/**
+ * Build a markdown table header row + separator from column definitions.
+ *
+ * @param cols - Column definitions: `[name]` or `[name, "right"]` for right-align
+ * @returns Two-line string: `| A | B |\n| --- | ---: |`
+ */
+export function mdTableHeader(
+  cols: ReadonlyArray<string | readonly [string, "right"]>
+): string {
+  const names = cols.map((c) => (typeof c === "string" ? c : c[0]));
```
BYK: Type checking every column name is very inefficient. Can we not have special syntax like `#` at the start or end of column name to indicate it should be right-aligned? Or some other existing convention if there's one?

src/lib/formatters/trace.ts
```diff
@@ -84,23 +86,17 @@ const DURATION_WIDTH = 10;
...
-  const traceId = item.trace.slice(0, TRACE_ID_WIDTH).padEnd(TRACE_ID_WIDTH);
-  const transaction = (item.transaction || "unknown")
-    .slice(0, MAX_TRANSACTION_LENGTH)
-    .padEnd(MAX_TRANSACTION_LENGTH);
-  const duration = formatTraceDuration(item["transaction.duration"]).padStart(
-    DURATION_WIDTH
-  );
-  const when = formatRelativeTime(item.timestamp);
-  return `${traceId}  ${transaction}  ${duration}  ${when}\n`;
+  return `| ${cells.map((c) => renderInlineMarkdown(c)).join(" | ")} |\n`;
```
BYK: Same as my earlier comment, these could be joined even more. In fact, this looks like a pattern that we can probably just extract into a helper?

src/lib/formatters/markdown.ts
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * @param cells - Cell values (may contain inline markdown)
+ * @returns `| a | b |\n`
+ */
+export function mdRow(cells: readonly string[]): string {
+  const out = isPlainOutput()
+    ? cells
+    : cells.map((c) => renderInlineMarkdown(c));
+  return `| ${out.join(" | ")} |\n`;
+}
```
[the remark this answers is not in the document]
BYK: This is already handled: in rendered TTY mode, `mdRow` calls `renderInlineMarkdown()` per cell then runs `.replace(/\|/g, "│")` on the result (added in a previous commit). Any `|` characters — including those that `marked.parseInline` unescapes from `\|` — are replaced with U+2502 before the row is assembled, so they cannot break column alignment.

src/lib/formatters/human.ts
```diff
@@ -301,163 +234,45 @@ export function formatRelativeTime(dateString: string | undefined): string {
...
+ * Abbreviate large numbers with K/M/B/T/P/E suffixes (up to 10^18).
  *
  * The decimal is only shown when the rounded value is < 100 (e.g. "12.3K",
- * "1.5M" but not "100M"). The result is always exactly COL_COUNT chars wide.
- *
- * Note: `Number(raw)` loses precision above `Number.MAX_SAFE_INTEGER`
- * (~9P / 9×10^15), which is far beyond any realistic Sentry event count.
+ * "1.5M" but not "100M").
  *
- * Examples: 999 → "  999", 12345 → "12.3K", 150000 → " 150K", 1500000 → "1.5M"
```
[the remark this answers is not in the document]
BYK: The `/` extraction was part of the original code but was missing in an intermediate version. It was restored in this PR at [`human.ts:306-308`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/human.ts#L306-L308):

```ts
const aliasPart = projectAlias.includes("/")
  ? (projectAlias.split("/").pop() ?? projectAlias)
  : projectAlias;
```

This correctly strips the org prefix from cross-org collision aliases (e.g. `"o1/d"` → `"d"`) before matching against the short ID parts.

src/lib/formatters/human.ts
```diff
@@ -301,163 +233,45 @@ export function formatRelativeTime(dateString: string | undefined): string {
...
 /**
- * Abbreviate large numbers to fit within {@link COL_COUNT} characters.
- * Uses K/M/B/T/P/E suffixes up to 10^18 (exa).
+ * Abbreviate large numbers with K/M/B/T/P/E suffixes (up to 10^18).
  *
  * The decimal is only shown when the rounded value is < 100 (e.g. "12.3K",
- * "1.5M" but not "100M"). The result is always exactly COL_COUNT chars wide.
- *
- * Note: `Number(raw)` loses precision above `Number.MAX_SAFE_INTEGER`
- * (~9P / 9×10^15), which is far beyond any realistic Sentry event count.
```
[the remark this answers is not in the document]
BYK: The `/` extraction was part of the original code but was missing in an intermediate version. It was restored in this PR at [`human.ts:306-308`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/human.ts#L306-L308):

```ts
const aliasPart = projectAlias.includes("/")
  ? (projectAlias.split("/").pop() ?? projectAlias)
  : projectAlias;
```

This correctly strips the org prefix from cross-org collision aliases (e.g. `"o1/d"` → `"d"`) before matching against the short ID parts.

src/lib/formatters/human.ts
```diff
@@ -301,163 +232,45 @@ export function formatRelativeTime(dateString: string | undefined): string {
...
```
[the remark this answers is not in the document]
BYK: The `/` extraction was part of the original code but was missing in an intermediate version. It was restored in this PR at [`human.ts:306-308`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/human.ts#L306-L308):

```ts
const aliasPart = projectAlias.includes("/")
  ? (projectAlias.split("/").pop() ?? projectAlias)
  : projectAlias;
```

This correctly strips the org prefix from cross-org collision aliases (e.g. `"o1/d"` → `"d"`) before matching against the short ID parts.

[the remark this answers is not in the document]
BYK: The `/` extraction was part of the original code but was missing in an intermediate version. It was restored in this PR at [`human.ts:306-308`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/human.ts#L306-L308):

```ts
const aliasPart = projectAlias.includes("/")
  ? (projectAlias.split("/").pop() ?? projectAlias)
  : projectAlias;
```

This correctly strips the org prefix from cross-org collision aliases (e.g. `"o1/d"` → `"d"`) before matching against the short ID parts.

src/lib/formatters/human.ts
```diff
@@ -301,163 +237,45 @@ export function formatRelativeTime(dateString: string | undefined): string {
...
+ * "1.5M" but not "100M").
  *
- * Examples: 999 → "  999", 12345 → "12.3K", 150000 → " 150K", 1500000 → "1.5M"
```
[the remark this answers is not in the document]
BYK: The `/` extraction was part of the original code but was missing in an intermediate version. It was restored in this PR at [`human.ts:306-308`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/human.ts#L306-L308):

```ts
const aliasPart = projectAlias.includes("/")
  ? (projectAlias.split("/").pop() ?? projectAlias)
  : projectAlias;
```

This correctly strips the org prefix from cross-org collision aliases (e.g. `"o1/d"` → `"d"`) before matching against the short ID parts.

src/lib/formatters/markdown.ts
```diff
@@ -0,0 +1,428 @@
...
+ * @param md - Inline markdown text
+ * @returns Styled string (TTY) or raw markdown text (non-TTY / plain mode)
+ */
+export function renderInlineMarkdown(md: string): string {
+  if (isPlainOutput()) {
+    return md;
+  }
+  const tokens = marked.lexer(md);
+  return renderInline(tokens.flatMap(flattenInline));
+}
```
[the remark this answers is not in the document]
BYK: `renderInlineMarkdown` is used in production at [`table.ts:97`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/table.ts#L97):

```ts
const rows = items.map((item) =>
  columns.map((c) => renderInlineMarkdown(c.value(item)))
);
```

`writeTable` calls it to render each cell value in rendered (TTY) mode for every table in the CLI (issue list, project list, org list, etc.). The export is intentional so `table.ts` can import it from `markdown.ts`.

src/lib/formatters/markdown.ts
```diff
@@ -0,0 +1,496 @@
...
+  if (m) {
+    const tagName = m[1];
+    const inner = m[2];
+    if (tagName !== undefined && inner !== undefined) {
+      const colorFn = COLOR_TAGS[tagName.toLowerCase()];
+      return colorFn ? colorFn(inner) : inner;
+    }
+  }
+  return "";
+}
```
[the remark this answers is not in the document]
BYK: This is a false positive. The `renderInline` function at [`markdown.ts:329-364`](https://github.com/getsentry/cli/blob/feat/markdown-terminal-rendering/src/lib/formatters/markdown.ts#L329-L364) handles paired color tags correctly using an open/close buffering approach:

1. When it encounters an open `html` token matching `<red>` etc., it buffers all following tokens into `inner[]`
2. When it finds the matching close tag `</red>`, it applies the color function to the recursively-rendered inner content
3. This handles the case `marked` emits: separate `html` tokens for open/close with text/inline tokens in between

The `RE_SELF_TAG` in `renderHtmlToken` is only a fallback for self-contained `<tag>text</tag>` in a single token (which marked rarely produces inline). The primary path is the paired-tag buffering loop in `renderInline`. I verified this works correctly in TTY mode — color tags applied by `colorTag()` render with ANSI colors as expected.

PR #298, opened 2026-02-26
feat(nightly): distribute via GHCR instead of GitHub Releases

BYK:
## Summary

GitHub Releases prevents a rolling `nightly` tag — assets can't be modified and tags can't be reused after deletion. GHCR supports freely overwritable OCI artifact tags, making it the right host for nightly binaries.

This PR replaces the nightly distribution channel with GHCR while leaving stable releases (GitHub Releases + Craft) untouched.

## Changes

**CI (`publish-nightly` job)**
- Compresses binaries with gzip, generates `0.0.0-nightly.<unix_timestamp>` version
- Pushes `.gz` binaries to `ghcr.io/getsentry/cli:nightly` via ORAS with the version in the manifest `annotations.version` field
- Tag is overwritten on every main-branch push; unchanged files are deduplicated by GHCR

**`src/lib/ghcr.ts`** (new)
- `getAnonymousToken()` — anonymous token exchange with ghcr.io
- `fetchNightlyManifest(token)` — fetches the `:nightly` OCI manifest
- `getNightlyVersion(manifest)` — extracts version from manifest annotations (2 requests total for a version check: token + manifest)
- `downloadNightlyBlob(token, digest)` — downloads a blob with manual 307 redirect; auth header must **not** be forwarded to the Azure Blob Storage redirect target (returns 404 otherwise)

**`src/lib/upgrade.ts`**
- `fetchLatestNightlyVersion()` now fetches from GHCR manifest annotation instead of GitHub `version.json`
- `downloadBinaryToTemp()` routes nightly versions through `downloadNightlyToPath()` (GHCR) and stable through `downloadStableToPath()` (GitHub Releases)

**`src/lib/version-check.ts`**
- Background version check uses GHCR for nightly channel, GitHub Releases for stable

**`install` script**
- `--nightly` flag triggers a 6-step GHCR flow using only `curl` and `awk` (no `jq`)
- `--nightly` and `--version` are mutually exclusive

## Notes

- The `ghcr.io/getsentry/cli` package already exists and is public (one-time setup done)
- Stable release distribution (GitHub Releases, Homebrew, npm) is unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,13 +55,25 @@ jobs:
       - name: Compute build matrix
         id: targets
         run: |
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            # PRs only need linux-x64 for smoke test and e2e — skip macOS/Windows
-            echo 'matrix={"include":[{"target":"linux-x64","os":"ubuntu-latest","can-test":true}]}' >> "$GITHUB_OUTPUT"
-          else
-            # main, release/**, workflow_call: full cross-platform matrix
-            echo 'matrix={"include":[{"target":"darwin-arm64","os":"macos-latest","can-test":true},{"target":"linux-x64","os":"ubuntu-latest","can-test":true},{"target":"windows-x64","os":"windows-latest","can-test":true},{"target":"darwin-x64","os":"macos-latest","can-test":false},{"target":"linux-arm64","os":"ubuntu-latest","can-test":false}]}' >> "$GITHUB_OUTPUT"
-          fi
+          {
+            echo 'matrix<<MATRIX_EOF'
+            if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+              # PRs only need linux-x64 for smoke test and e2e — skip macOS/Windows
+              echo '{"include":[
+                {"target":"linux-x64", "os":"ubuntu-latest", "can-test":true}
+              ]}'
+            else
+              # main, release/**, workflow_call: full cross-platform matrix
+              echo '{"include":[
+                {"target":"darwin-arm64",  "os":"macos-latest",  "can-test":true},
+                {"target":"linux-x64",     "os":"ubuntu-latest", "can-test":true},
+                {"target":"windows-x64",   "os":"windows-latest","can-test":true},
+                {"target":"darwin-x64",    "os":"macos-latest",  "can-test":false},
+                {"target":"linux-arm64",   "os":"ubuntu-latest", "can-test":false}
+              ]}'
+            fi
+            echo 'MATRIX_EOF'
+          } >> "$GITHUB_OUTPUT"
       - name: Compute nightly version
         id: nightly
         if: github.ref == 'refs/heads/main' && github.event_name == 'push'
--- a/script/build.ts
+++ b/script/build.ts
@@ -99,20 +99,18 @@ async function buildTarget(target: BuildTarget): Promise<boolean> {
 
   console.log(`    -> ${outfile}`);
 
-  // On main and release branches (RELEASE_BUILD=1), binpunch and compress.
-  // Skipped on PRs where only linux-x64 is built for smoke/e2e tests.
-  if (process.env.RELEASE_BUILD) {
-    // Hole-punch: zero unused ICU data entries so they compress to nearly nothing.
-    // Must run before gzip so the compressed output benefits from zeroed regions.
-    const hpStats = processBinary(outfile);
-    if (hpStats && hpStats.removedEntries > 0) {
-      console.log(
-        `    -> hole-punched ${hpStats.removedEntries}/${hpStats.totalEntries} ICU entries`
-      );
-    }
+  // Hole-punch: zero unused ICU data entries so they compress to nearly nothing.
+  // Always runs so the smoke test exercises the same binary as the release.
+  const hpStats = processBinary(outfile);
+  if (hpStats && hpStats.removedEntries > 0) {
+    console.log(
+      `    -> hole-punched ${hpStats.removedEntries}/${hpStats.totalEntries} ICU entries`
+    );
+  }
 
-    // Create gzip-compressed copies for release downloads / GHCR nightly.
-    // With hole-punch, reduces download size by ~70% (99 MB → 28 MB).
+  // On main and release branches (RELEASE_BUILD=1), create gzip-compressed
+  // copies for release downloads / GHCR nightly (~70% smaller with hole-punch).
+  if (process.env.RELEASE_BUILD) {
     const binary = await Bun.file(outfile).arrayBuffer();
     const compressed = await gzipAsync(Buffer.from(binary), { level: 6 });
     await Bun.write(`${outfile}.gz`, compressed);


.github/workflows/ci.yml
```diff
@@ -216,6 +221,73 @@ jobs:
...
 
+  publish-nightly:
+    name: Publish Nightly to GHCR
+    # Only run on pushes to main, not on PRs or release branches
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    needs: [build-binary]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
BYK: Is this needed since we already have workflow-level perms defined?

.github/workflows/ci.yml
```diff
@@ -216,6 +221,73 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
```
BYK:
```suggestion
      - uses: actions/checkout@v6
```

Also, do we even need to check out??

.github/workflows/ci.yml
```diff
@@ -216,6 +221,73 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: sentry-*
+          path: artifacts
+          merge-multiple: true
+
+      - name: Install ORAS CLI
+        run: |
+          VERSION=1.2.3
+          curl -sfL "https://github.com/oras-project/oras/releases/download/v${VERSION}/oras_${VERSION}_linux_amd64.tar.gz" \
```
BYK: Let's also add hash verification for this, just in case?

.github/workflows/ci.yml
```diff
@@ -216,6 +221,73 @@ jobs:
...
+        run: |
+          ls artifacts/
+          for f in artifacts/sentry-linux-* artifacts/sentry-darwin-*; do
+            gzip -k "$f"
+          done
+          # Windows binary — compress but keep .exe name visible in title
+          for f in artifacts/sentry-windows-*.exe; do
+            gzip -k "$f"
+          done
+          ls artifacts/*.gz
```
BYK: Why don't we have compressed binaries already? I thought the build job was producing them.

src/lib/upgrade.ts
```diff
@@ -327,50 +327,51 @@ export async function fetchLatestFromNpm(): Promise<string> {
...
-
-  if (!response.ok) {
-    throw new UpgradeError(
-      "network_error",
-      `Failed to fetch nightly version: ${response.status}`
-    );
+  // AbortSignal is not threaded through ghcr helpers, but checking it before
+  // each network call ensures we bail out promptly when the process exits.
+  if (signal?.aborted) {
+    throw Object.assign(new Error("AbortError"), { name: "AbortError" });
```
BYK: can we create a proper class for this?

src/lib/upgrade.ts
```diff
@@ -458,13 +459,115 @@ async function streamDecompressToFile(
...
+ */
+function getNightlyGzFilename(): string {
+  let os: string;
+  if (process.platform === "darwin") {
+    os = "darwin";
+  } else if (process.platform === "win32") {
+    os = "windows";
+  } else {
+    os = "linux";
+  }
```
BYK: Can we use a mapping for this so we can do `platformName[process.platform]`?

.github/workflows/ci.yml
```diff
@@ -216,6 +221,73 @@ jobs:
...
+
+      - name: Compute nightly version
+        id: version
+        run: |
+          # Derive a deterministic Unix timestamp from the commit so that all
+          # jobs in this workflow run produce the same version string.
+          TS=$(date -d '${{ github.event.head_commit.timestamp }}' +%s 2>/dev/null \
+               || date -j -f "%Y-%m-%dT%H:%M:%S%z" '${{ github.event.head_commit.timestamp }}' +%s 2>/dev/null \
+               || echo ${{ github.run_number }})
+          VERSION="0.0.0-nightly.${TS}"
```
BYK: What prompted us to switch from `-dev.` format to `-nightly.` format? I think `-dev.` is better.

install
```diff
@@ -13,7 +13,8 @@ Usage: install [options]
 
 Options:
     -h, --help              Display this help message
-    -v, --version <version> Install a specific version (e.g., 0.2.0) or "nightly"
+    -v, --version <version> Install a specific version (e.g., 0.2.0)
+        --nightly           Install the latest nightly build from GHCR
```
BYK: Why change this? Just keep the old `--version nightly` thing

test/commands/cli/upgrade.test.ts
```diff
@@ -577,36 +667,68 @@
...
+          }
         );
       }
-      // Binary download — return gzipped content for all other URLs
-      return new Response(gzipped, { status: 200 });
+      if (urlStr.includes("/blobs/sha256:abc123")) {
+        // Redirect to blob storage (GHCR blob endpoint returns 307)
+        return Response.redirect("https://blob.example.com/file.gz", 307);
+      }
+      if (urlStr.includes("blob.example.com")) {
```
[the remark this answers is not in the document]
BYK: False positive — this is test-only mock routing code that dispatches fake responses based on URL patterns. The `blob.example.com` domain is a made-up test host simulating Azure Blob Storage redirects from GHCR. No user input flows through this code path and there is no production URL validation happening here.

test/commands/cli/upgrade.test.ts
```diff
@@ -668,17 +790,46 @@
...
+            },
+          }
         );
       }
-      // binary download (.gz)
-      if (urlStr.includes(".gz")) {
+      if (urlStr.includes("/blobs/sha256:abc456")) {
+        return Response.redirect("https://blob.example.com/nightly.gz", 307);
+      }
+      if (urlStr.includes("blob.example.com")) {
```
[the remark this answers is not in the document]
BYK: Same as the other `blob.example.com` alert — false positive. Test-only mock routing code for a fake host, no production URL validation.

.github/workflows/ci.yml
BYK: Removed. The workflow-level `permissions` at lines 14-16 already grants `packages: write` to all jobs including `publish-nightly`.

.github/workflows/ci.yml
BYK: Removed entirely — the job doesn't need the repo. The nightly version is derived from the `COMMIT_TIMESTAMP` env var + a `curl` fetch of `package.json` from the raw GitHub content API (using `${{ github.sha }}`). Not bumping to `@v6` since all other jobs in this workflow use `@v4` — changing just one would be inconsistent.

.github/workflows/ci.yml
BYK: Added. Downloads to a file first, verifies `sha256sum`, then extracts:
```
EXPECTED_SHA256="b4efc97a91f471f323f193ea4b4d63d8ff443ca3aab514151a30751330852827"
echo "${EXPECTED_SHA256}  ${TARBALL}" | sha256sum -c -
```
Hash sourced from `oras_1.2.3_checksums.txt` on the ORAS GitHub release.

.github/workflows/ci.yml
BYK: The `build-binary` job produces raw binaries — compression for stable releases happens in Craft during publishing, not during build (moving it would change artifact contents for e2e tests and Craft). The `publish-nightly` job needs `.gz` for efficient OCI layers, so it compresses here. Added a comment to make this explicit.

src/lib/upgrade.ts
BYK: Added an `AbortError` class to `src/lib/errors.ts` (where all other error classes live):
```ts
export class AbortError extends Error {
  override name = "AbortError" as const;
  constructor() {
    super("The operation was aborted");
  }
}
```
Both `Object.assign` sites now do `throw new AbortError()`. Chose this over `DOMException` — while available in Bun/Node, it's a browser/DOM API and semantically wrong for a CLI tool.

src/lib/upgrade.ts
BYK: Done — replaced the if/else chain with a `Record` lookup:
```ts
const platformNames: Record<string, string> = {
  darwin: "darwin",
  win32: "windows",
};
const os = platformNames[process.platform] ?? "linux";
```

.github/workflows/ci.yml
BYK: Agreed — keeping `-dev.`. The `build-binary` job already bakes `X.Y.Z-dev.\<timestamp\>` into each binary (via `sed "s/-dev\.[0-9]*$/-dev.${TS}/"`). The `publish-nightly` version computation now derives the same string using the same `COMMIT_TIMESTAMP` env var, so the GHCR manifest annotation always matches the version baked inside the binary. Updated `isNightlyVersion` to check for `-dev.` and updated all tests accordingly.

install
BYK: Reverted. Removed the `--nightly` flag; `--version nightly` is the special value again. The GHCR download path is now triggered by `if [[ "$requested_version" == "nightly" ]]`.

.github/workflows/ci.yml
BYK: Correction to my earlier reply — the build script `script/build.ts` **does** already produce `.gz` files in CI (line 113: `if (process.env.CI)`). It runs `binpunch` (zeroes unused ICU data for better compression) then gzips each binary at level 6. These `.gz` files are included in the uploaded artifacts via `dist-bin/sentry-*`.

Removed the compress step entirely — it was redundant and would have re-gzipped the raw binaries without the binpunch benefit.

.github/workflows/ci.yml
```diff
@@ -243,38 +238,26 @@ jobs:
...
-          VERSION="0.0.0-nightly.${TS}"
+          # The build-binary job bakes X.Y.Z-dev.<TS> into each binary using the
+          # same COMMIT_TIMESTAMP env var. Re-derive the version here so the GHCR
+          # manifest annotation matches the version string inside the binary.
+          TS=$(date -d "$COMMIT_TIMESTAMP" +%s)
+          CURRENT=$(curl -sf "https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/package.json" \
+                    | awk -F'"' '/"version"/{print $4}')
+          VERSION=$(echo "$CURRENT" | sed "s/-dev\.[0-9]*$/-dev.${TS}/")
           echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
           echo "Nightly version: ${VERSION}"
```
BYK: Let's make this part of the regular builds as otherwise when you run `sentry --version` it would show `0.13.0-dev.0` or something like that as the TS part is not baked into the binary.

src/lib/upgrade.ts
```diff
@@ -468,14 +469,11 @@ async function streamDecompressToFile(
...
-    os = "darwin";
-  } else if (process.platform === "win32") {
-    os = "windows";
-  } else {
-    os = "linux";
-  }
+  const platformNames: Record<string, string> = {
+    darwin: "darwin",
+    win32: "windows",
+  };
```
BYK: This should be a module level constant with `Object.freeze()`. I'd also put `"linux"` with a keyname like `"default"` or make it a constant too like `DEFAULT_PLATFORM`

.github/workflows/ci.yml
BYK: Good call. Moved the nightly version computation into the `changes` job (runs once, early). It outputs `nightly-version` which is consumed by:
- **build-binary**: `Set nightly version` step now just does `jq --arg v "${{ needs.changes.outputs.nightly-version }}"` — no more re-deriving the timestamp per matrix leg
- **publish-nightly**: reads `needs.changes.outputs.nightly-version` directly for the GHCR annotation — no more `curl` to raw.githubusercontent.com

Single source of truth, computed once.

src/lib/upgrade.ts
BYK: Done. Extracted to a frozen module-level constant with `DEFAULT_PLATFORM`:
```ts
const PLATFORM_NAMES: Readonly<Record<string, string>> = Object.freeze({
  darwin: "darwin",
  win32: "windows",
});

const DEFAULT_PLATFORM = "linux";
```
Usage: `PLATFORM_NAMES[process.platform] ?? DEFAULT_PLATFORM`

.github/workflows/ci.yml
```diff
@@ -50,6 +52,25 @@ jobs:
...
+      - name: Compute build matrix
+        id: targets
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            # PRs only need linux-x64 for smoke test and e2e — skip macOS/Windows
+            echo 'matrix={"include":[{"target":"linux-x64","os":"ubuntu-latest","can-test":true}]}' >> "$GITHUB_OUTPUT"
+          else
+            # main, release/**, workflow_call: full cross-platform matrix
+            echo 'matrix={"include":[{"target":"darwin-arm64","os":"macos-latest","can-test":true},{"target":"linux-x64","os":"ubuntu-latest","can-test":true},{"target":"windows-x64","os":"windows-latest","can-test":true},{"target":"darwin-x64","os":"macos-latest","can-test":false},{"target":"linux-arm64","os":"ubuntu-latest","can-test":false}]}' >> "$GITHUB_OUTPUT"
+          fi
```
BYK: I wonder if we can make the matrix definition more readable and less fragile (as they are jsons in strings)

.github/workflows/ci.yml
```diff
@@ -215,23 +214,32 @@ jobs:
...
-          path: dist-bin/sentry-*
+          path: |
+            dist-bin/sentry-*
+            !dist-bin/*.gz
+
+      - name: Upload compressed artifact
+        if: env.RELEASE_BUILD == '1'
+        uses: actions/upload-artifact@v4
+        with:
+          name: sentry-${{ matrix.target }}-gz
```
BYK, line 241: This name change would require a craft config change in `.craft.yml`

script/build.ts
```diff
@@ -99,18 +99,20 @@ async function buildTarget(target: BuildTarget): Promise<boolean> {
...
+  // On main and release branches (RELEASE_BUILD=1), binpunch and compress.
+  // Skipped on PRs where only linux-x64 is built for smoke/e2e tests.
+  if (process.env.RELEASE_BUILD) {
+    // Hole-punch: zero unused ICU data entries so they compress to nearly nothing.
+    // Must run before gzip so the compressed output benefits from zeroed regions.
+    const hpStats = processBinary(outfile);
+    if (hpStats && hpStats.removedEntries > 0) {
+      console.log(
+        `    -> hole-punched ${hpStats.removedEntries}/${hpStats.totalEntries} ICU entries`
+      );
```
BYK: On second thought, I think we should `binpunch` regardless as otherwise the smoke test would not catch any corruption that might occur during this phase.

.github/workflows/ci.yml
BYK: Agreed — switched to a multiline heredoc format using the `<<MATRIX_EOF` delimiter syntax for `$GITHUB_OUTPUT`. Each target entry is on its own line with aligned columns:
```json
{"include":[
  {"target":"darwin-arm64",  "os":"macos-latest",  "can-test":true},
  {"target":"linux-x64",     "os":"ubuntu-latest", "can-test":true},
  ...
]}
```

.github/workflows/ci.yml
BYK: Checked — `.craft.yml` works as-is without changes. Craft's artifact regex `/^sentry-.*$/` matches both `sentry-linux-x64` (raw) and `sentry-linux-x64-gz` (compressed). Craft downloads all matching artifacts and flattens the files inside. The brew target's `includeNames: '/^sentry-(darwin|linux)-(arm64|x64)\.gz$/'` correctly filters to only the `.gz` files from the `-gz` artifacts. The `github` target uploads all files to the Release. No config change needed.

script/build.ts
BYK: Good point. Moved binpunch back outside the `RELEASE_BUILD` gate — it always runs now so the smoke test exercises the exact same binary that ships. Only gzip creation remains gated on `RELEASE_BUILD`.

PR #306, opened 2026-02-27
feat(issue-list): global limit with fair distribution, compound cursor, and richer progress

BYK:
## Summary

- `--limit` is now a **global** cap across all detected projects, not per-project. 10 projects with `--limit 25` fetches at most 25 issues total, not 250.
- Budget is distributed evenly across projects (`ceil(limit / N)` per project). Projects that return fewer issues than their share give their surplus to projects that have more (two-phase fetch with redistribution).
- After merging results, at least one issue per project is guaranteed before filling remaining slots by global sort — so quiet projects aren't hidden by high-volume ones.
- `-c last` now works in multi-target mode via a **compound cursor**: each project's resume cursor is JSON-encoded into a single stored string, so the next page picks up exactly where each project left off.
- Progress messages change from the misleading `5/250` format to `"Fetching issues from 10 projects, 5 and counting (up to 25)..."`.
- JSON output for multi-target mode now always emits `{data: [...], hasMore: bool}`, aligning with org-all mode.

## Changed files

- `src/lib/api-client.ts` — `listIssuesAllPages` gains optional `startCursor` to resume mid-stream.
- `src/commands/issue/list.ts` — new helpers (`fetchWithBudget`, `runPhase2`, `trimWithProjectGuarantee`, compound cursor encode/decode), updated `handleResolvedTargets` and `handleOrgAllIssues`.
- `test/commands/issue/list.test.ts` — two tests updated for new JSON schema.

### DIFF
--- a/src/commands/issue/list.ts
+++ b/src/commands/issue/list.ts
@@ -937,7 +937,15 @@ async function handleResolvedTargets(
       }
       return `${r.data.target.org}/${r.data.target.project}` === key;
     });
-    return result?.success ? (result.data.nextCursor ?? null) : null;
+    if (result?.success) {
+      // Successful fetch: null = exhausted (no more pages), string = has more
+      return result.data.nextCursor ?? null;
+    }
+    // Target failed this fetch — preserve the cursor it was given so the next
+    // `-c last` retries from the same position rather than skipping it entirely.
+    // If no start cursor was given (first-page failure), null means not retried
+    // via cursor; the user can run without -c last to restart all projects.
+    return startCursors.get(key) ?? null;
   });
   const hasAnyCursor = cursorValues.some((c) => c !== null);
   if (hasAnyCursor) {


src/commands/issue/list.ts
```diff
@@ -423,6 +432,262 @@ async function fetchIssuesForTarget(
...
+          typeof e.project === "string" &&
+          (e.cursor === null || typeof e.cursor === "string")
+      )
+    ) {
+      return parsed as ProjectCursorEntry[];
+    }
+  } catch {
+    // Fall through to return empty array
+  }
+  return [];
```
BYK: Looks like all this information can be extracted from the context key or the fingerprint and this can just be a long string where multiple cursors are joined by `-` or some other special character?

BYK:
Addressed both review comments in a8dca67:

**Comment 1 (BugBot - dispatch guard blocks compound cursor):** Real bug, fixed. Added `allowCursorInModes?: readonly ParsedOrgProject["type"][]` to `DispatchOptions` in `org-list.ts`. The guard now checks `!cursorAllowedModes.includes(parsed.type)` instead of hard-coding `parsed.type !== "org-all"`. The `issue list` command passes `allowCursorInModes: ["auto-detect", "explicit", "project-search"]`, making compound cursor pagination in `handleResolvedTargets` fully reachable. Updated the unit test that expected `ValidationError` — it now verifies the command completes without error when cursor is used in explicit mode.

**Comment 2 (BugBot - project guarantee uses slug only):** Real bug, fixed. Added `orgSlug: string` to `IssueWithOptions`, populated from `result.target.org` in `attachFormatOptions` and from the `org` param in the org-all path. `trimWithProjectGuarantee` now uses `org/project` as the per-project key (same pattern as `buildProjectAliasMap` and `attachFormatOptions` already use), so two orgs with identically-named projects are correctly treated as distinct.

BYK:
Fixed in 3b1d1a4.

**Seer (trimWithProjectGuarantee uses projectSlug only):** Real bug, fixed properly. Three changes:
1. Added `orgSlug: string` field to the `IssueWithOptions` type (it was set inline on the org-all path but absent from the type definition, so `attachFormatOptions` didn't populate it for multi-target mode either).
2. Updated `attachFormatOptions` to include `orgSlug: result.target.org` in the returned objects.
3. Updated `trimWithProjectGuarantee` to use ```\`${issues[i]!.orgSlug}/${issues[i]!.formatOptions.projectSlug ?? ""}```` as the per-project key, so `org-one/dashboard` and `org-two/dashboard` are treated as distinct projects and each gets a guaranteed slot.

BYK:
Fixed in b9c3a7a.

**BugBot (failed targets become permanently exhausted in compound cursor):** Real bug, fixed.

Root cause: line 940 returned `null` for any target not found in `results` (i.e., failed fetches). Since `null` in the compound cursor means "exhausted — skip on next `-c last`", a transient failure (e.g., 500) would permanently evict that project from the pagination session until the 5-minute cursor TTL expired.

Fix: when a target is not in `results` (fetch failed), re-encode the start cursor it was given (`startCursors.get(key)`) so the next `-c last` retries from the same position. Only successful fetches with no `nextCursor` produce `null` (truly exhausted). First-page failures with no start cursor still produce `null` (unrecoverable via cursor — the user retries with a fresh run).

PR #316, opened 2026-03-02
fix(region): resolve DSN org prefix at resolution layer

BYK:
## Problem

When a user or AI agent extracts an org identifier from a Sentry DSN host (e.g. `o1081365` from `o1081365.ingest.us.sentry.io`) and passes it as a CLI argument (`sentry event view o1081365/ <event-id>`), the CLI forwards it as-is to the API. The Sentry API accepts org slugs or bare numeric IDs via `organization_id_or_slug`, but not the `o`-prefixed DSN form — resulting in a `ResolutionError`.

## Approach

Instead of eagerly stripping the `o` prefix at the arg-parsing layer (which transforms raw user input), this fix applies normalization at the **resolution layer** as a best-effort fallback:

1. `resolveEffectiveOrg()` in `src/lib/region.ts` first tries the raw org identifier
2. If resolution fails and the identifier matches the DSN pattern (`/^o(\d+)$/`), it strips the prefix and retries with the numeric ID
3. This preserves raw input through the parsing pipeline and only transforms when necessary

### Key changes

- Add `resolveEffectiveOrg()` to `src/lib/region.ts` with cache-aware fast path and `AuthError`-safe fallback
- Move `stripDsnOrgPrefix()` to `src/lib/dsn/parser.ts` (co-located with DSN parsing logic)
- Remove normalization from `arg-parsing.ts` — parser preserves raw input unchanged
- Apply `resolveEffectiveOrg()` at 7 call sites across `event/view`, `issue/utils`, `org-list`, and `resolve-target`
- Update tests to expect raw `oNNNNN` values from parser


src/lib/dsn/parser.ts
```diff
@@ -209,3 +209,35 @@ export function createDsnFingerprint(dsns: DetectedDsn[]): string {
...
+ *
+ * @example
+ * stripDsnOrgPrefix("o1081365")  // "1081365"
+ * stripDsnOrgPrefix("o123")      // "123"
+ * stripDsnOrgPrefix("sentry")    // "sentry"  (no change)
+ * stripDsnOrgPrefix("organic")   // "organic" (no change — not all digits after 'o')
+ */
+export function stripDsnOrgPrefix(org: string): string {
+  return org.replace(DSN_ORG_PREFIX_PATTERN, "$1");
+}
```
BYK, line 234: You know this can be achieved by just doing `org.slice(1)` or something right? Using a `.replace()` with a regex substitiuon is overkill

src/lib/dsn/parser.ts
```diff
@@ -209,3 +209,35 @@ export function createDsnFingerprint(dsns: DetectedDsn[]): string {
...
 }
+
+/**
+ * Pattern matching bare DSN-style org identifiers: "o" followed by only digits.
+ *
+ * This is the bare-identifier counterpart of {@link ORG_ID_PATTERN} (which
+ * matches the full ingest host). Used to normalize org identifiers that
+ * were extracted from a DSN host and passed as CLI arguments.
+ */
+const DSN_ORG_PREFIX_PATTERN = /^o(\d+)$/;
```
BYK: This is overlapping functionality over `extractOrgIdFromHost`. We should just reuse that.

src/lib/dsn/parser.ts
BYK: Fixed in 02ef204 — removed `DSN_ORG_PREFIX_PATTERN` entirely and rewrote to delegate to `extractOrgIdFromHost` via a synthetic ingest hostname:

```ts
export function stripDsnOrgPrefix(org: string): string {
  return extractOrgIdFromHost(`${org}.ingest.sentry.io`) ?? org;
}
```

Single source of truth for the `oNNNNN` pattern, no regex duplication.

src/lib/dsn/parser.ts
BYK: Fixed in the same commit (02ef204) — `stripDsnOrgPrefix` now delegates to `extractOrgIdFromHost` instead of using its own regex pattern.

src/lib/region.ts
```diff
@@ -78,3 +79,80 @@ export function isMultiRegionEnabled(): boolean {
...
+  const afterRefresh = await getOrgRegion(orgSlug);
+  if (afterRefresh) {
+    return orgSlug;
+  }
+
+  // Retry: check numeric ID after refresh
+  const refreshedSlug = await lookupDsnOrgInCache(orgSlug);
+  if (refreshedSlug) {
+    return refreshedSlug;
+  }
```
BYK: Would it be possible to abstract this portion out instead of repeating?

BYK: Done in 7901274 — extracted `resolveOrgFromCache()` that handles both the slug lookup and DSN numeric ID lookup. `resolveEffectiveOrg()` now reads as a clean two-phase flow: try cache → refresh API → retry cache → fallback.

src/lib/region.ts
```diff
@@ -78,3 +79,71 @@ export function isMultiRegionEnabled(): boolean {
...
+  // Cache is cold or identifier is unknown — refresh the org list.
+  // listOrganizations() populates org_regions with slug, region, and org_id.
+  try {
+    const { listOrganizations } = await import("./api-client.js");
+    await listOrganizations();
+  } catch (error) {
+    if (error instanceof AuthError) {
+      return orgSlug;
+    }
+    throw error;
```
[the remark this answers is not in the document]
BYK: Valid point — `resolveEffectiveOrg` is a best-effort normalization helper, so any error from the API refresh step should fall back to the original slug rather than propagating. Fixed in 1e62bf9: simplified the catch to a bare `catch {}` that returns `orgSlug` regardless of error type. The downstream API call will produce a relevant error if the slug is genuinely invalid or the service is unavailable.

PR #320, opened 2026-03-02
feat(api): add --data/-d flag and auto-detect JSON body in fields

BYK:
## Summary

Fixes [CLI-AF](https://sentry.sentry.io/issues/CLI-AF) — users passing raw JSON as `--raw-field` now get auto-correction instead of a cryptic error.

## Problem

A user ran:
```bash
sentry api /api/0/issues/7303827789/ -X PUT -f '{"status":"ignored","statusDetails":{"ignoreCount":1}}'
```
and got: `Error: Invalid field format: {...}. Expected key=value`

Worse, on current main, `normalizeFields` would silently *mangle* the JSON by converting the first `:` to `=`, producing `{"status"="ignored",...}` — silent data corruption sent to the API.

### 1. `--data`/`-d` flag (like curl)

New explicit inline JSON body flag, following curl convention:

```bash
sentry api issues/123/ -X PUT -d '{"status":"resolved"}'
```

- Tries `JSON.parse()`, falls back to raw string (like `--input`)
- Mutually exclusive with `--input` (clear `ValidationError` if both used)

### 2. JSON auto-detection in field values

When someone passes `-f '{"status":"ignored"}'`, we detect it's JSON (no `=`, starts with `{`/`[`, valid `JSON.parse`), use it as the request body, and hint:

```
hint: '{"status":"ignored"}' was used as the request body. Use --data/-d to pass inline JSON next time.
```

If there are also normal `key=value` fields, they're merged into the JSON body. Multiple bare JSON fields throw a `ValidationError`.

### 3. `normalizeFields` JSON guard

Skip colon→equals auto-correction for strings starting with `{` or `[`. This fixes the **silent data corruption bug** where JSON was mangled into garbage.

## Tests

30 new tests across 4 describe blocks:
- `normalizeFields` — JSON passthrough (objects, arrays, mixed with other fields)
- `parseDataBody` — valid JSON, arrays, nested, invalid, partial
- `extractJsonBody` — extraction + hint, remaining fields, invalid JSON, multiple JSON error, key=value with JSON value, preview truncation
- `buildFromFields` — CLI-AF scenario, merge JSON + fields, GET params routing, no-JSON passthrough

All 2312 unit tests pass. Typecheck and lint clean.

### DIFF
--- a/src/commands/api.ts
+++ b/src/commands/api.ts
@@ -943,6 +943,53 @@ export function buildFromFields(
   return { body, params: options.params };
 }
 
+/**
+ * Resolve the request body and query params from the user-provided flags.
+ *
+ * Priority order: `--data` > `--input` > field flags (`-F`/`-f`).
+ * Mutually-exclusive combinations throw {@link ValidationError}.
+ *
+ * @returns body and params ready for the API request
+ * @internal Exported for testing
+ */
+export async function resolveBody(
+  flags: Pick<ApiFlags, "method" | "data" | "input" | "field" | "raw-field">,
+  stdin: NodeJS.ReadStream & { fd: 0 },
+  stderr: Writer
+): Promise<{
+  body?: Record<string, unknown> | unknown[] | string;
+  params?: Record<string, string | string[]>;
+}> {
+  if (flags.data !== undefined && flags.input !== undefined) {
+    throw new ValidationError(
+      "Cannot use --data and --input together. " +
+        "Use --data/-d for inline JSON, or --input for file/stdin.",
+      "data"
+    );
+  }
+
+  if (
+    flags.data !== undefined &&
+    (flags.field?.length || flags["raw-field"]?.length)
+  ) {
+    throw new ValidationError(
+      "Cannot use --data with --field or --raw-field. " +
+        "Use --data/-d for a full JSON body, or -F/-f for individual fields.",
+      "data"
+    );
+  }
+
+  if (flags.data !== undefined) {
+    return { body: parseDataBody(flags.data) };
+  }
+
+  if (flags.input !== undefined) {
+    return { body: await buildBodyFromInput(flags.input, stdin) };
+  }
+
+  return buildFromFields(flags.method, flags, stderr);
+}
+
 // Command Definition
 
 export const apiCommand = buildCommand({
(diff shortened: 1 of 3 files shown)


src/commands/api.ts
```diff
@@ -333,6 +334,13 @@ export function normalizeFields(
       return field;
     }
 
+    // JSON-shaped strings (starting with { or [) must not be "corrected" —
+    // the colon inside is JSON syntax, not a key:value separator.  Let the
+    // downstream pipeline handle it (extractJsonBody or processField error).
+    if (field.startsWith("{") || field.startsWith("[")) {
+      return field;
+    }
```
BYK, line 342: Why not just do a `JSON.parse(field)` in a `try`-`catch` block??

src/commands/api.ts
```diff
@@ -600,6 +608,100 @@ export function parseHeaders(headers: string[]): Record<string, string> {
...
+ * otherwise returns `undefined`.
+ * @internal
+ */
+function tryParseJsonField(field: string): unknown | undefined {
+  if (field.includes("=")) {
+    return;
+  }
+  if (!(field.startsWith("{") || field.startsWith("["))) {
+    return;
+  }
```
BYK, line 649: This looks redundant with the `JSON.parse()` below?

src/commands/api.ts
BYK: The `startsWith` guard is intentionally broader than `JSON.parse`. We want to skip colon-correction for *any* JSON-shaped string — even malformed ones like `{broken: value}`. With `JSON.parse`, malformed JSON would fail, fall through to the colon-correction, and become `{broken= value}` — equally wrong and potentially more confusing. The conservative `startsWith` check prevents mangling anything that *looks like it was intended as JSON*, regardless of validity.

src/commands/api.ts
BYK: You are right — the `startsWith` check in `tryParseJsonField` was redundant. `JSON.parse` in the `catch` branch handles non-JSON strings anyway; the check was just a minor early-exit optimisation that added noise. Removed in the latest commit.

src/commands/api.ts
```diff
@@ -766,6 +868,45 @@ export function handleResponse(
...
+  const options = prepareRequestOptions(method, field, rawField);
+  if (options.body) {
+    // Merge field-built key=value entries into the auto-detected JSON body
+    body =
+      body && typeof body === "object" && !Array.isArray(body)
+        ? { ...body, ...options.body }
+        : (options.body ?? body);
+  }
+
+  return { body, params: options.params };
```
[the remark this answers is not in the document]
BYK: Fixed in the latest commit: `buildFromFields` now skips `extractJsonBody` when `method === "GET"`. A JSON-shaped raw-field on a GET request falls through to `buildRawQueryParams`, which throws a clear `ValidationError` ("Invalid field format") rather than silently attaching a body to a GET request. No hint is emitted since the extraction was never attempted.

src/commands/api.ts
```diff
@@ -766,6 +868,45 @@ export function handleResponse(
...
+  rawField = extracted.remaining;
+
+  // Route remaining fields to body (merge) or params based on HTTP method
+  const options = prepareRequestOptions(method, field, rawField);
+  if (options.body) {
+    // Merge field-built key=value entries into the auto-detected JSON body
+    body =
+      body && typeof body === "object" && !Array.isArray(body)
+        ? { ...body, ...options.body }
+        : (options.body ?? body);
```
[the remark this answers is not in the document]
BYK: Fixed in the latest commit: before merging, `buildFromFields` now checks for top-level key overlap between the auto-detected JSON body and the field-parsed body. If any key appears in both (e.g. `statusDetails` in the JSON body *and* `statusDetails[minCount]=5` as a field flag), a `ValidationError` is thrown pointing the user to `--data/-d`. This prevents the shallow spread from silently dropping nested data from the original JSON body.

src/commands/api.ts
```diff
@@ -600,6 +608,97 @@ export function parseHeaders(headers: string[]): Record<string, string> {
...
+  if (field.includes("=")) {
+    return;
+  }
+
+  try {
+    return JSON.parse(field);
+  } catch {
+    return;
+  }
+}
```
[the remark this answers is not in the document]
BYK: BugBot is right — this was a real bug introduced when I removed the `startsWith` guard. The guard was not just an optimisation: it also restricted detection to objects and arrays. Without it, `tryParseJsonField("42")` returns `42`, which gets set as `body`. Then the key-conflict check does `k in (42 as Record<…>)` → `TypeError` at runtime.

Fixed in the latest commit: restored the `startsWith` guard with an explicit JSDoc explaining *why* it exists (both optimisation and semantic restriction to objects/arrays only). Also updated the `extractJsonBody` docs and added a test covering that primitives (`42`, `true`, `"string"`, `null`) are never extracted.

src/commands/api.ts
BYK: Correction to my earlier reply: the check was NOT redundant. BugBot caught the consequence — removing it allowed `tryParseJsonField` to return primitives like `42` or `true`, causing a downstream `TypeError` when the `in` operator was used against a non-object body. The `startsWith` guard serves a dual purpose: optimisation *and* semantic restriction to objects/arrays only. Restored in the latest commit with the JSDoc updated to make this clear.

PR #326, opened 2026-03-03
fix(setup): suppress shell completion messages on upgrade

BYK:
## Problem

Shell completion messages were printed after every upgrade, even when completions were already installed. On supported shells this was `Completions: Updated /path/...`; on unsupported shells with a bash fallback it printed `Your shell (xonsh) is not directly supported` + `Updated bash completions as a fallback` every single time. Users on unsupported shells without bash got `Not supported for X shell` on every upgrade.

`installCompletions()` already returned a `created: boolean` flag — it just wasn't being used to gate output.

## Changes

- **`tryBashCompletionFallback()`** — return type changed from `string | null` to `CompletionLocation | null`, exposing the `created` flag to the caller
- **`handleCompletions()`** — returns `[]` (no output) when `created === false`; this covers supported shells, the bash fallback path, and the no-completion-possible path
- First-time installs still print the same messages as before
- Updated existing test for the unsupported-shell-no-bash case; added an explicit upgrade-scenario test asserting silence on the second run

### DIFF
--- a/src/commands/cli/setup.ts
+++ b/src/commands/cli/setup.ts
@@ -136,19 +136,19 @@ async function handlePathModification(
  * @param pathEnv - The PATH to search for bash, forwarded from the process env.
  * @returns The completion location if bash is available, null otherwise.
  */
-function tryBashCompletionFallback(
+async function tryBashCompletionFallback(
   homeDir: string,
   xdgDataHome: string | undefined,
   pathEnv: string | undefined
 ): Promise<CompletionLocation | null> {
   if (!isBashAvailable(pathEnv)) {
-    return Promise.resolve(null);
+    return null;
   }
 
   // Defensive: installCompletions returns null only if the shell type has no
   // completion script or path configured. "bash" is always supported, but
   // we guard here in case that changes in future.
-  return installCompletions("bash", homeDir, xdgDataHome);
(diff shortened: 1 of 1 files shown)


src/commands/cli/setup.ts
```diff
@@ -131,25 +134,21 @@ async function handlePathModification(
...
-    // completion script or path configured. "bash" is always supported, but
-    // we guard here in case that changes in future.
-    return null;
-  }
-  const action = fallback.created ? "Installed" : "Updated";
-  return `      ${action} bash completions as a fallback: ${fallback.path}`;
+  // Defensive: installCompletions returns null only if the shell type has no
+  // completion script or path configured. "bash" is always supported, but
+  // we guard here in case that changes in future.
+  return installCompletions("bash", homeDir, xdgDataHome);
```
BYK: Shouldn't we await `installCompletions` for proper stack trace generation on error?
```suggestion
  return await installCompletions("bash", homeDir, xdgDataHome);
```

BYK: Good catch! Applied in the follow-up commit — restored `async` and changed to `return await installCompletions(...)` so the stack trace includes `tryBashCompletionFallback` when the write fails.

PR #329, opened 2026-03-04
feat(log-list): add --trace flag to filter logs by trace ID

### LINKED ISSUE #317: sentry logs: add --trace / --trace-id filter flag
## Summary
Users should be able to filter logs by trace ID directly from the CLI via `sentry logs --trace <id>` (or `--trace-id <id>`). Currently navigating from a trace to its associated logs requires going through the UI or running a separate query manually.

## Analysis
- Current behavior: `sentry logs` does not accept a trace ID filter flag.
- Gap: trace-centric workflows are common; users often start from a trace and want to inspect associated logs without leaving the CLI.
- Direction: add `--trace` / `--trace-id` flag to `sentry logs` that scopes the query to the given trace ID; also consider `sentry trace logs <id>` as a subcommand alias for discoverability.

## Concerns
- Decide on flag name consistency: `--trace` vs `--trace-id` — pick one canonical form and alias the other.

BYK:
Add `--trace` flag to `sentry log list` that scopes the query to a specific trace ID.

When `--trace` is given, the command uses the dedicated trace-logs endpoint (`/organizations/{org}/trace-logs/`) instead of the project-scoped events endpoint, since the Explore/Events logs API does not support trace ID filtering in query syntax. This means only org resolution is needed — the positional target is treated as an org slug, not an org/project pair.

### Changes
- `--trace` flag with 32-char hex validation on `sentry log list`
- Works in both single-fetch and `--follow` streaming modes
- Trace follow mode polls the trace-logs endpoint and deduplicates by `timestamp_precise`
- Refactored `writeLogs()` to accept both `SentryLog` and `TraceLog` via shared `LogLike` type
- Trace-logs mock route + 5 E2E tests

### Usage
```bash
sentry log list --trace abc123def456abc123def456abc123de
sentry log list myorg --trace abc123def456abc123def456abc123de
sentry log list myorg --trace abc123def456abc123def456abc123de -f
sentry log list myorg --trace abc123def456abc123def456abc123de --json
```

Closes #317


src/commands/log/list.ts
```diff
@@ -141,120 +170,191 @@ async function executeSingleFetch(
...
   }
 
-  // In TTY mode, use a bordered StreamingTable for aligned columns.
-  // In plain mode, use raw markdown rows for pipe-friendly output.
   const plain = flags.json || isPlainOutput();
   const table = plain ? undefined : createLogStreamingTable();
 
-  // Track if header has been printed (for human/plain mode)
   let headerPrinted = false;
+  let lastTimestamp = Date.now() * 1_000_000;
```
BYK, line 229: Why are we multiplying milliseconds by 1 million??

src/commands/log/list.ts
```diff
@@ -141,120 +170,191 @@ async function executeSingleFetch(
...
+        if (initialLogs[0]) {
+          lastTimestamp = initialLogs[0].timestamp_precise;
+        }
+        scheduleNextPoll();
+      })
+      .catch((error: unknown) => {
+        process.removeListener("SIGINT", stop);
+        reject(error);
+      });
+  });
```
BYK, line 323: Why are we not using `async`/`await` here??

src/commands/log/list.ts
```diff
@@ -325,18 +436,82 @@ export const listCommand = buildListCommand("log", {
...
+              statsPeriod: "1m",
+            }),
+          fetchPoll: (lastTs) =>
+            listLogs(org, project, {
+              query: flags.query,
+              limit: flags.limit,
+              statsPeriod: "10m",
+              afterTimestamp: lastTs,
+            }),
+          extractNew: (logs) => logs,
```
BYK, line 495: This call seems to be duplicated, any way to DRY it?

src/commands/log/list.ts
BYK: `timestamp_precise` is in nanoseconds (per Sentry API), `Date.now()` returns milliseconds. 1 ms = 1,000,000 ns. Added an inline comment explaining the conversion in c1f7d3b1.

src/commands/log/list.ts
BYK: Refactored to async/await IIFE inside the Promise constructor in c1f7d3b1. The Promise constructor is still needed because SIGINT calls `resolve()` externally, but the fetch now uses `await` + `try/catch` instead of `.then()/.catch()`.

src/commands/log/list.ts
BYK: DRYed in c1f7d3b1. Replaced `fetchInitial` + `fetchPoll` with a single `fetch(statsPeriod, afterTimestamp?)` callback in `FollowConfig`. `executeFollowMode` calls `config.fetch("1m")` for initial and `config.fetch("10m", lastTimestamp)` for polls. Each caller now provides one function instead of two near-identical ones.

src/commands/log/list.ts
```diff
@@ -141,120 +170,195 @@ async function executeSingleFetch(
...
+    return;
+  }
+
+  const chronological = [...logs].reverse();
+  stdout.write(formatLogTable(chronological, false));
+
+  const hasMore = logs.length >= flags.limit;
+  const countText = `Showing ${logs.length} log${logs.length === 1 ? "" : "s"} for trace ${traceId}.`;
+  const tip = hasMore ? " Use --limit to show more." : "";
+  writeFooter(stdout, `${countText}${tip}`);
```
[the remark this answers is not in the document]
BYK: These two functions live in separate commands (`log list --trace` vs `trace logs`) and have intentional differences: different `statsPeriod` sources (hardcoded 14d vs user-supplied `--period` flag), different empty-state help messages (suggesting different commands), and `trace/logs.ts` does not reverse the array for JSON output. Extracting a shared helper would over-couple two independent command files for marginal DRY benefit. The display pattern (fetch → JSON branch → empty branch → format table → footer) is intentionally co-located with each command for readability.

src/commands/log/list.ts
```diff
@@ -289,9 +293,9 @@ function executeFollowMode<T extends LogLike>(
       }
     }
 
-    config
-      .fetchInitial()
-      .then((initialLogs) => {
+    (async () => {
```
BYK: I'm sorry why the hell are we wrapping this in an `async` function and then immediately calling it? Is this because we don't want to `await` it? If yes then revert to the old version and add a comment about why

BYK: You're right, the async IIFE was awkward. Reverted to `.then()/.catch()` and added a comment explaining why we cannot `await` here: `resolve` must remain callable by the SIGINT handler (`stop`) at any time, so the initial fetch is fire-and-forget. Fixed in fffc2270.

src/commands/log/list.ts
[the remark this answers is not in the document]
BYK: I really want you to try your hand at abstraction and see if it actually ends up worse.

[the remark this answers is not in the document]
BYK: Done in 4ca38012. Extracted `displayTraceLogs()` into `src/lib/formatters/log.ts` — both commands now call it with an options object:

```ts
displayTraceLogs({
  stdout, logs, traceId, limit: flags.limit,
  asJson: flags.json,
  emptyMessage: "...",
});
```

The abstraction actually worked out well. Each caller still owns the fetch and the empty-state message (which differ), but JSON output, table rendering, and footer generation are shared. This also normalized `trace/logs.ts` JSON output to use chronological order (oldest-first), consistent with all other commands.

src/lib/formatters/log.ts
```diff
@@ -317,3 +323,49 @@ export function formatLogDetails(
...
+ *
+ * Handles JSON output, empty state, and human-readable table formatting.
+ * Used by both `sentry log list --trace` and `sentry trace logs`.
+ */
+export function displayTraceLogs(options: DisplayTraceLogsOptions): void {
+  const { stdout, logs, traceId, limit, asJson, emptyMessage } = options;
+
+  if (asJson) {
+    writeJson(stdout, [...logs].reverse());
+    return;
```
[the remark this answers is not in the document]
BYK: This is an intentional normalization, not a breaking change. `sentry trace logs` was introduced in PR #311 (not yet released), so there are no existing consumers relying on newest-first JSON output. Every other command in the CLI (`log list`, `log list --trace`, `issue list`, etc.) outputs JSON in chronological order. The old `trace logs` was the sole outlier. This change makes the behavior consistent across the CLI.

PR #330, opened 2026-03-04
feat: add response caching for read-only API calls

### LINKED ISSUE #318: Add response caching for read-only API calls
## Summary
The CLI should cache API responses to avoid redundant network calls and improve perceived performance, especially in interactive and iterative workflows. This was identified as a top usability concern after real-world usage.

## Analysis
- Current behavior: every command hits the Sentry API with no caching layer.
- Gap: repeated invocations (e.g. listing issues, browsing events) re-fetch identical data, adding latency and unnecessary load.
- Direction: implement a local response cache (e.g. TTL-based, keyed on URL + query params) with a reasonable default TTL; expose `--no-cache` / `--refresh` flags to bypass.
- Scope: read-only endpoints only (`GET`); mutations must never be cached.

## Concerns
- Cache invalidation strategy: time-based TTL is simplest; consider shorter TTLs for high-churn data like issues vs. longer for events/traces.
- Cache storage: filesystem (e.g. `~/.cache/sentry-cli/`) preferred over in-memory for persistence across invocations.

BYK:
## Summary

Implements RFC 7234/9111-aware HTTP response caching for GET requests to the Sentry API, integrated transparently at the fetch level inside `createAuthenticatedFetch()`.

Closes #318

## Cache Design

- **Filesystem-based** at `~/.sentry/cache/responses/` with JSON files keyed by SHA-256 hash of normalized URL + Authorization header
- **Standards-compliant** via `http-cache-semantics` for freshness evaluation
- **Tiered TTL system** based on URL pattern matching:
  | Tier | TTL | Examples |
  |------|-----|---------|
  | Immutable | 1 hour | `/events/{id}/`, `/issues/{id}/hashes/` |
  | Stable | 5 min | `/organizations/`, `/projects/`, `/teams/` |
  | Volatile | 60 sec | `/issues/` (lists), issue details |
  | No-cache | 0 sec | `/autofix/`, `/root-cause/` (polling) |
- **Probabilistic cleanup** (10% chance per write) with 500-file LRU cap
- **Only successful (2xx) responses** are cached

## New Flags & Env Vars

- `--refresh` / `-r` flag on all read commands — bypasses cache for that request
- `SENTRY_NO_CACHE=1` env var — disables caching globally
- Cache is automatically cleared on login and logout for security

## Commands Updated

All read-only commands now support `--refresh`: issue list/view/explain/plan, project list/view, org list/view, event view, trace list/view/logs, log list/view, auth whoami/status, team list, repo list.

## Testing

- **18 unit tests** for cache operations, TTL tiers, cleanup, and edge cases
- **15 property-based tests** for URL classification, key normalization, TTL bounds, and round-trip invariants
- **4 integration-style tests** for end-to-end cache behavior
- All existing tests pass (2604 unit + 100 isolated + 101 e2e)

### DIFF
--- a/src/lib/response-cache.ts
+++ b/src/lib/response-cache.ts
@@ -367,36 +367,48 @@ export async function getCachedResponse(
     return;
   }
 
+  const key = buildCacheKey(method, url);
+
   return await withCacheSpan(
-    "cache.lookup",
-    async () => {
-      const key = buildCacheKey(method, url);
+    url,
+    "cache.get",
+    async (span) => {
       const entry = await readCacheEntry(key);
       if (!entry) {
+        span.setAttribute("cache.hit", false);
         return;
       }
 
       try {
         const policy = CachePolicy.fromObject(entry.policy);
         if (!isEntryFresh(policy, entry, requestHeaders, url)) {
+          span.setAttribute("cache.hit", false);
           return;
         }
 
+        const body = JSON.stringify(entry.body);
+        span.setAttribute("cache.hit", true);
+        span.setAttribute("cache.item_size", body.length);
+
         const responseHeaders = buildResponseHeaders(policy, entry);
-        return new Response(JSON.stringify(entry.body), {
+        return new Response(body, {
           status: entry.status,
           headers: responseHeaders,
         });
       } catch {
         // Corrupted or version-incompatible policy object — treat as cache miss.
         // Best-effort cleanup of the broken entry.
+        span.setAttribute("cache.hit", false);
         unlink(cacheFilePath(key)).catch(() => {
           // Ignored — fire-and-forget
         });
         return;
       }
     },
-    { "cache.url": url }
+    {
+      "cache.key": [key],
+      "network.peer.address": getCacheDir(),
+    }
   );
 }
 
@@ -455,38 +467,60 @@ export async function storeCachedResponse(
     return;
   }
 
+  const key = buildCacheKey(method, url);
+
   try {
     await withCacheSpan(
-      "cache.store",
-      () => writeResponseToCache(method, url, requestHeaders, response),
-      { "cache.url": url }
+      url,
+      "cache.put",
+      async (span) => {
+        const size = await writeResponseToCache(
+          key,
+          url,
+          requestHeaders,
+          response
+        );
+        if (size > 0) {
+          span.setAttribute("cache.item_size", size);
+        }
+      },
+      {
+        "cache.key": [key],
+        "network.peer.address": getCacheDir(),
+      }
     );
   } catch {
     // Cache write failures are non-fatal — silently ignore
   }
 }
 
-/** Core cache write logic, separated for complexity management */
+/**
+ * Core cache write logic, separated for complexity management.
+ *
+ * Always called for GET requests (caller checks method), so "GET" is hardcoded
+ * for the CachePolicy constructor.
+ *
+ * @returns The serialized body size in bytes (0 if not storable).
+ */
 async function writeResponseToCache(
-  method: string,
+  key: string,
   url: string,
   requestHeaders: Record<string, string>,
   response: Response
-): Promise<void> {
+): Promise<number> {
   const responseHeadersObj = headersToObject(response.headers);
 
   const policy = new CachePolicy(
-    { url, method, headers: requestHeaders },
+    { url, method: "GET", headers: requestHeaders },
     { status: response.status, headers: responseHeadersObj },
     POLICY_OPTIONS
   );
 
   if (!policy.storable()) {
-    return;
+    return 0;
   }
 
   const body: unknown = await response.json();
-  const key = buildCacheKey(method, url);
   const now = Date.now();
 
   // Pre-compute expiry for cheap cleanup checks (avoids CachePolicy deserialization).
@@ -507,8 +541,9 @@ async function writeResponseToCache(
     expiresAt: now + ttl,
   };
 
+  const serialized = JSON.stringify(entry);
   await mkdir(getCacheDir(), { recursive: true, mode: 0o700 });
-  await writeFile(cacheFilePath(key), JSON.stringify(entry), "utf-8");
+  await writeFile(cacheFilePath(key), serialized, "utf-8");
 
   // Probabilistic cleanup to avoid unbounded cache growth
   if (Math.random() < CLEANUP_PROBABILITY) {
@@ -516,6 +551,8 @@ async function writeResponseToCache(
       // Non-fatal: cleanup failure doesn't affect cache correctness
     });
   }
+
+  return serialized.length;
 }
 
 /**
--- a/src/lib/telemetry.ts
+++ b/src/lib/telemetry.ts
@@ -952,20 +952,35 @@ export function withFsSpan<T>(
 }
 
 /**
- * Wrap a cache operation with a span for tracing.
+ * Wrap a cache operation with a Sentry Cache Module span.
  *
- * Creates a child span under the current active span to track
- * response cache hit/miss/store operations.
+ * Implements the [Sentry Cache Module spec](https://develop.sentry.dev/sdk/performance/modules/caches/)
+ * for the Caches Insights dashboard. The span is passed to the callback so
+ * callers can set `cache.hit`, `cache.item_size`, etc. after the lookup.
  *
- * @param operation - Name of the operation (e.g., "cache.lookup", "cache.store")
- * @param fn - The function that performs the cache operation
- * @param attributes - Optional span attributes (e.g., url, cache.hit)
+ * @param name - Span name (typically the cache key or a descriptive label)
+ * @param op - Cache operation: `"cache.get"` for reads, `"cache.put"` for writes
+ * @param fn - Function to execute, receives the span for dynamic attribute setting
+ * @param attributes - Initial span attributes (e.g., `cache.key`, `network.peer.address`)
  * @returns The result of the function
  */
 export function withCacheSpan<T>(
-  operation: string,
-  fn: () => T | Promise<T>,
-  attributes?: Record<string, string | number | boolean>
+  name: string,
+  op: "cache.get" | "cache.put",
+  fn: (span: Span) => T | Promise<T>,
+  attributes?: Record<string, string | number | boolean | string[]>
 ): Promise<T> {
-  return withTracing(operation, "cache", fn, attributes);
+  return Sentry.startSpan(
+    { name, op, attributes, onlyIfParent: true },
+    async (span) => {
+      try {
+        const result = await fn(span);
+        span.setStatus({ code: 1 }); // OK
+        return result;
+      } catch (error) {
+        span.setStatus({ code: 2 }); // Error
+        throw error;
+      }
+    }
+  );
 }


src/commands/auth/login.ts
```diff
@@ -90,6 +96,11 @@ export const loginCommand = buildCommand({
       return;
     }
 
+    // Clear stale cached responses from a previous session
```
BYK, line 115: Should we do the same for logout too?

src/commands/auth/status.ts
```diff
@@ -25,10 +25,13 @@ import {
   formatUserIdentity,
   maskToken,
 } from "../../lib/formatters/human.js";
+import { REFRESH_FLAG } from "../../lib/list-command.js";
```
BYK: Instead of "refresh", how about just "fresh"? Shorter and playful. Plus, we can make the short version -f if it's available?

src/commands/issue/explain.ts
```diff
@@ -61,13 +64,17 @@ export const explainCommand = buildCommand({
         brief: "Force new analysis even if one exists",
         default: false,
       },
+      refresh: REFRESH_FLAG,
```
BYK: Do you think we can have a "common flags" primitive or even a command builder where we automatically add this cache related flag to? Like we had for all list endpoints?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,572 @@
...
+  { pattern: /\/autofix\//, tier: "no-cache" },
+  { pattern: /\/root-cause\//, tier: "no-cache" },
+
+  // Immutable: specific resources by ID (events, traces)
+  { pattern: /\/events\/[^/?]+\/?(?:\?|$)/, tier: "immutable" },
+  { pattern: /\/trace\/[0-9a-f]{32}\//, tier: "immutable" },
+
+  // Volatile: list endpoints with high churn
+  { pattern: /\/issues\/?\?/, tier: "volatile" },
+  { pattern: /\/issues\/?$/, tier: "volatile" },
```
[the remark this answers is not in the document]
BYK: Fixed in 0fea830. Broadened the volatile URL pattern to match all `/issues/` URLs (including detail endpoints like `/issues/12345/`), not just list endpoints. The two narrow patterns `\/issues\/?\?` and `\/issues\/?$` are replaced with a single `\/issues\/` catch-all at the volatile tier. No-cache patterns (autofix, root-cause) still take priority since they come first in the ordered list. Added tests for detail URLs to confirm.

src/lib/sentry-client.ts
```diff
@@ -205,9 +210,45 @@ function extractUrlPath(input: Request | string | URL): string {
...
+ */
+async function tryCacheHit(
+  method: string,
+  fullUrl: string
+): Promise<Response | undefined> {
+  if (method !== "GET") {
+    return;
+  }
+  return await getCachedResponse(method, fullUrl, {});
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 0fea830. Both `tryCacheHit` and `cacheResponse` now accept and forward `requestHeaders` to the cache layer. For cache lookups, we use `getAuthToken()` (sync, no refresh overhead) to build the Authorization header. For cache storage after a successful fetch, we use the refreshed token. Extracted `authHeaders()` helper and `fetchWithRetry()` function to keep complexity under the biome limit.

src/commands/auth/login.ts
BYK: Already covered! `logout` calls `clearAuth()` (in `src/lib/db/auth.ts:108-111`) which already includes `clearResponseCache()` as a fire-and-forget call. So login clears the cache directly, and logout clears it via `clearAuth()`. No additional code needed.

src/commands/auth/status.ts
BYK: Love it! Renamed to `--fresh` / `-f` in 0fea830. The `-f` alias is available in all commands except `log list` where it's already taken by `--follow`. Renamed the constant from `REFRESH_FLAG` → `FRESH_FLAG` and the flag from `refresh:` → `fresh:` across all 16 commands.

src/commands/issue/explain.ts
BYK: Done in 0fea830! Created two primitives in `list-command.ts`:

1. **`FRESH_FLAG`** — the flag constant, replaces per-command definitions
2. **`applyFreshFlag(flags)`** — the helper that checks `flags.fresh` and calls `disableResponseCache()`

Commands now just do:
```ts
import { applyFreshFlag, FRESH_FLAG } from "../../lib/list-command.js";
// in parameters:
flags: { fresh: FRESH_FLAG },
aliases: { f: "fresh" },
// in func:
applyFreshFlag(flags);
```

This removes the direct `disableResponseCache` import from all 16 command files. `buildOrgListCommand` (used by team list, repo list) also includes it automatically with the `-f` alias.

src/lib/response-cache.ts
```diff
@@ -0,0 +1,572 @@
...
+  if (serverTtl > 0) {
+    // Server provided a TTL and it expired — respect the server
+    return false;
+  }
+
+  // No server TTL — use our fallback tier
+  const tier = classifyUrl(url);
+  const fallbackTtl = FALLBACK_TTL_MS[tier];
+  const age = Date.now() - entry.createdAt;
+  return age <= fallbackTtl;
```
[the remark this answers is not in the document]
BYK: Fixed in 408e85e. Two changes:
1. `isEntryFresh()`: Changed `serverTtl > 0` → `serverTtl !== 0` so negative server TTL (expired) returns false immediately instead of falling through to the longer fallback tier.
2. `collectEntryMetadata()`: Added explicit three-way logic: `> 0` → not expired (server says fresh), `< 0` → expired (server TTL elapsed), `=== 0` → no server TTL, use URL-based fallback tier.

src/lib/sentry-client.ts
```diff
@@ -205,9 +210,95 @@ function extractUrlPath(input: Request | string | URL): string {
...
+  const { token } = await refreshToken();
+  const headers = prepareHeaders(input, init, token);
+
+  for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
+    const isLastAttempt = attempt === MAX_RETRIES;
+    const result = await executeAttempt(input, init, headers, isLastAttempt);
+
+    if (result.action === "done") {
+      cacheResponse(method, fullUrl, authHeaders(token), result.response);
+      return result.response;
```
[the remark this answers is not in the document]
BYK: Fixed in 408e85e. `createAuthenticatedFetch()` now captures `getAuthToken()` at call time (not construction time), so the current token is always used when storing responses. The stored entry correctly reflects the auth context of the actual request.

src/lib/db/auth.ts
```diff
@@ -103,6 +104,11 @@ export function clearAuth(): void {
     db.query("DELETE FROM org_regions").run();
     db.query("DELETE FROM pagination_cursors").run();
   });
+
+  // Clear cached API responses — they are tied to the current user's permissions
+  clearResponseCache().catch(() => {
+    // Non-fatal: cache directory may not exist yet
+  });
```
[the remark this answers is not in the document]
BYK: Fixed in 8dcc111. `clearAuth()` is now `async` and `await`s `clearResponseCache()` with try/catch.

This uncovered a pre-existing bug in model-based tests: `fcAssert(asyncProperty(...))` returns a Promise that was never awaited (test functions were not `async`). Before this change it was invisible because all commands were essentially synchronous. With the real I/O yield from `await rm()`, the `createIsolatedDbContext()` cleanups interleaved with subsequent fast-check iterations, corrupting the `SENTRY_CONFIG_DIR` env var and causing reads from wrong databases.

Fixed all `asyncProperty` tests to properly `async`/`await`.

BYK:
**BugBot #9 (corrupted cache entry) — fixed in 85a119d.**

Wrapped `CachePolicy.fromObject()`, `isEntryFresh()`, `buildResponseHeaders()`, and the Response construction in a try-catch block inside `getCachedResponse()`. Corrupted or version-incompatible policy objects now:
1. Return `undefined` (cache miss) — the request proceeds normally via fetch
2. Best-effort delete the broken cache entry via `unlink()` so it doesn't keep failing

**BugBot #10 (async clearAuth not awaited) — fixed in 85a119d.**

Added `await` to both `clearAuth()` call sites in `test/commands/project/list.test.ts` (lines 1080 and 1362).

src/commands/auth/login.ts
```diff
@@ -90,6 +96,11 @@ export const loginCommand = buildCommand({
...
+    await clearResponseCache().catch(() => {
+      // Non-fatal: cache directory may not exist
+    });
```
BYK: Better not to mix await with callback syntax?
```suggestion
    try {
	    await clearResponseCache();
    catch {
      // Non-fatal: cache directory may not exist
	  }
```

We should also consider reporting these to Sentry?

src/commands/auth/whoami.ts
```diff
@@ -33,9 +35,12 @@ export const whoamiCommand = buildCommand({
         brief: "Output as JSON",
         default: false,
       },
+      fresh: FRESH_FLAG,
     },
+    aliases: { f: "fresh" },
   },
   async func(this: SentryContext, flags: WhoamiFlags): Promise<void> {
+    applyFreshFlag(flags);
```
BYK, line 47: Can we have a factory function or wrapper/decorater instead like we have for all `list` based functions? Still too much boilerplate to remember and do.

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+ * - `immutable`: data that never changes once created (events, traces)
+ * - `stable`: data that changes infrequently (orgs, projects, teams)
+ * - `volatile`: data that changes often (issue lists, log lists)
+ * - `no-cache`: never cache (polling endpoints like autofix state)
+ */
+type TtlTier = "immutable" | "stable" | "volatile" | "no-cache";
+
+/** Fallback TTL durations by tier (milliseconds). `no-cache` uses 0 as a sentinel. */
+const FALLBACK_TTL_MS: Record<TtlTier, number> = {
+  immutable: 60 * 60 * 1000, // 1 hour
```
BYK: If we know it will _never_ change why are we not caching it forever? Or at least up to a day or multiple days?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  { pattern: /\/events\/[^/?]+\/?(?:\?|$)/, tier: "immutable" },
+  { pattern: /\/trace\/[0-9a-f]{32}\//, tier: "immutable" },
+
+  // Volatile: issue endpoints (lists AND detail views — status/assignee change often)
+  { pattern: /\/issues\//, tier: "volatile" },
+  { pattern: /\/issues\/?$/, tier: "volatile" },
+  { pattern: /[?&]dataset=logs/, tier: "volatile" },
+  { pattern: /[?&]dataset=transactions/, tier: "volatile" },
+  { pattern: /\/trace-logs\//, tier: "volatile" },
+
```
BYK: Isn't this better as key-value pairs. Either the pattern or the tier can be the key based on how you wanna approach

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  { pattern: /\/autofix\//, tier: "no-cache" },
+  { pattern: /\/root-cause\//, tier: "no-cache" },
+
+  // Immutable: specific resources by ID (events, traces)
+  { pattern: /\/events\/[^/?]+\/?(?:\?|$)/, tier: "immutable" },
+  { pattern: /\/trace\/[0-9a-f]{32}\//, tier: "immutable" },
+
+  // Volatile: issue endpoints (lists AND detail views — status/assignee change often)
+  { pattern: /\/issues\//, tier: "volatile" },
+  { pattern: /\/issues\/?$/, tier: "volatile" },
```
BYK: You can probably combine these two in a single regex?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  { pattern: /[?&]dataset=logs/, tier: "volatile" },
+  { pattern: /[?&]dataset=transactions/, tier: "volatile" },
```
BYK: Should be easy to combine these two?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+ * Normalize method + URL into a stable string for cache key derivation.
+ * Sorts query params alphabetically for deterministic key generation.
+ *
+ * @internal Exported for testing
+ */
+export function normalizeUrl(method: string, url: string): string {
+  try {
+    const parsed = new URL(url);
+    const sortedParams = new URLSearchParams(
+      [...parsed.searchParams.entries()].sort(([a], [b]) => a.localeCompare(b))
```
BYK: Do we really need `localeCompare` for URL bits? They should all be ASCII?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+    );
+    parsed.search = sortedParams.toString()
+      ? `?${sortedParams.toString()}`
+      : "";
+    return `${method.toUpperCase()}|${parsed.toString()}`;
+  } catch {
+    // Malformed URL — use as-is
```
BYK: If we have a malformed URL we should already be throwing?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  url: string;
+  /** When this entry was created (epoch ms) */
+  createdAt: number;
+};
+
+/** CachePolicy options for a single-user CLI cache */
+const POLICY_OPTIONS: CachePolicy.Options = {
+  shared: false,
+  cacheHeuristic: 0.1,
+  immutableMinTimeToLive: 3_600_000,
```
BYK: This value should be linked to the one in `FALLBACK_TTL_MS` somehow, right?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  }
+  return result;
+}
+
+/** Convert Headers to a plain object for http-cache-semantics */
+function headersToObject(headers: Headers): Record<string, string> {
+  const obj: Record<string, string> = {};
+  headers.forEach((value, key) => {
+    obj[key] = value;
+  });
```
BYK: Isn't this `Object.fromEntries()`?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+
+    const responseHeaders = buildResponseHeaders(policy, entry);
+    return new Response(JSON.stringify(entry.body), {
+      status: entry.status,
+      headers: responseHeaders,
+    });
+  } catch {
+    // Corrupted or version-incompatible policy object — treat as cache miss.
+    // Delete the broken entry so it doesn't keep failing on every request.
+    await unlink(cacheFilePath(key)).catch(() => {
```
BYK: No need to `await` this, right? Also `.catch(() => {})` inside a `catch` block, really?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  const body: unknown = await response.json();
+  const key = buildCacheKey(method, url);
+
+  const entry: CacheEntry = {
+    policy: policy.toObject(),
+    body,
+    status: response.status,
+    headers: pickHeaders(response.headers),
+    url,
+    createdAt: Date.now(),
```
BYK: Instead of this, should we calculate `expiresAt` for cheaper clean up ops in the future? We may also get rid of the cache policy in this case?

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+
+/** Read all cache files and determine which are expired */
+async function collectEntryMetadata(
+  cacheDir: string,
+  jsonFiles: string[]
+): Promise<EntryMetadata[]> {
+  const entries: EntryMetadata[] = [];
+  const now = Date.now();
+
+  for (const file of jsonFiles) {
```
BYK: This entire loop should be replaced with a `Promise.all()` full-parallel block with concurrency limiting using `p-limit`

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+
+  return entries;
+}
+
+/** Delete cache files that have expired */
+async function deleteExpiredEntries(
+  cacheDir: string,
+  entries: EntryMetadata[]
+): Promise<void> {
+  for (const entry of entries) {
```
BYK: Similar to above, we should parallelize this entire thing but limit the concurrency with `p-limit`

src/lib/response-cache.ts
```diff
@@ -0,0 +1,588 @@
...
+  entries: EntryMetadata[]
+): Promise<void> {
+  const remaining = entries.filter((e) => !e.expired);
+  if (remaining.length <= MAX_CACHE_ENTRIES) {
+    return;
+  }
+
+  remaining.sort((a, b) => a.createdAt - b.createdAt);
+  const toEvict = remaining.slice(0, remaining.length - MAX_CACHE_ENTRIES);
+  for (const entry of toEvict) {
```
BYK: Same as above, parallel with concurrency limit.

BYK:
**Re: response-cache.ts — assorted cleanup** (threads #7–#11)

All addressed in 2cee1488:

- **#7 localeCompare**: Replaced with simple `<`/`>` comparison — URL query params are always ASCII
- **#8 malformed URL**: Removed try-catch in `normalizeUrl()` — URLs reaching the cache layer already came from a successful fetch
- **#9 immutableMinTimeToLive**: Now derives from `FALLBACK_TTL_MS.immutable` instead of hardcoded `3_600_000`
- **#10 headersToObject**: Simplified to `Object.fromEntries(headers.entries())`
- **#11 await in catch**: Removed `await` from fire-and-forget `unlink()` — we `return` immediately after, no need to wait

src/commands/issue/list.ts
```diff
@@ -1184,7 +1185,7 @@ export const listCommand = buildListCommand("issue", {
     },
     aliases: {
       ...LIST_BASE_ALIASES,
-      f: "fresh",
+      ...FRESH_ALIASES,
```
BYK, line 1204: Does this not override `f: "follow"`?

src/lib/response-cache.ts
```diff
@@ -150,13 +159,19 @@ type CacheEntry = {
...
+   * Allows cleanup to check freshness without deserializing CachePolicy.
+   * Optional for backwards compatibility with entries written before this field.
+   */
+  expiresAt?: number;
 };
 
 /** CachePolicy options for a single-user CLI cache */
 const POLICY_OPTIONS: CachePolicy.Options = {
   shared: false,
   cacheHeuristic: 0.1,
```
BYK, line 174: Is this the same as `CLEANUP_PROBABILITY`? If so they should share the same constant

src/lib/response-cache.ts
```diff
@@ -473,6 +491,25 @@ export async function clearResponseCache(): Promise<void> {
...
+/**
+ * Run an async function over items with bounded concurrency.
+ * Uses p-limit to prevent overwhelming the filesystem with simultaneous reads.
+ */
+async function parallel<T>(
+  items: readonly T[],
+  fn: (item: T) => Promise<void>
+): Promise<void> {
+  const limit = pLimit(CACHE_IO_CONCURRENCY);
+  await Promise.all(items.map((item) => limit(() => fn(item))));
```
BYK: Why not simpler:

```ts
 const limetedFn = limit(fn);
 await Promise.all(items.map(limitedFn);
```

src/lib/response-cache.ts
```diff
@@ -580,9 +610,9 @@ async function evictExcessEntries(
 
   remaining.sort((a, b) => a.createdAt - b.createdAt);
   const toEvict = remaining.slice(0, remaining.length - MAX_CACHE_ENTRIES);
-  for (const entry of toEvict) {
+  await parallel(toEvict, async (entry) => {
     await unlink(join(cacheDir, entry.file)).catch(() => {
       // Best-effort eviction
     });
-  }
+  });
```
BYK: If this is also best effort, should we still `await`?

src/lib/response-cache.ts
BYK, line 1: I think we should instrument all this cache-related operations with Sentry to see the impact of caching (good or bad)

src/lib/response-cache.ts
```diff
@@ -0,0 +1,620 @@
...
+// ---------------------------------------------------------------------------
+
+let cacheDisabledFlag = false;
+
+/**
+ * Disable the response cache for the current process.
+ * Called when `--refresh` flag is passed to a command.
+ */
+export function disableResponseCache(): void {
+  cacheDisabledFlag = true;
```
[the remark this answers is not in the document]
BYK: Fixed in 87587314. Added `resetCacheState()` export to `response-cache.ts` and wired it into both `beforeEach` and `afterEach` in the test file so the `cacheDisabledFlag` resets between test runs.

src/lib/db/auth.ts
```diff
@@ -103,6 +104,14 @@ export function clearAuth(): void {
...
     db.query("DELETE FROM pagination_cursors").run();
   });
+
+  // Clear cached API responses — they are tied to the current user's permissions.
+  // Awaited so cache is fully removed before the process exits.
+  try {
+    await clearResponseCache();
+  } catch {
+    // Non-fatal: cache directory may not exist yet
+  }
```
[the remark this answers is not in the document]
BYK: The lore entry references a now-resolved issue. The original problem was that `fcAssert(asyncProperty(...))` calls were not being `await`ed in model-based tests, which was fixed in 8dcc111 (all 6 async test functions now properly `await`). The `async clearAuth()` pattern was explicitly requested by BugBot in review round 3 (thread #8) and all production callers already `await` it. The stale lore entry exists on `origin/main` and is not changed by this PR.

src/lib/response-cache.ts
```diff
@@ -0,0 +1,620 @@
...
+}
```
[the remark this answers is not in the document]
BYK: Duplicate of thread above — same fix in 87587314. Added `resetCacheState()` export and wired into test setup/teardown.

src/lib/response-cache.ts
```diff
@@ -0,0 +1,633 @@
...
+    // Server provided an explicit TTL — respect it (even if expired)
+    return false;
+  }
+
+  // No server TTL (0) — use our URL-based fallback tier
+  const tier = classifyUrl(url);
+  const fallbackTtl = FALLBACK_TTL_MS[tier];
+  const age = Date.now() - entry.createdAt;
+  return age <= fallbackTtl;
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 56f46c1a. Added `hasServerCacheDirectives(policy)` helper that checks `Object.keys(policy.toObject().rescc).length > 0` — `rescc` is part of the typed `CachePolicyObject` interface and is empty when no Cache-Control header was sent by the server.

Updated both locations:
- `isEntryFresh`: checks `hasServerCacheDirectives` instead of `serverTtl !== 0` — if the server sent explicit directives and CachePolicy says stale, return false immediately.
- `writeResponseToCache` expiry computation: `hasServerCacheDirectives ? serverTtl : fallbackTtl` — so `max-age=0` gets `expiresAt = now + 0` (expired on creation).

src/lib/response-cache.ts
```diff
@@ -0,0 +1,685 @@
...
+    return;
+  }
+
+  const entries = await collectEntryMetadata(cacheDir, jsonFiles);
+
+  // Both operations are best-effort — run them in parallel without blocking
+  await Promise.all([
+    deleteExpiredEntries(cacheDir, entries),
+    evictExcessEntries(cacheDir, entries),
+  ]);
```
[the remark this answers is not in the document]
BYK: This is a false positive — there's no race condition here.

Both `deleteExpiredEntries` and `evictExcessEntries` operate on the **same in-memory snapshot** of `EntryMetadata[]` from `collectEntryMetadata()`. Neither modifies the shared array. The concurrent `unlink()` calls are on different files:

- `deleteExpiredEntries` operates on `entries.filter(e => e.expired)` — expired files
- `evictExcessEntries` operates on `entries.filter(e => !e.expired)` — non-expired files, sorted by age

The count `remaining.length` in `evictExcessEntries` already **excludes** expired entries (line 673: `const remaining = entries.filter((e) => !e.expired)`). So the expired entries being deleted concurrently don't affect the eviction count at all.

The only theoretical overlap would be if an entry transitions from non-expired to expired between `collectEntryMetadata` and `evictExcessEntries`, but that's inherent to any point-in-time snapshot and the `.catch()` on `unlink` handles double-deletion gracefully.

src/lib/response-cache.ts
```diff
@@ -0,0 +1,685 @@
...
+ *
+ * @param method - HTTP method (e.g., "GET")
+ * @param url - Full URL string
+ * @returns Hex-encoded SHA-256 hash suitable for use as a filename
+ * @internal Exported for testing
+ */
+export function buildCacheKey(method: string, url: string): string {
+  const normalized = normalizeUrl(method, url);
+  return createHash("sha256").update(normalized).digest("hex");
+}
```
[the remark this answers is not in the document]
BYK: The cache key intentionally uses only method+URL. Authorization is handled at the **validation** layer, not the key layer:

1. **`http-cache-semantics`** stores the full request headers (including `Authorization`) in the `CachePolicy` object. At read time, `policy.satisfiesWithoutRevalidation(newRequest)` compares headers per `Vary`.

2. **Cache is wiped on auth changes**: `clearResponseCache()` is called on both `login` (line 56 and 105 in login.ts) and `logout` (via `clearAuth()` in auth.ts). So stale tokens never persist across sessions.

3. **Single-user CLI**: `~/.sentry/` is per-user. Multiple users on the same machine have separate config dirs.

The PR description was inaccurate — I've updated it. The key is method+URL only; auth validation happens via CachePolicy header comparison at read time.

PR #334, opened 2026-03-04
feat(install): add Sentry error telemetry to install script

BYK:
## Summary

Add fire-and-forget error reporting to the install script via Sentry's
envelope API. Uses the CLI's existing public write-only DSN — no new
secrets or projects needed. This would have caught the macOS digest
extraction bug (#331) automatically.

## Changes

- `report_error()` — builds a minimal Sentry envelope and sends it via
  background `curl` (2s timeout, never blocks installation)
- `die()` — replaces all 11 `echo + exit 1` patterns with error
  reporting + exit
- `ERR` trap — catches unexpected `set -e` / `pipefail` exits like the
  gunzip pipeline failure that triggered #331
- UUID generation with fallback chain: `/proc` → `uuidgen` → `awk`
- Opt-out via `SENTRY_CLI_NO_TELEMETRY=1` (same as the CLI binary)

Each error event is tagged with `os`, `arch`, `channel`, `step`,
`install.version`, and bash version for quick triage.

### DIFF
--- a/install
+++ b/install
@@ -39,15 +39,19 @@ report_error() {
     local timestamp
     timestamp=$(date -u +"%Y-%m-%dT%H:%M:%SZ" 2>/dev/null || echo "")
 
-    # Escape message for JSON (backslashes, quotes, newlines)
-    local json_msg
-    json_msg=$(printf '%s' "$msg" | sed 's/\\/\\\\/g;s/"/\\"/g' | tr '\n' ' ')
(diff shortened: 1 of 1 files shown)


BYK:
Fixed in 9c36ef79 — extracted a `_esc()` helper and applied it to `step`, `channel` (from `requested_version`), and `version` tags. The `os` and `arch` values are hardcoded constants so they don't need escaping.

PR #338, opened 2026-03-04
feat(logger): add consola-based structured logging with Sentry integration

BYK:
## Summary

Add a structured logging module built on [consola](https://github.com/unjs/consola) v3.x that provides:

1. **FancyReporter** (built-in) — stderr output with colors/icons for TTY, BasicReporter fallback for CI
2. **Sentry.createConsolaReporter()** — auto-forwards all log messages to Sentry structured logs for telemetry

## Key Features

- **`SENTRY_LOG_LEVEL` env var** for controlling verbosity: `error`, `warn`, `info` (default), `debug`, `trace`
- **Global `--verbose` / `--log-level` flags** pre-parsed from argv before Stricli (same pattern as `shouldSuppressNotification`)
- **Tag-based scoping** via `logger.withTag('scope')` for domain-specific logging
- **Idempotent Sentry reporter attachment** after `Sentry.init()`

## Initial Instrumentation

- **delta-upgrade**: chain resolution, patch application, fallback decisions
- **upgrade**: delta vs full download path selection  
- **dsn/code-scanner**: large file skipping, unreadable file warnings (replaces TODO comments)

## Architecture

```
┌─────────────────────────────────────────────────────────┐
│ consola instance (src/lib/logger.ts)                     │
│                                                          │
│  Reporters:                                              │
│  ├─ FancyReporter (built-in → stderr)                   │
│  └─ Sentry.createConsolaReporter() (auto-forwards       │
│     to Sentry structured logs)                           │
│                                                          │
│  Level: SENTRY_LOG_LEVEL env var or --log-level flag     │
│  Tags: logger.withTag('upgrade'), .withTag('dsn-scan')   │
└─────────────────────────────────────────────────────────┘
```

## Tests

- 27 unit tests covering `parseLogLevel`, `extractLogLevelFromArgs`, `setLogLevel`, `attachSentryReporter`, and the logger instance
- 11 property-based tests (fast-check) verifying invariants: level ordering, case insensitivity, flag extraction correctness
- 884 assertions total

## Files Changed

| File | Change |
|------|--------|
| `src/lib/logger.ts` | **New** — Core logger module (234 lines) |
| `src/bin.ts` | Extract `--verbose`/`--log-level` from argv before Stricli |
| `src/lib/telemetry.ts` | Attach Sentry reporter after `Sentry.init()` |
| `src/lib/delta-upgrade.ts` | Debug logs for chain resolution + patching |
| `src/lib/upgrade.ts` | Info/debug logs for delta vs full download |
| `src/lib/dsn/code-scanner.ts` | Replace TODO comments with debug logs |
| `AGENTS.md` | Add consola architecture lore entry |
| `test/lib/logger.test.ts` | **New** — Unit tests |
| `test/lib/logger.property.test.ts` | **New** — Property-based tests |

### DIFF
--- a/test/lib/formatters/markdown.test.ts
+++ b/test/lib/formatters/markdown.test.ts
@@ -232,10 +232,14 @@ describe("isPlainOutput", () => {
     });
   });
 
-  describe("isTTY fallback (both env vars unset)", () => {
+  describe("isTTY fallback (all env vars unset)", () => {
     test("non-TTY → plain", () => {
       withEnv(
-        { SENTRY_PLAIN_OUTPUT: undefined, NO_COLOR: undefined },
+        {
+          SENTRY_PLAIN_OUTPUT: undefined,
+          NO_COLOR: undefined,
+          FORCE_COLOR: undefined,
+        },
         false,
         () => {
           expect(isPlainOutput()).toBe(true);
@@ -245,7 +249,11 @@ describe("isPlainOutput", () => {
 
     test("TTY → rendered", () => {
       withEnv(
-        { SENTRY_PLAIN_OUTPUT: undefined, NO_COLOR: undefined },
+        {
+          SENTRY_PLAIN_OUTPUT: undefined,
+          NO_COLOR: undefined,
+          FORCE_COLOR: undefined,
+        },
         true,
         () => {
           expect(isPlainOutput()).toBe(false);
@@ -255,7 +263,11 @@ describe("isPlainOutput", () => {
 
     test("isTTY=undefined → plain", () => {
       withEnv(
-        { SENTRY_PLAIN_OUTPUT: undefined, NO_COLOR: undefined },
+        {
+          SENTRY_PLAIN_OUTPUT: undefined,
+          NO_COLOR: undefined,
+          FORCE_COLOR: undefined,
+        },
         undefined,
         () => {
           expect(isPlainOutput()).toBe(true);


src/lib/upgrade.ts
```diff
@@ -581,7 +585,10 @@ export async function downloadBinaryToTemp(
     // Try delta upgrade first — downloads tiny patches instead of full binary.
     // Falls back to full download on any failure (missing patches, hash mismatch, etc.)
     const deltaResult = await tryDeltaUpgrade(version, tempPath);
-    if (!deltaResult) {
+    if (deltaResult) {
+      log.info("Applied delta patch — skipped full download");
```
BYK: Show us delta size at least (and the full download size if it is freely available -- from the GitHub API where we get the SHA256 for instance)

src/lib/logger.ts
```diff
@@ -0,0 +1,234 @@
...
+ * // Scoped diagnostic messages
+ * const log = logger.withTag("delta-upgrade");
+ * log.debug("Resolved stable chain: 0.12.0 → 0.13.0");
+ * ```
+ *
+ * ## Environment variables
+ *
+ * - `SENTRY_LOG_LEVEL` — Sets log level: `error`, `warn`, `info` (default), `debug`, `trace`
+ * - `NO_COLOR` — Disables color output (respected by consola natively)
+ * - `FORCE_COLOR` — Forces color output even in non-TTY
```
BYK, line 45: We should support `FORCE_COLOR` in other places too if not already

src/lib/logger.ts
```diff
@@ -0,0 +1,234 @@
...
+] as const;
+
+/** A valid log level name string */
+export type LogLevelName = (typeof LOG_LEVEL_NAMES)[number];
+
+/** Maps level names to consola numeric levels */
+const LOG_LEVEL_MAP: Record<LogLevelName, number> = {
+  error: 0,
+  warn: 1,
+  info: 3,
```
BYK: What is level 2? If we can add that to our `LOG_LEVEL_NAMES` array above, we can just use their index for the numeric values.

src/lib/logger.ts
BYK, line 1: We should probably do a sweep on where we use `stderr` or `console.error()` directly and replace those with `logger` calls?

src/lib/upgrade.ts
BYK: Done in 540a3bd6. Added `DeltaResult` type with `patchBytes` field. `attemptDeltaUpgrade` now returns `{ sha256, patchBytes }` and the log shows: `Applied delta patch (XX.X KB downloaded)`. Full download size isn't readily available (the `.gz` layer size is only fetched inside the nightly path for threshold comparison, not for stable releases), so just showing the delta size for now.

src/lib/logger.ts
BYK: Done in 540a3bd6. Added `FORCE_COLOR` support to `isPlainOutput()` in `markdown.ts` — checked between `NO_COLOR` and `isTTY` so that `NO_COLOR` always wins if both are set. Same convention as chalk/consola: any non-empty value forces rich output.

src/lib/logger.ts
BYK: Good catch. Level 2 is `"log"` in consola (which includes `console.log()`-style output). Added it to `LOG_LEVEL_NAMES` as `"log"` in 540a3bd6. The mapping is now: `["error", "warn", "log", "info", "debug", "trace"]` matching consola's internal `LogLevels` enum (0=fatal/error, 1=warn, 2=log, 3=info, 4=debug, 5=trace).

src/lib/logger.ts
BYK: Done in 540a3bd6. Converted 7 direct stderr/console.error calls to logger in 3 files:

- **migration.ts**: 4 calls → `log.warn()`, `log.info()`, `log.success()`, `log.error()` (via `logger.withTag("migration")`)
- **schema.ts**: 1 call → `logger.info()` for auto-repair notification
- **api-client.ts**: 2 calls → `logger.warn()` for pagination limit and team creation warnings

Left untouched (follow-up):
- **bin.ts** (lines 60, 71, 111, 123): Top-level error handler and auto-auth retry — these run before/outside the command lifecycle and format errors directly. Converting these requires care since logger output format may differ from expected error formatting.
- **telemetry.ts** (lines 640, 715, 736, 756): Telemetry shutdown diagnostics — these fire during process exit when the logger may already be torn down.

src/lib/upgrade.ts
BYK: Good call. Switched from a flat constant to `DELTA_SIZE_FRACTION = 0.6` (60%) derived from `DELTA_SIZE_THRESHOLD` logic. The check is now `patchSize >= fullSize * DELTA_SIZE_FRACTION` — if the patch is ≥60% of full, skip it. Constant lives in `constants.ts` alongside existing thresholds.

src/lib/formatters/markdown.ts
```diff
@@ -69,6 +71,14 @@ export function isPlainOutput(): boolean {
     return noColor !== "";
   }
 
+  // FORCE_COLOR: any non-empty value forces rich output (same convention
+  // as chalk, supports-color, consola). Checked after NO_COLOR so that
+  // NO_COLOR always wins if both are set.
+  const forceColor = process.env.FORCE_COLOR;
+  if (forceColor !== undefined && forceColor !== "") {
+    return false;
+  }
```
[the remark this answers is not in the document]
BYK: Good catch. Fixed in dda4f8a0 — `FORCE_COLOR=0` now returns `true` (plain output) per the chalk/supports-color convention. Any other non-empty value still forces color on.

src/lib/logger.ts
```diff
@@ -0,0 +1,232 @@
...
+      createConsolaReporter: (options?: Record<string, unknown>) => {
+        log: (logObj: unknown) => void;
+      };
+    };
+
+    const sentryReporter = Sentry.createConsolaReporter();
+    logger.addReporter(sentryReporter);
+  } catch {
+    // Sentry not available (e.g., telemetry disabled) — continue without reporter
+  }
```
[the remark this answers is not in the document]
BYK: Fixed in dda4f8a0 — moved `sentryReporterAttached = true` inside the `try` block, after `logger.addReporter()` succeeds. If the require or createConsolaReporter throws, the flag stays false so a future call can retry.

test/lib/formatters/markdown.test.ts
```diff
@@ -183,6 +188,50 @@ describe("isPlainOutput", () => {
...
+          FORCE_COLOR: "1",
+        },
+        true,
+        () => {
+          expect(isPlainOutput()).toBe(true);
+        }
+      );
+    });
+  });
+
```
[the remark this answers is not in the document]
BYK: Fixed in 2983787c — all three isTTY fallback tests now pass `FORCE_COLOR: undefined` to `withEnv`.

PR #356, opened 2026-03-05
feat: support SENTRY_AUTH_TOKEN and SENTRY_TOKEN env vars for headless auth

### LINKED ISSUE #348: Support SENTRY_AUTH_TOKEN env var for headless/agent authentication
## Summary

Support `SENTRY_AUTH_TOKEN` environment variable for authentication, bypassing the OAuth device flow and SQLite token storage. This enables headless environments (CI, Docker, agent sandboxes) to authenticate without interactive login.

## Motivation

From [You Need to Rewrite Your CLI for AI Agents](https://justin.poehnelt.com/posts/rewrite-your-cli-for-ai-agents/) (Justin Poehnelt, Google DevRel):

> "Agents can do OAuth but not easily and probably shouldn't. `GOOGLE_WORKSPACE_CLI_TOKEN` and `GOOGLE_WORKSPACE_CLI_CREDENTIALS_FILE` enable credential injection via environment — the only auth path that works when nobody is sitting at a browser."

This is especially relevant for:
- **CI/CD pipelines** — no browser available for OAuth
- **AI agents** — sandboxed environments where persisting tokens to disk is undesirable
- **Docker containers** — ephemeral environments where SQLite state is lost between runs
- **Quick API scripting** — `SENTRY_AUTH_TOKEN=sntrys_... sentry issue list` without `sentry auth login` first
- **Backwards compatibility** — the legacy `sentry-cli` used `SENTRY_AUTH_TOKEN`

## Current behavior

The only way to authenticate is:
1. `sentry auth login` (OAuth device flow — requires browser)
2. `sentry auth login --token <token>` (stores to SQLite DB — requires prior login step)

No env var override exists. Agents and CI must run `sentry auth login --token` before every ephemeral session.

## Proposed behavior

```bash
# Env var takes precedence over stored OAuth token
SENTRY_AUTH_TOKEN=sntrys_... sentry issue list

# Works with all commands
SENTRY_AUTH_TOKEN=sntrys_... sentry api /organizations/

# auth status shows env var source
SENTRY_AUTH_TOKEN=sntrys_... sentry auth status
# → Authenticated via SENTRY_AUTH_TOKEN environment variable
```

## Implementation

The change is minimal — modify `getAuthToken()` in `src/lib/db/auth.ts` (and/or `refreshToken()`) to check for the env var first:

```typescript
export function getAuthToken(): string | undefined {
  // Env var takes precedence (headless/CI/agent mode)
  const envToken = process.env.SENTRY_AUTH_TOKEN;
  if (envToken) {
    return envToken;
  }

  // Fall back to stored OAuth token
  return withDbSpan("getAuthToken", () => {
    const db = getDatabase();
    const row = db.query("SELECT * FROM auth WHERE id = 1").get() as AuthRow | undefined;
    // ...existing logic
  });
}
```

Key decisions:
- **Env var wins over stored token** — consistent with `SENTRY_URL`, `SENTRY_ORG`, `SENTRY_PROJECT` which all override stored defaults
- **No refresh logic for env tokens** — env var tokens are assumed to be valid; don't attempt OAuth refresh
- **`sentry auth status` should indicate source** — show "Authenticated via SENTRY_AUTH_TOKEN" vs "Authenticated via OAuth"
- **Variable name: `SENTRY_AUTH_TOKEN`** — matches the legacy `sentry-cli` convention and what the Sentry SDK uses

## Existing env var pattern

The CLI already respects env vars for other config:
- `SENTRY_URL` — API base URL (self-hosted)
- `SENTRY_ORG` — default organization
- `SENTRY_PROJECT` — default project
- `SENTRY_DSN` — project DSN
- `SENTRY_CLIENT_ID` — OAuth client ID

`SENTRY_AUTH_TOKEN` is the missing piece.

## Security considerations

- Env var tokens should never be logged or included in error reports
- `sentry auth status --json` should indicate the auth source but not echo the full token
- The test preload (`test/preload.ts`) already deletes `SENTRY_AUTH_TOKEN` for test isolation

## References

- [Multi-Surface: MCP, Extensions, Env Vars](https://justin.poehnelt.com/posts/rewrite-your-cli-for-ai-agents/#multi-surface-mcp-extensions-env-vars) — Justin Poehnelt
- Legacy sentry-cli used `SENTRY_AUTH_TOKEN`: [docs](https://docs.sentry.io/cli/configuration/#environment-variables)
- `test/preload.ts` already cleans `SENTRY_AUTH_TOKEN` in test setup

BYK:
## Summary

Support `SENTRY_AUTH_TOKEN` and `SENTRY_TOKEN` environment variables for authentication, bypassing the OAuth device flow and SQLite token storage. This enables headless environments (CI, Docker, agent sandboxes) to authenticate without interactive login.

Closes #348

### Core auth layer (`src/lib/db/auth.ts`)
- New `AuthSource` type: `"env:SENTRY_AUTH_TOKEN" | "env:SENTRY_TOKEN" | "oauth"`
- `getAuthToken()` and `getAuthConfig()` check env vars first, then fall back to SQLite
- `refreshToken()` short-circuits for env tokens (no refresh/expiry logic)
- New `isEnvTokenActive()` helper for commands that need to branch on auth source

### HTTP layer (`src/lib/sentry-client.ts`)
- `handleUnauthorized()` skips token refresh when env token is active (401s propagate as-is)

### Commands
- **`auth status`** — Shows "Authenticated via SENTRY_AUTH_TOKEN environment variable ✓"; skips config path, expiry, and auto-refresh display for env tokens
- **`auth refresh`** — Errors early with clear message when env token is active
- **`auth login`** — Tells user to remove env var instead of re-authenticating
- **`auth logout`** — Clears stored auth but warns env var still provides authentication

### Tests
- 22 unit tests covering env var priority, source tracking, trimming, refresh bypass
- 9 property-based tests (fast-check) verifying priority invariants for any valid token values
- Model-based test extensions with 5 new env var command classes
- Test preload cleans `SENTRY_TOKEN` alongside existing `SENTRY_AUTH_TOKEN`

## Design decisions
- `SENTRY_AUTH_TOKEN` takes priority over `SENTRY_TOKEN` (matches legacy sentry-cli and Sentry SDK convention)
- Empty/whitespace-only values are treated as unset
- `AuthConfig.source` is required on every config object (not optional) so consumers can't forget to handle it

### DIFF
--- a/src/commands/auth/login.ts
+++ b/src/commands/auth/login.ts
@@ -3,8 +3,7 @@ import { getCurrentUser, getUserRegions } from "../../lib/api-client.js";
 import { buildCommand, numberParser } from "../../lib/command.js";
 import {
   clearAuth,
-  ENV_SOURCE_PREFIX,
-  getAuthConfig,
+  getActiveEnvVarName,
   isAuthenticated,
   isEnvTokenActive,
   setAuthToken,
@@ -52,10 +51,7 @@ export const loginCommand = buildCommand({
     // Check if already authenticated
     if (await isAuthenticated()) {
       if (isEnvTokenActive()) {
-        const config = getAuthConfig();
-        const envVar = config?.source.startsWith(ENV_SOURCE_PREFIX)
-          ? config.source.slice(ENV_SOURCE_PREFIX.length)
-          : "SENTRY_AUTH_TOKEN";
+        const envVar = getActiveEnvVarName();
         stdout.write(
           `Authentication is provided via ${envVar} environment variable. ` +
             `Unset ${envVar} to use OAuth-based login instead.\n`
--- a/src/commands/auth/logout.ts
+++ b/src/commands/auth/logout.ts
@@ -8,8 +8,7 @@ import type { SentryContext } from "../../context.js";
 import { buildCommand } from "../../lib/command.js";
 import {
   clearAuth,
-  ENV_SOURCE_PREFIX,
-  getAuthConfig,
+  getActiveEnvVarName,
   isAuthenticated,
   isEnvTokenActive,
 } from "../../lib/db/auth.js";
@@ -34,10 +33,7 @@ export const logoutCommand = buildCommand({
     }
 
     if (isEnvTokenActive()) {
-      const config = getAuthConfig();
-      const envVar = config?.source.startsWith(ENV_SOURCE_PREFIX)
-        ? config.source.slice(ENV_SOURCE_PREFIX.length)
-        : "SENTRY_AUTH_TOKEN";
+      const envVar = getActiveEnvVarName();
       stdout.write(
         `Authentication is provided via ${envVar} environment variable.\n` +
           `Unset ${envVar} to log out.\n`
--- a/src/lib/db/auth.ts
+++ b/src/lib/db/auth.ts
@@ -59,6 +59,22 @@ export function isEnvTokenActive(): boolean {
   return getEnvToken() !== undefined;
 }
 
+/**
+ * Get the name of the active env var providing authentication.
+ * Returns the specific variable name (e.g. "SENTRY_AUTH_TOKEN" or "SENTRY_TOKEN").
+ *
+ * **Important**: Call only after checking {@link isEnvTokenActive} returns true.
+ * Falls back to "SENTRY_AUTH_TOKEN" if no env source is active, which is a safe
+ * default for error messages but may be misleading if used unconditionally.
+ */
+export function getActiveEnvVarName(): string {
+  const env = getEnvToken();
+  if (env) {
+    return env.source.slice(ENV_SOURCE_PREFIX.length);
+  }
+  return "SENTRY_AUTH_TOKEN";
+}
+
 export function getAuthConfig(): AuthConfig | undefined {
   const envToken = getEnvToken();
   if (envToken) {
--- a/src/commands/auth/refresh.ts
+++ b/src/commands/auth/refresh.ts
@@ -7,7 +7,7 @@
 import type { SentryContext } from "../../context.js";
 import { buildCommand } from "../../lib/command.js";
 import {
-  ENV_SOURCE_PREFIX,
+  getActiveEnvVarName,
   getAuthConfig,
   isEnvTokenActive,
   refreshToken,
@@ -69,10 +69,7 @@ Examples:
 
     // Env var tokens can't be refreshed
     if (isEnvTokenActive()) {
-      const config = getAuthConfig();
-      const envVar = config?.source.startsWith(ENV_SOURCE_PREFIX)
-        ? config.source.slice(ENV_SOURCE_PREFIX.length)
-        : "SENTRY_AUTH_TOKEN";
+      const envVar = getActiveEnvVarName();
       throw new AuthError(
         "invalid",
         "Cannot refresh an environment variable token.\n" +
--- a/test/commands/auth/login.test.ts
+++ b/test/commands/auth/login.test.ts
@@ -130,18 +130,17 @@ describe("loginCommand.func --token path", () => {
   test("already authenticated (env token SENTRY_TOKEN): shows specific var name", async () => {
     isAuthenticatedSpy.mockResolvedValue(true);
     isEnvTokenActiveSpy.mockReturnValue(true);
-    const getAuthConfigSpy = spyOn(dbAuth, "getAuthConfig");
-    getAuthConfigSpy.mockReturnValue({
-      token: "sntrys_token_456",
-      source: "env:SENTRY_TOKEN",
-    });
+    // Set env var directly — getActiveEnvVarName() reads env vars via getEnvToken()
+    process.env.SENTRY_TOKEN = "sntrys_token_456";
 
     const { context, getStdout } = createContext();
-    await func.call(context, { timeout: 900 });
-
-    expect(getStdout()).toContain("SENTRY_TOKEN");
-    expect(getStdout()).not.toContain("SENTRY_AUTH_TOKEN");
-    getAuthConfigSpy.mockRestore();
+    try {
+      await func.call(context, { timeout: 900 });
+      expect(getStdout()).toContain("SENTRY_TOKEN");
+      expect(getStdout()).not.toContain("SENTRY_AUTH_TOKEN");
+    } finally {
+      delete process.env.SENTRY_TOKEN;
+    }
   });
 
   test("--token: stores token, fetches user, writes success", async () => {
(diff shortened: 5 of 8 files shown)


src/commands/auth/logout.ts
```diff
@@ -27,6 +32,21 @@ export const logoutCommand = buildCommand({
       return;
     }
 
+    if (isEnvTokenActive()) {
+      const config = getAuthConfig();
+      const envVar = config?.source.startsWith("env:")
+        ? config.source.slice(4)
+        : "SENTRY_AUTH_TOKEN";
+      // Still clear stored auth so if env var is removed later, user is cleanly logged out
```
BYK: I think this is the wrong approach. When something runs `sentry auth logout` while providing a token through env variable, we should mention this and they cannot login with that. They can unset the env variable or use `-f/--force` to force removing local credentials maybe but I'm not a fan of the latter.

src/commands/auth/status.ts
```diff
@@ -42,6 +43,16 @@ function writeUserInfo(stdout: Writer): void {
...
 }
 
+/** Check if the auth source is an environment variable */
+function isEnvSource(source: AuthSource): boolean {
+  return source.startsWith("env:");
+}
+
+/** Extract the env var name from an env-based AuthSource (e.g. "env:SENTRY_AUTH_TOKEN" → "SENTRY_AUTH_TOKEN") */
+function envVarName(source: AuthSource): string {
+  return source.slice(4);
```
BYK: Instead of hard-coding `4` here, abstract the `"env:"` part into a constant and use its `.length` property here.

src/lib/db/auth.ts
```diff
@@ -43,12 +77,18 @@ export function getAuthConfig(): AuthConfig | undefined {
...
       expiresAt: row.expires_at ?? undefined,
       issuedAt: row.issued_at ?? undefined,
+      source: "oauth" as const,
     };
   });
 }
 
-/** Get the stored token, or undefined if expired. Use refreshToken() for auto-refresh. */
+/** Get the active auth token. Checks env vars first, then falls back to SQLite. */
 export function getAuthToken(): string | undefined {
```
BYK, line 105: We may want to move this function out of `db/auth.ts` as it now does unrelated things to the DB itself?

src/commands/auth/logout.ts
BYK: Fixed: logout no longer clears stored auth when env var is active. It now just informs the user that auth is via env var and they should unset it to log out.

src/commands/auth/status.ts
BYK: Done: extracted `ENV_SOURCE_PREFIX = "env:"` constant in `db/auth.ts` and use `.length` instead of hardcoded `4` in both `status.ts` and `logout.ts`.

src/lib/db/auth.ts
BYK: Considered this but decided to keep them in `db/auth.ts` for now — `getEnvToken()` is a private helper tightly coupled with `getAuthToken()`, `getAuthConfig()`, and `refreshToken()` which all live in this file and call it. Moving it out would either create a circular dependency or require extracting all of auth into a new module. The DB-specific functions (SQLite reads/writes) are still clearly scoped within `withDbSpan` blocks, while the env var check is a lightweight prefix that short-circuits before any DB access. Happy to revisit if the file grows further.

src/commands/auth/login.ts
```diff
@@ -44,9 +49,16 @@ export const loginCommand = buildCommand({
...
     // Check if already authenticated
     if (await isAuthenticated()) {
-      stdout.write(
-        "You are already authenticated. Use 'sentry auth logout' first to re-authenticate.\n"
-      );
+      if (isEnvTokenActive()) {
+        stdout.write(
+          "Authentication is provided via environment variable. " +
+            "Remove the env var to use OAuth-based login instead.\n"
+        );
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in aa85330 — both `login` and `refresh` now use `getAuthConfig().source` with `ENV_SOURCE_PREFIX` to dynamically show the specific active env var name, matching the pattern used by `logout` and `status`.

src/commands/auth/login.ts
```diff
@@ -44,9 +51,20 @@ export const loginCommand = buildCommand({
...
     // Check if already authenticated
     if (await isAuthenticated()) {
-      stdout.write(
-        "You are already authenticated. Use 'sentry auth logout' first to re-authenticate.\n"
-      );
+      if (isEnvTokenActive()) {
+        const config = getAuthConfig();
+        const envVar = config?.source.startsWith(ENV_SOURCE_PREFIX)
+          ? config.source.slice(ENV_SOURCE_PREFIX.length)
+          : "SENTRY_AUTH_TOKEN";
```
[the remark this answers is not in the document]
BYK: Extracted `getActiveEnvVarName()` helper in `db/auth.ts`. All three commands now use the shared helper instead of duplicating the pattern.

src/lib/db/auth.ts
```diff
@@ -20,14 +20,64 @@ type AuthRow = {
...
+ * Returns the specific variable name (e.g. "SENTRY_AUTH_TOKEN") or
+ * falls back to "SENTRY_AUTH_TOKEN" if no env token is active.
+ */
+export function getActiveEnvVarName(): string {
+  const config = getAuthConfig();
+  if (config?.source.startsWith(ENV_SOURCE_PREFIX)) {
+    return config.source.slice(ENV_SOURCE_PREFIX.length);
+  }
+  return "SENTRY_AUTH_TOKEN";
+}
```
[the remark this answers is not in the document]
BYK: Valid concern. Added JSDoc noting that callers should check `isEnvTokenActive()` first. The fallback to `"SENTRY_AUTH_TOKEN"` is intentionally safe — all three call sites already guard with `if (isEnvTokenActive())`, and the JSDoc now documents this contract.

src/lib/db/auth.ts
```diff
@@ -20,14 +20,67 @@ type AuthRow = {
...
+ * Falls back to "SENTRY_AUTH_TOKEN" if no env source is active, which is a safe
+ * default for error messages but may be misleading if used unconditionally.
+ */
+export function getActiveEnvVarName(): string {
+  const config = getAuthConfig();
+  if (config?.source.startsWith(ENV_SOURCE_PREFIX)) {
+    return config.source.slice(ENV_SOURCE_PREFIX.length);
+  }
+  return "SENTRY_AUTH_TOKEN";
+}
```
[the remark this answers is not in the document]
BYK: Good catch — now calls `getEnvToken()` directly instead of `getAuthConfig()`, eliminating the unnecessary SQLite fallback path.

PR #362, opened 2026-03-06
fix(log-view): support multiple log IDs and validate hex format

BYK:
Fixes CLI-BC

## Problem

An AI agent called `sentry log view brandai/brandai "<7 log IDs joined by newlines>"`. The newline-delimited blob was treated as a single log ID, interpolated into the API query `sentry.item_id:${logId}`, and the Sentry API rejected it with 400 Bad Request.

## Solution

Instead of just validating and rejecting, **accept multiple log IDs**. The Sentry Explore API already supports bracket syntax (`sentry.item_id:[id1,id2,...]`), and the Stricli positional is already `kind: "array"`.

### Changes

- **`src/lib/hex-id.ts`** (new) — Shared 32-char hex ID validator with whitespace trimming and truncated error messages
- **`src/lib/trace-id.ts`** — Delegates to `hex-id.ts` (backward-compatible re-export)
- **`src/lib/api-client.ts`** — New `getLogs()` with bracket query syntax; `getLog()` becomes a convenience wrapper
- **`src/commands/log/view.ts`** — `parsePositionalArgs()` returns `logIds[]`; splits newline-in-arg; validates each ID; displays multiple logs with `---` separators; warns about missing IDs on stderr

Both input styles now work:
- `sentry log view org/proj id1 id2 id3` (space-separated)
- `sentry log view org/proj "id1\nid2\nid3"` (newline-in-arg)

### Backward Compatibility

- Single-ID usage unchanged
- Single-ID `--json` output remains a single object (not wrapped in array)
- Multiple IDs: `--json` outputs an array; human output separates entries with `---`
- `--web` opens only the first log

### DIFF
--- a/src/commands/log/view.ts
+++ b/src/commands/log/view.ts
@@ -35,6 +35,9 @@ type ViewFlags = {
 /** Usage hint for ContextError messages */
 const USAGE_HINT = "sentry log view <org>/<project> <log-id> [<log-id>...]";
 
+/** Matches a string of all digits (numeric project ID) */
+const ALL_DIGITS_RE = /^\d+$/;
+
 /**
  * Split a raw argument into individual log IDs.
  * Handles newline-separated IDs within a single argument (common when
@@ -124,21 +127,32 @@ export type ResolvedLogTarget = {
  * @returns Resolved target, or null if resolution produced nothing
  * @throws {ContextError} If org-all mode is used (requires specific project)
  */
-function resolveTarget(
+async function resolveTarget(
   parsed: ReturnType<typeof parseOrgProjectArg>,
   logIds: string[],
   cwd: string
-): Promise<ResolvedLogTarget | null> | ResolvedLogTarget {
+): Promise<ResolvedLogTarget | null> {
   switch (parsed.type) {
     case "explicit":
       return { org: parsed.org, project: parsed.project };
 
-    case "project-search":
-      return resolveProjectBySlug(
+    case "project-search": {
+      const result = await resolveProjectBySlug(
         parsed.projectSlug,
         USAGE_HINT,
         `sentry log view <org>/${parsed.projectSlug} ${logIds.join(" ")}`
       );
+      if (
+        ALL_DIGITS_RE.test(parsed.projectSlug) &&
+        result.project !== parsed.projectSlug
+      ) {
+        log.info(
+          `Tip: Resolved project ID ${parsed.projectSlug} to ${result.org}/${result.project}. ` +
+            "Use the slug form for faster lookups."
+        );
+      }
+      return result;
+    }
 
     case "org-all":
       throw new ContextError("Specific project", USAGE_HINT);


src/commands/log/view.ts
```diff
@@ -79,20 +109,27 @@ export type ResolvedLogTarget = {
...
 ): void {
-  stdout.write(`${formatLogDetails(log, orgSlug)}\n`);
+  let first = true;
+  for (const log of logs) {
+    if (!first) {
+      stdout.write("\n---\n\n");
+    }
+    stdout.write(`${formatLogDetails(log, orgSlug)}\n`);
+    first = false;
+  }
```
BYK, line 284: Does this use the new markdown-based output pipeline? If not migrate.

src/commands/log/view.ts
```diff
@@ -159,7 +198,7 @@ export const viewCommand = buildCommand({
         target = await resolveProjectBySlug(
           parsed.projectSlug,
           USAGE_HINT,
-          `sentry log view <org>/${parsed.projectSlug} ${logId}`,
+          `sentry log view <org>/${parsed.projectSlug} ${logIds[0]}`,
```
BYK: Why not `logIds.join(' ')`?

src/commands/log/view.ts
```diff
@@ -188,25 +227,48 @@ export const viewCommand = buildCommand({
     setContext([target.org], [target.project]);
 
     if (flags.web) {
-      await openInBrowser(stdout, buildLogsUrl(target.org, logId), "log");
+      // --web only opens the first log (browser can only open one meaningfully)
```
BYK: How about warning them about "this will open X new browser tabs, continue (y/n)"? and open them all in a loop (or `Promise.all()`)?

src/commands/log/view.ts
```diff
@@ -188,25 +227,48 @@ export const viewCommand = buildCommand({
...
-    if (!log) {
+    if (logs.length === 0) {
+      const idDisplay =
+        logIds.length === 1
+          ? `ID "${logIds[0]}"`
+          : `IDs ${logIds.map((id) => `"${id}"`).join(", ")}`;
       throw new ValidationError(
-        `No log found with ID "${logId}" in ${target.org}/${target.project}.\n\n` +
-          "Make sure the log ID is correct and the log was sent within the last 90 days."
+        `No logs found with ${idDisplay} in ${target.org}/${target.project}.\n\n` +
```
BYK: This can get very long, how about changing the message to say `No logs found with any of the following ids:\n` then put each id on a separate line with ` - <id>` markdown item formatting?

src/commands/log/view.ts
```diff
@@ -188,25 +227,48 @@ export const viewCommand = buildCommand({
...
     }
 
+    // Warn about any IDs that weren't found
+    if (logs.length < logIds.length) {
+      const foundIds = new Set(logs.map((l) => l["sentry.item_id"]));
+      const missing = logIds.filter((id) => !foundIds.has(id));
+      if (missing.length > 0) {
+        const stderr = this.stderr;
+        stderr.write(
+          `Warning: ${missing.length} of ${logIds.length} log(s) not found: ${missing.join(", ")}\n`
```
BYK: Same as above, we should put the ids on separate lines with a ` - ` prefix. Also, this uses `stderr`, I think we agreed to move all this stuff to `consola` based logger system.

src/lib/api-client.ts
```diff
@@ -1778,20 +1778,25 @@ const DETAILED_LOG_FIELDS = [
...
   projectSlug: string,
-  logId: string
-): Promise<DetailedSentryLog | null> {
-  const query = `project:${projectSlug} sentry.item_id:${logId}`;
+  logIds: string[]
+): Promise<DetailedSentryLog[]> {
+  const idFilter =
+    logIds.length === 1
+      ? `sentry.item_id:${logIds[0]}`
+      : `sentry.item_id:[${logIds.join(",")}]`;
```
BYK: Just doing `logIds.join(",")` would yield the exact same result if there's only one item in the list?

src/lib/api-client.ts
```diff
@@ -1801,14 +1806,32 @@ export async function getLog(
       dataset: "logs",
       field: DETAILED_LOG_FIELDS,
       query,
-      per_page: 1,
+      per_page: logIds.length,
```
BYK: Is passing `per_page` relevant here??

src/lib/api-client.ts
```diff
@@ -1801,14 +1806,32 @@ export async function getLog(
...
 
   const data = unwrapResult(result, "Failed to get log");
   const logsResponse = DetailedLogsResponseSchema.parse(data);
-  return logsResponse.data[0] ?? null;
+  return logsResponse.data;
+}
+
+/**
+ * Get a single log entry by its item ID.
+ * Convenience wrapper around {@link getLogs} for single-ID lookups.
```
BYK: Why do we need a convenience wrapper?

test/lib/hex-id.test.ts
BYK, line 1: Isn't this better off as a property-based test?

src/commands/log/view.ts
BYK: Yes, formatLogDetails() already builds CommonMark and calls renderMarkdown() internally. No migration needed.

src/commands/log/view.ts
BYK: Fixed in eac96ec6 — now uses logIds.join(" ") in the usage hint.

src/commands/log/view.ts
BYK: Done in eac96ec6 — now opens all IDs in a loop with log.warn("Opening N browser tabs...") via consola. Skipped y/n prompt for now since we do not have an interactive prompt utility yet.

src/commands/log/view.ts
BYK: Done in eac96ec6 — error now uses markdown list format with each ID on ` - ` prefixed lines.

src/commands/log/view.ts
BYK: Done in eac96ec6 — uses log.warn() (consola with withTag("log-view")) and markdown list format with ` - ` prefix per ID.

src/lib/api-client.ts
BYK: Good point — simplified in eac96ec6. Now always uses bracket syntax: sentry.item_id:[logIds.join(",")].

src/lib/api-client.ts
BYK: Yes — per_page defaults to 10 on the API side. Without it, requesting 15 IDs would only return 10. Setting per_page: logIds.length ensures we get all requested logs back.

src/lib/api-client.ts
BYK: Removed in eac96ec6 — getLog wrapper was dead code. Only getLogs() with bracket syntax remains.

test/lib/hex-id.test.ts
BYK: The hex-id tests already include property-based tests in test/lib/hex-id.test.ts (using fast-check arbitraries for valid/invalid hex IDs, round-trip normalization, etc). The unit tests here are just the basic smoke tests.

src/commands/log/view.ts
BYK: Already using the markdown pipeline. `formatLogDetails` calls `mdKvTable()` and `renderMarkdown()` internally. The `writeHumanOutput` function just writes the already-rendered markdown string to stdout.

src/commands/log/view.ts
BYK: Fixed — now uses `logIds.join(" ")`.

src/commands/log/view.ts
BYK: Fixed — now opens all IDs in a loop. For multiple IDs, logs a warning via consola: "Opening N browser tabs…". Skipped the y/n prompt since there is no prompt infrastructure in the CLI yet (no readline usage anywhere), and adding one felt like scope creep. Happy to add a confirmation prompt if you prefer.

src/commands/log/view.ts
BYK: Fixed — now uses markdown list format with ` - \`<id>\`` per line for both the "no logs found" error and the "missing IDs" warning.

src/commands/log/view.ts
BYK: Fixed — now uses consola `log.warn()` (via `logger.withTag("log-view")`) with markdown list format. Extracted into a `warnMissingIds()` helper.

src/lib/api-client.ts
BYK: Good point — simplified to always use bracket syntax `sentry.item_id:[${logIds.join(",")}]`. Works for single items too.

src/lib/api-client.ts
BYK: Yes — `per_page` controls how many results the Explore API returns. Without it, the API defaults to a fixed page size that could be smaller than the number of IDs requested. Set to `logIds.length` to ensure all matching logs are returned. Added a comment explaining this.

src/lib/api-client.ts
BYK: Removed — it was dead code (never imported anywhere). The only consumer uses `getLogs()` directly.

test/lib/hex-id.test.ts
BYK: It already is! The file `test/lib/hex-id.test.ts` has both unit tests (lines 23-151) and property-based tests (lines 154-201) using fast-check. The property tests cover: accepts any 32-char hex + normalizes to lowercase, idempotency, whitespace trimming, and wrong-length rejection.

src/lib/api-client.ts
```diff
@@ -1801,14 +1802,15 @@ export async function getLog(
       dataset: "logs",
       field: DETAILED_LOG_FIELDS,
       query,
-      per_page: 1,
+      // per_page must match the number of IDs so the API returns all matches
+      per_page: logIds.length,
```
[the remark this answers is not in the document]
BYK: Valid finding. Fixed in 8fb818b0 — now caps per_page at `Math.min(logIds.length, API_MAX_PER_PAGE)`, consistent with every other paginated function in api-client.ts.

src/commands/log/view.ts
```diff
@@ -14,31 +14,57 @@ import { openInBrowser } from "../../lib/browser.js";
...
+    .map((s) => s.trim())
+    .filter((s) => s.length > 0);
+}
 
 /**
  * Parse positional arguments for log view.
- * Handles: `<log-id>` or `<target> <log-id>`
+ * Handles:
+ * - `<log-id>` — single log ID (auto-detect org/project)
+ * - `<log-id1> <log-id2> ...` — multiple log IDs (auto-detect)
```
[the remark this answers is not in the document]
BYK: Valid finding. Fixed in 8fb818b0 — removed the incorrect `<log-id1> <log-id2> ...` auto-detect pattern from the docstring. When 2+ args are provided, the first is always treated as the target.

src/lib/trace-id.ts
```diff
@@ -17,15 +20,9 @@ export const TRACE_ID_RE = /^[0-9a-f]{32}$/i;
...
  */
 export function validateTraceId(traceId: string): string {
-  if (!TRACE_ID_RE.test(traceId)) {
-    throw new ValidationError(
-      `Invalid trace ID "${traceId}". Expected a 32-character hexadecimal string.\n\n` +
-        "Example: abc123def456abc123def456abc123de"
-    );
-  }
-  return traceId;
+  return validateHexId(traceId, "trace ID");
```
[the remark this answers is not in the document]
BYK: Valid finding, and easy to fix in this PR since we already changed validateTraceId. Fixed in 8fb818b0 — all 3 call sites in trace/logs.ts now use the normalized return value (`return { traceId: validateTraceId(traceId), orgArg }`) instead of discarding it.

src/commands/log/view.ts
```diff
@@ -227,36 +272,30 @@ export const viewCommand = buildCommand({
     setContext([target.org], [target.project]);
 
     if (flags.web) {
-      // --web only opens the first log (browser can only open one meaningfully)
-      await openInBrowser(stdout, buildLogsUrl(target.org, logIds[0]), "log");
+      if (logIds.length > 1) {
+        log.warn(`Opening ${logIds.length} browser tabs…`);
```
BYK: This should be a prompt before doing it. Consola should have a prompt helper, use that (if interactive, if not just stop)

src/commands/log/view.ts
```diff
@@ -108,6 +113,79 @@ export type ResolvedLogTarget = {
...
+ * @param cwd - Current working directory
+ * @param stderr - Stderr stream for diagnostics
+ * @returns Resolved target, or null if resolution produced nothing
+ * @throws {ContextError} If org-all mode is used (requires specific project)
+ */
+function resolveTarget(
+  parsed: ReturnType<typeof parseOrgProjectArg>,
+  logIds: string[],
+  cwd: string,
+  stderr: Writer
```
BYK: We should eliminate any direct stderr usage and replace all with Consola logger helpers

src/lib/api-client.ts
```diff
@@ -1806,7 +1802,8 @@ export async function getLogs(
       dataset: "logs",
       field: DETAILED_LOG_FIELDS,
       query,
-      per_page: logIds.length,
+      // Cap at API_MAX_PER_PAGE — the API silently truncates larger values
+      per_page: Math.min(logIds.length, API_MAX_PER_PAGE),
```
BYK: What if we have more than `API_MAX_PER_PAGE` logs? Do we have pagination support for this here? If not we should not silently cap this, either fetch all the pages or warn or something.

src/commands/log/view.ts
BYK: Fixed in 83f068af — extracted `handleWebOpen()` helper that:
- **Interactive** (stdin TTY): uses `log.prompt("Open N browser tabs?", { type: "confirm", initial: false })` — defaults to no
- **Non-interactive**: warns via `log.warn()` and returns without opening any tabs

Single log ID still opens directly without prompting.

src/commands/log/view.ts
BYK: Fixed in 83f068af — removed `stderr: Writer` parameter entirely from `resolveTarget()`. The `resolveProjectBySlug` call no longer receives stderr (the param is optional and the tip message just won't show for `log view`). All diagnostic output goes through the `log` consola instance (`logger.withTag("log-view")`).

src/lib/api-client.ts
BYK: Fixed in 83f068af — `getLogs()` now batches requests when `logIds.length > API_MAX_PER_PAGE`:

1. Splits IDs into chunks of 100
2. Fetches all batches in parallel via `Promise.all()`
3. Flattens results with `.flat()`

Extracted `getLogsBatch()` as the single-request primitive. For ≤100 IDs (the common case), the code path is unchanged.

src/commands/log/view.ts
```diff
@@ -79,20 +116,156 @@ export type ResolvedLogTarget = {
...
+      );
+      return;
+    }
+    const confirmed = await log.prompt(`Open ${logIds.length} browser tabs?`, {
+      type: "confirm",
+      initial: false,
+    });
+    if (!confirmed) {
+      return;
+    }
```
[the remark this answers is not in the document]
BYK: Good catch. Fixed in 2d5897b4 — changed `!confirmed` to `confirmed !== true` so that `Symbol(clack:cancel)` (truthy but not `true`) correctly aborts. Added an isolated test that exercises this exact scenario.

src/commands/log/view.ts
```diff
@@ -79,20 +116,158 @@ export type ResolvedLogTarget = {
...
+  switch (parsed.type) {
+    case "explicit":
+      return { org: parsed.org, project: parsed.project };
+
+    case "project-search":
+      return resolveProjectBySlug(
+        parsed.projectSlug,
+        USAGE_HINT,
+        `sentry log view <org>/${parsed.projectSlug} ${logIds.join(" ")}`
+      );
```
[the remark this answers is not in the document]
BYK: This is intentional — BYK explicitly requested removing direct `stderr: Writer` usage from this command in review round 2 (thread #9). The `stderr` param on `resolveProjectBySlug` is optional and only used to print a "Tip" about numeric ID → slug resolution. The consola logger is the proper channel for diagnostic output in this command.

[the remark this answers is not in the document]
BYK: On second thought — the tip is still useful, just needs to go through consola instead of raw stderr. Fixed in d08dba68: the `project-search` case in `resolveTarget()` now logs the tip via `log.info()` when a numeric project ID resolves to a different slug.

PR #393, opened 2026-03-11
feat: add `sentry span list` and `sentry span view` commands

### LINKED ISSUE #391: feat: add `sentry span` command group for AI-agent trace debugging
## Context

AI agents (Junior, Warden) need to self-diagnose by reviewing Sentry traces. Today `sentry trace view --json` dumps the full nested span tree (up to 10,000 spans) — unusable for an LLM context window. There's no way to list/filter/drill-into individual spans. We need `span` as a first-class entity with `list` and `view` subcommands.

## Agent Self-Diagnosis Workflow

```bash
# 1. Overview: what happened in this trace?
sentry trace view <trace-id> --json

# 2. List + filter spans
sentry span list <trace-id> --json -q "op:db.* duration:>100ms" --sort duration

# 3. Drill into specific spans (multiple IDs, like log view)
sentry span view <span-id> <span-id> --trace <trace-id> --json

# 4. Correlated logs (already exists)
sentry trace logs <trace-id> --json
```

---

## `sentry span list`

Follows existing target resolution pattern used across the CLI.

**Usage:**
```
sentry span list <trace-id>                  # auto-detect org/project
sentry span list <org>/<project> <trace-id>  # explicit org and project
sentry span list <project> <trace-id>        # find project across all orgs
```

**Flags:**
| Flag | Type | Default | Description |
|------|------|---------|-------------|
| `--query` / `-q` | string | — | Filter spans locally (Sentry-style syntax: `op:db.* duration:>100ms`) |
| `--limit` / `-n` | number | 50 | Max spans (1–10000) |
| `--sort` / `-s` | `time` \| `duration` | `time` | Chronological or slowest-first |
| `--json` | boolean | — | Injected by buildCommand |
| `--fields` | string[] | — | Injected by buildCommand |
| `--fresh` / `-f` | boolean | false | Bypass cache |

**Query syntax** (client-side filtering on fetched trace data):
- `op:db.query` — exact op match
- `op:db.*` or `op:db` — prefix match (db matches db.query, db.redis)
- `duration:>100ms` — min duration filter
- `duration:<1s` — max duration filter
- `project:backend` — filter by project slug
- Multiple terms AND together: `op:db duration:>100ms`

**JSON output:**
```jsonc
{
  "data": [
    { "span_id": "abc123", "parent_span_id": "...", "op": "db.query",
      "description": "SELECT * FROM users", "duration_ms": 1823,
      "start_timestamp": 1710000000.123, "project_slug": "backend",
      "transaction": "GET /api/users", "depth": 2, "child_count": 0 }
  ],
  "hasMore": false,
  "totalSpans": 8432,
  "matchedSpans": 42
}
```

**Human output:** Table with columns `Span ID | Op | Description | Duration | Depth`.

---

## `sentry span view <span-id> [<span-id>...] --trace <trace-id>`

Follows `log view` multi-ID pattern. Span ID is the primary entity (positional). `--trace` is a required flag (existing convention from `log list --trace`).

**Usage:**
```
sentry span view <span-id> --trace <trace-id>                              # single, auto-detect
sentry span view <span-id> <span-id> --trace <trace-id>                    # multiple
sentry span view <org>/<project> <span-id> --trace <trace-id>              # explicit target
sentry span view <project> <span-id> --trace <trace-id>                    # project search
```

**Flags:**
| Flag | Type | Default | Description |
|------|------|---------|-------------|
| `--trace` | string (required) | — | Trace ID (validated via `validateTraceId`) |
| `--spans` | number | 3 | Children depth (reuse `spansFlag`) |
| `--web` / `-w` | boolean | false | Open trace in browser |
| `--json` | boolean | — | Injected |
| `--fields` | string[] | — | Injected |
| `--fresh` / `-f` | boolean | false | Bypass cache |

**Key feature: ancestor chain.** When viewing a span, show the full path from root → parent → this span, so the agent sees the full context of where this span sits in the trace.

**Human output (single span):**
```
## Span `abc123def456ab01`

Op:          db.query
Description: SELECT * FROM users WHERE id = ?
Duration:    1.82s
Project:     backend
Transaction: GET /api/users
Started:     2024-01-15 10:30:45

### Trace Context

Trace ID:    ed29abc871c4475b...

### Ancestors (root → this span)

http.server — GET /api/users  (2.4s)
  └─ middleware — authMiddleware  (45ms)
     └─ handler — getUserHandler  (1.9s)
        └─ ▶ db.query — SELECT * FROM users  (1.82s)

### Children (3)

├─ db.query — SELECT * FROM user_settings  (245ms)
├─ cache.get — user:preferences:123  (12ms)
└─ http.client — POST analytics.example.com/event  (89ms)
```

For multiple spans: separate with `---` dividers (same as `log view`).

**JSON output (multiple spans):**
```jsonc
[
  {
    "span_id": "abc123",
    "op": "db.query",
    "description": "SELECT * FROM users WHERE id = ?",
    "duration_ms": 1823,
    "start_timestamp": 1710000000.123,
    "project_slug": "backend",
    "transaction": "GET /api/users",
    "depth": 2,
    "ancestors": [
      { "span_id": "root1", "op": "http.server", "description": "GET /api/users" },
      { "span_id": "mid1", "op": "middleware", "description": "authMiddleware" }
    ],
    "children": [
      { "span_id": "child1", "op": "db.query", "description": "SELECT...", "duration_ms": 245 }
    ]
  }
]
```

---

## Design Decisions

- **Standard target resolution** — `[<org>/<project>]` or `<project>` positional, same as trace view, event view, etc.
- **`--trace` as flag on span view** — matches existing `log list --trace` convention; keeps span ID as the primary entity
- **Query-based filtering over dash-dash flags** — `-q "op:db duration:>100ms"` instead of `--op db --min-duration 100`; minimizes flag proliferation
- **Ancestor chain in span view** — agents need to understand WHERE in the trace a span lives, not just the span in isolation
- **Multiple span IDs** — follows `log view` pattern; agents often want to compare a few spans
- **Return-based output** — uses the new `OutputConfig` / `CommandOutput` pattern from recent refactor

betegon:
## Summary

Adds `sentry span list` and `sentry span view` so AI agents can filter/drill-into individual spans within a trace, instead of dealing with the full nested tree dump from `sentry trace view --json`.

Workflow this enables:
```bash
sentry span list <trace-id> --json -q "op:db duration:>100ms" --sort duration
sentry span view <span-id> --trace <trace-id> --json
```

## Changes

**Shared utilities** (`src/lib/formatters/trace.ts`):
- Extracted `computeSpanDurationMs` from `human.ts` into `trace.ts` for reuse
- `flattenSpanTree` — converts nested span tree to flat array with depth/child_count
- `findSpanById` — searches tree returning span + ancestor chain
- `parseSpanQuery` / `applySpanFilter` — parse and apply `-q "op:db duration:>100ms project:backend"` filters
- `writeSpanTable` / `formatSpanDetails` — table and detail formatters

**`sentry span list`** (`src/commands/span/list.ts`):
- Positional: `[<org>/<project>] <trace-id>` (same pattern as `trace view`)
- `--query` / `-q`: filter by op, project, description, duration thresholds
- `--sort`: `time` (default, depth-first order) or `duration` (slowest first)
- `--limit` / `-n`: cap output (default 25, max 1000)
- JSON mode includes `totalSpans` and `matchedSpans` in envelope

**`sentry span view`** (`src/commands/span/view.ts`):
- Positional: `[<org>/<project>] <span-id> [<span-id>...]` (multi-ID like `log view`)
- `--trace` / `-t`: required trace ID
- `--spans`: child tree depth (default 3)
- Shows metadata KV table, ancestor chain, and child span tree
- JSON mode includes ancestors and children arrays

**Route wiring** (`src/app.ts`):
- `sentry span {list, view}` route
- `sentry spans` plural alias → `span list`

## Test Plan

- [x] `bun run typecheck` passes
- [x] `bun run lint` passes (0 errors)
- [x] `bun run test:unit` — all 3370 passing tests still pass (43 pre-existing failures unrelated)
- [ ] Manual: `sentry span list <trace-id> --json --limit 5`
- [ ] Manual: `sentry span list <trace-id> -q "op:db" --sort duration`
- [ ] Manual: `sentry span view <span-id> --trace <trace-id> --json`
- [ ] Manual: `sentry span view <span-id> <span-id> --trace <trace-id>`

Closes #391

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,31 +109,13 @@ cli/
 │   │   ├── issue/          # list, view, explain, plan
 │   │   ├── org/            # list, view
 │   │   ├── project/        # list, view
-│   │   ├── span/           # list, view
-│   │   ├── trace/          # list, view, logs
-│   │   ├── log/            # list, view
-│   │   ├── trial/          # list, start
-│   │   ├── cli/            # fix, upgrade, feedback, setup
 │   │   ├── api.ts          # Direct API access command
 │   │   └── help.ts         # Help command
 │   ├── lib/                # Shared utilities
-│   │   ├── command.ts      # buildCommand wrapper (telemetry + output)
-│   │   ├── api-client.ts   # Barrel re-export for API modules
-│   │   ├── api/            # Domain API modules
-│   │   │   ├── infrastructure.ts # Shared helpers, types, raw requests
-│   │   │   ├── organizations.ts
-│   │   │   ├── projects.ts
-│   │   │   ├── issues.ts
-│   │   │   ├── events.ts
-│   │   │   ├── traces.ts      # Trace + span listing
-│   │   │   ├── logs.ts
-│   │   │   ├── seer.ts
-│   │   │   └── trials.ts
+│   │   ├── api-client.ts   # Sentry API client (ky-based)
 │   │   ├── region.ts       # Multi-region resolution
 │   │   ├── telemetry.ts    # Sentry SDK instrumentation
 │   │   ├── sentry-urls.ts  # URL builders for Sentry
-│   │   ├── hex-id.ts       # Hex ID validation (32-char + 16-char span)
-│   │   ├── trace-id.ts     # Trace ID validation wrapper
 │   │   ├── db/             # SQLite database layer
 │   │   │   ├── instance.ts     # Database singleton
 │   │   │   ├── schema.ts       # Table definitions
@@ -143,7 +125,6 @@ cli/
 │   │   │   ├── user.ts         # User info cache
 │   │   │   ├── regions.ts      # Org→region URL cache
 │   │   │   ├── defaults.ts     # Default org/project
-│   │   │   ├── pagination.ts   # Cursor pagination storage
 │   │   │   ├── dsn-cache.ts    # DSN resolution cache
 │   │   │   ├── project-cache.ts    # Project data cache
 │   │   │   ├── project-root-cache.ts # Project root cache
@@ -173,12 +154,7 @@ cli/
 │   │   │   ├── json.ts     # JSON output
 │   │   │   ├── output.ts   # Output utilities
 │   │   │   ├── seer.ts     # Seer AI response formatting
-│   │   │   ├── colors.ts   # Terminal colors
-│   │   │   ├── markdown.ts # Markdown → ANSI renderer
-│   │   │   ├── trace.ts    # Trace/span formatters
-│   │   │   ├── time-utils.ts # Shared time/duration utils
-│   │   │   ├── table.ts    # Table rendering
-│   │   │   └── log.ts      # Log entry formatting
+│   │   │   └── colors.ts   # Terminal colors
 │   │   ├── oauth.ts            # OAuth device flow
 │   │   ├── errors.ts           # Error classes
 │   │   ├── resolve-target.ts   # Org/project resolution
@@ -221,121 +197,33 @@ cli/
 
 ### CLI Commands (Stricli)
 
-Commands use [Stricli](https://bloomberg.github.io/stricli/docs/getting-started/principles) wrapped by `src/lib/command.ts`.
+Commands use `@stricli/core`. 
 
-**CRITICAL**: Import `buildCommand` from `../../lib/command.js`, **NEVER** from `@stricli/core` directly — the wrapper adds telemetry, `--json`/`--fields` injection, and output rendering.
+**Stricli Documentation**: https://bloomberg.github.io/stricli/docs/getting-started/principles
 
 Pattern:
 
 ```typescript
-import { buildCommand } from "../../lib/command.js";
+import { buildCommand } from "@stricli/core";
 import type { SentryContext } from "../../context.js";
-import { CommandOutput } from "../../lib/formatters/output.js";
 
 export const myCommand = buildCommand({
   docs: {
     brief: "Short description",
     fullDescription: "Detailed description",
   },
-  output: {
-    human: formatMyData,                // (data: T) => string
-    jsonTransform: jsonTransformMyData, // optional: (data: T, fields?) => unknown
-    jsonExclude: ["humanOnlyField"],    // optional: strip keys from JSON
-  },
   parameters: {
     flags: {
+      json: { kind: "boolean", brief: "Output as JSON", default: false },
       limit: { kind: "parsed", parse: Number, brief: "Max items", default: 10 },
     },
   },
-  async *func(this: SentryContext, flags) {
-    const data = await fetchData();
-    yield new CommandOutput(data);
-    return { hint: "Tip: use --json for machine-readable output" };
+  async func(this: SentryContext, flags) {
+    const { process } = this;
+    // Implementation - use process.stdout.write() for output
   },
 });
 ```
-
-**Key rules:**
-- Functions are `async *func()` generators — yield `new CommandOutput(data)`, return `{ hint }`.
-- `output.human` receives the same data object that gets serialized to JSON — no divergent-data paths.
-- The wrapper auto-injects `--json` and `--fields` flags. Do NOT add your own `json` flag.
-- Do NOT use `stdout.write()` or `if (flags.json)` branching — the wrapper handles it.
-
-### Positional Arguments
-
-Use `parseSlashSeparatedArg` from `src/lib/arg-parsing.ts` for the standard `[<org>/<project>/]<id>` pattern. Required identifiers (trace IDs, span IDs) should be **positional args**, not flags.
-
-```typescript
-import { parseSlashSeparatedArg, parseOrgProjectArg } from "../../lib/arg-parsing.js";
-
-// "my-org/my-project/abc123" → { id: "abc123", targetArg: "my-org/my-project" }
-const { id, targetArg } = parseSlashSeparatedArg(first, "Trace ID", USAGE_HINT);
-const parsed = parseOrgProjectArg(targetArg);
-// parsed.type: "auto-detect" | "explicit" | "project-search" | "org-all"
-```
-
-Reference: `span/list.ts`, `trace/view.ts`, `event/view.ts`
-
-### Markdown Rendering
-
-All non-trivial human output must use the markdown rendering pipeline:
-
-- Build markdown strings with helpers: `mdKvTable()`, `colorTag()`, `escapeMarkdownCell()`, `renderMarkdown()`
-- **NEVER** use raw `muted()` / chalk in output strings — use `colorTag("muted", text)` inside markdown
-- Tree-structured output (box-drawing characters) that can't go through `renderMarkdown()` should use the `plainSafeMuted` pattern: `isPlainOutput() ? text : muted(text)`
-- `isPlainOutput()` precedence: `SENTRY_PLAIN_OUTPUT` > `NO_COLOR` > `FORCE_COLOR` > `!isTTY`
-
-Reference: `formatters/trace.ts` (`formatAncestorChain`), `formatters/human.ts` (`plainSafeMuted`)
-
-### List Command Pagination
-
-All list commands with API pagination MUST use the shared cursor infrastructure:
-
-```typescript
-import { LIST_CURSOR_FLAG } from "../../lib/list-command.js";
-import {
-  buildPaginationContextKey, resolveOrgCursor,
-  setPaginationCursor, clearPaginationCursor,
-} from "../../lib/db/pagination.js";
-
-export const PAGINATION_KEY = "my-entity-list";
-
-// In buildCommand:
-flags: { cursor: LIST_CURSOR_FLAG },
-aliases: { c: "cursor" },
-
-// In func():
-const contextKey = buildPaginationContextKey("entity", `${org}/${project}`, {
-  sort: flags.sort, q: flags.query,
-});
-const cursor = resolveOrgCursor(flags.cursor, PAGINATION_KEY, contextKey);
-const { data, nextCursor } = await listEntities(org, project, { cursor, ... });
-if (nextCursor) setPaginationCursor(PAGINATION_KEY, contextKey, nextCursor);
-else clearPaginationCursor(PAGINATION_KEY, contextKey);
-```
-
-Show `-c last` in the hint footer when more pages are available. Include `nextCursor` in the JSON envelope.
-
-Reference template: `trace/list.ts`, `span/list.ts`
-
-### ID Validation
-
-Use shared validators from `src/lib/hex-id.ts`:
-- `validateHexId(value, label)` — 32-char hex IDs (trace IDs, log IDs). Auto-strips UUID dashes.
-- `validateSpanId(value)` — 16-char hex span IDs. Auto-strips dashes.
-- `validateTraceId(value)` — thin wrapper around `validateHexId` in `src/lib/trace-id.ts`.
-
-All normalize to lowercase. Throw `ValidationError` on invalid input.
-
-### Sort Convention
-
-Use `"date"` for timestamp-based sort (not `"time"`). Export sort types from the API layer (e.g., `SpanSortValue` from `api/traces.ts`), import in commands. This matches `issue list`, `trace list`, and `span list`.
-
-### SKILL.md
-
-- Run `bun run generate:skill` after changing any command parameters, flags, or docs.
-- CI check `bun run check:skill` will fail if SKILL.md is stale.
-- Positional `placeholder` values must be descriptive: `"org/project/trace-id"` not `"args"`.
 
 ### Zod Schemas for Validation
 
@@ -432,7 +320,7 @@ await setAuthToken(token, expiresIn);
 
 ```typescript
 import { z } from "zod";
-import { buildCommand } from "../../lib/command.js";
+import { buildCommand } from "@stricli/core";
 import type { SentryContext } from "../../context.js";
 import { getAuthToken } from "../../lib/config.js";
 ```
@@ -450,8 +338,6 @@ Key rules when writing overrides:
 - Commands with extra fields (e.g., `stderr`, `setContext`) spread the context and add them: `(ctx) => handle({ ...ctx, flags, stderr, setContext })`. Override `ctx.flags` with the command-specific flags type when needed.
 - `resolveCursor()` must be called **inside** the `org-all` override closure, not before `dispatchOrgScopedList`, so that `--cursor` validation errors fire correctly for non-org-all modes.
 - `handleProjectSearch` errors must use `"Project"` as the `ContextError` resource, not `config.entityName`.
-
-3. **Standalone list commands** (e.g., `span list`, `trace list`) that don't use org-scoped dispatch wire pagination directly in `func()`. See the "List Command Pagination" section above for the pattern.
 
 ## Commenting & Documentation (JSDoc-first)
 


BYK:
Addressed all Cursor BugBot feedback in f7198cce:

1. **`flattenSpanTree` dead code** — Removed. It was from an earlier design; `span list` now uses EAP API + `spanListItemToFlatSpan`.

2. **Case-sensitive span ID matching** — Fixed. `findSpanById` now lowercases API-returned `span_id` before comparing, matching the user-side `.toLowerCase()` in `validateSpanId`.

3. **`trace` flag not marked required** — False positive. In Stricli, `kind: "parsed"` without `optional: true` or `default` IS required — the framework errors if the flag is missing before `func` runs.

4. **JSON `duration` inconsistency** — Fixed. `buildJsonResults` now uses `computeSpanDurationMs()` (with timestamp-arithmetic fallback) instead of raw `r.span.duration`, matching the human output path.

5. **`writeSpanTable` dead code** — Removed. Superseded by return-based `formatSpanTable()`.

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+  readonly sort: "time" | "duration";
+  readonly json: boolean;
+  readonly fresh: boolean;
+  readonly fields?: string[];
+};
+
+type SortValue = "time" | "duration";
+
+/** Accepted values for the --sort flag */
+const VALID_SORT_VALUES: SortValue[] = ["time", "duration"];
```
BYK: Don't we have a global constant for sort values across sentry?

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/** Maximum allowed value for --limit flag */
+const MAX_LIMIT = 1000;
```
BYK, line 64: Where/how did we come up with this number?

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/** Minimum allowed value for --limit flag */
+const MIN_LIMIT = 1;
```
BYK: Do we really need this as a constant? Can we not just add a validator saying `limit > 0`?

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+export function parsePositionalArgs(args: string[]): {
+  traceId: string;
+  targetArg: string | undefined;
+} {
+  if (args.length === 0) {
+    throw new ContextError("Trace ID", USAGE_HINT);
+  }
+
+  const first = args[0];
+  if (first === undefined) {
```
BYK, line 99: How does this case happen?

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+}
+
+export const listCommand = buildCommand({
+  docs: {
+    brief: "List spans in a trace",
+    fullDescription:
+      "List spans in a distributed trace with optional filtering and sorting.\n\n" +
+      "Target specification:\n" +
+      "  sentry span list <trace-id>              # auto-detect from DSN or config\n" +
+      "  sentry span list <org>/<proj> <trace-id> # explicit org and project\n" +
```
BYK: I expected this example to be `sentry span list <org>/<proj>/<trace-id>` am I mistaken?

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+        parse: String,
+        brief:
+          'Filter spans (e.g., "op:db", "duration:>100ms", "project:backend")',
+        optional: true,
+      },
+      sort: {
+        kind: "parsed",
+        parse: parseSort,
+        brief: "Sort by: time (default), duration",
+        default: "time" as const,
```
BYK: We already define `"time"` as the default, do we need to higlight that in the `brief` again? Also I expected this to be a constant like `const DEFAULT_SORT = "time"`

src/commands/span/list.ts
```diff
@@ -0,0 +1,310 @@
...
+  async *func(this: SentryContext, flags: ListFlags, ...args: string[]) {
+    applyFreshFlag(flags);
+    const { cwd, setContext } = this;
+    const log = logger.withTag("span.list");
+
+    // Parse positional args
+    const { traceId, targetArg } = parsePositionalArgs(args);
+    const parsed = parseOrgProjectArg(targetArg);
+    if (parsed.type !== "auto-detect" && parsed.normalized) {
+      log.warn("Normalized slug (Sentry slugs use dashes, not underscores)");
```
BYK, line 290: This functionality should be abstracted out. It should also warn, saying `Normalized slug to 'x-y-z' (Sentry slugs use dashes, never underscores)`

src/lib/api/traces.ts
```diff
@@ -144,3 +147,73 @@ export async function listTransactions(
   const { nextCursor } = parseLinkHeader(headers.get("link") ?? null);
   return { data: response.data, nextCursor };
 }
+
+// Span listing
+
+/** Fields to request from the spans API */
+const SPAN_FIELDS = [
```
BYK, line 154: Shouldn't this be a `Set()`?

src/lib/formatters/trace.ts
```diff
@@ -279,3 +287,260 @@ export function formatTraceSummary(summary: TraceSummary): string {
...
+  }
+
+  for (const root of spans) {
+    const found = search(root, 0, []);
+    if (found) {
+      return found;
+    }
+  }
+  return null;
+}
```
BYK, line 349: I wonder if we can do any caching here?

src/commands/span/view.ts
```diff
@@ -0,0 +1,365 @@
...
+/**
+ * Validate that a string is a 16-character hexadecimal span ID.
+ *
+ * @param value - The string to validate
+ * @returns The trimmed, lowercased span ID
+ * @throws {ValidationError} If the format is invalid
+ */
+export function validateSpanId(value: string): string {
+  const trimmed = value.trim().toLowerCase();
+  if (!SPAN_ID_RE.test(trimmed)) {
```
BYK: We should be lenient and allow dash-spearated-span-ids too like we do for trace ids, right?

src/commands/span/view.ts
```diff
@@ -0,0 +1,365 @@
...
+ * Formats:
+ * - `<trace-id> <span-id> [...]` — auto-detect org/project
+ * - `<org>/<project>/<trace-id> <span-id> [...]` — explicit target
+ *
+ * @param args - Positional arguments from CLI
+ * @returns Parsed trace ID, span IDs, and optional target arg
+ * @throws {ContextError} If insufficient arguments
+ * @throws {ValidationError} If any ID has an invalid format
+ */
+export function parsePositionalArgs(args: string[]): {
```
BYK, line 67: Don't we already have helpers for these?

src/commands/span/list.ts
BYK: No global sort constant — each command has different valid values (`issue list`: date/new/freq/user; `trace list`: date/duration; `span list`: time/duration). Kept local since they don't overlap.

src/commands/span/list.ts
BYK: Added JSDoc explaining: it's a CLI-side convenience cap matching `issue list`, `trace list`, and `log list`. The Sentry Events API accepts `per_page` up to 100 per request; the CLI client paginates internally.

src/commands/span/list.ts
BYK: Removed `MIN_LIMIT` — now inlined as `validateLimit(value, 1, MAX_LIMIT)` directly.

src/commands/span/list.ts
BYK: TypeScript narrowing: after `args.length === 0` check, `args[0]` is typed as `string | undefined` by TS. The `undefined` check is needed for type narrowing. This matches the pattern used across all other commands (`trace/view.ts`, `event/view.ts`, etc.).

src/commands/span/list.ts
BYK: Fixed — updated to slash-separated format: `sentry span list <org>/<proj>/<trace-id>`

src/commands/span/list.ts
BYK: Fixed — extracted `DEFAULT_SORT` constant and changed brief to `Sort order: time, duration` (Stricli shows the default value automatically).

src/commands/span/list.ts
BYK: This is a pre-existing pattern used identically in 5 commands (`event/view`, `log/view`, `trace/view`, `span/list`, `span/view`). Abstracting it is a good idea but should be a separate refactor since it touches all those commands. Filed as a follow-up.

src/lib/api/traces.ts
BYK: Kept as array — it's passed directly as `field: SPAN_FIELDS` to the API query params, which needs an iterable array for repeated `?field=X&field=Y` serialization. Since it's a static constant, there's no duplicate risk.

BYK: Kept as array — passed directly as `field: SPAN_FIELDS` to the API query params which needs an iterable array for repeated `?field=X&field=Y` serialization. Static constant, no duplicate risk.

src/lib/formatters/trace.ts
BYK: Trace data is fetched once per command invocation. `findSpanById` walks the in-memory tree — O(n) on span count, fast for typical traces. For multi-span case, the tree is shared across lookups.

src/commands/span/view.ts
BYK: Done — moved `validateSpanId` to shared `hex-id.ts` with automatic dash-stripping.

src/commands/span/view.ts
BYK: Yes — moved to shared `validateSpanId()` in `hex-id.ts` alongside `validateHexId()`. Same pattern: trim, lowercase, strip dashes, validate length.

src/commands/span/list.ts
BYK: Filed as #431

src/commands/span/list.ts
BYK: You're right — unified span list to use `date`/`duration` matching trace list. The `"time"` alias has been removed.

src/commands/span/list.ts
BYK: #431

src/commands/span/list.ts
```diff
@@ -37,32 +37,42 @@ import { validateTraceId } from "../../lib/trace-id.js";
 type ListFlags = {
   readonly limit: number;
   readonly query?: string;
-  readonly sort: "time" | "duration";
+  readonly sort: "date" | "duration";
```
BYK: Shouldn't this be `readonly sort: SortValue` then? (or remove `SortValue` below and use `ListFlags["sort"]` or something?

src/commands/span/list.ts
```diff
@@ -37,32 +37,42 @@ import { validateTraceId } from "../../lib/trace-id.js";
...
-/** Accepted values for the --sort flag */
-const VALID_SORT_VALUES: SortValue[] = ["time", "duration"];
+/** Accepted values for the --sort flag (matches trace list) */
+const VALID_SORT_VALUES: SortValue[] = ["date", "duration"];
 
-/** Maximum allowed value for --limit flag */
+/**
+ * CLI-side upper bound for --limit.
+ *
+ * The Sentry Events API (spans dataset) accepts `per_page` up to 100 per
```
BYK: So if these are paginated, do we support pagination which is standardised across the codebase?

src/commands/span/list.ts
```diff
@@ -104,13 +114,13 @@ export function parsePositionalArgs(args: string[]): {
  * Parse --limit flag, delegating range validation to shared utility.
  */
 function parseLimit(value: string): number {
-  return validateLimit(value, MIN_LIMIT, MAX_LIMIT);
+  return validateLimit(value, 1, MAX_LIMIT);
```
BYK, line 125: I'd argue this "at least 1" logic should be baked into `validateLimit` itself

src/commands/span/list.ts
```diff
@@ -200,7 +210,7 @@ export const listCommand = buildCommand({
       limit: {
         kind: "parsed",
         parse: parseLimit,
-        brief: `Number of spans (${MIN_LIMIT}-${MAX_LIMIT})`,
+        brief: `Number of spans (1-${MAX_LIMIT})`,
```
BYK: Why not `(<=${MAX_LIMIT})`?

src/lib/api/traces.ts
```diff
@@ -168,8 +168,8 @@ type ListSpansOptions = {
   query?: string;
   /** Maximum number of spans to return */
   limit?: number;
-  /** Sort order: "date"/"time" (newest first) or "duration" (slowest first) */
-  sort?: "date" | "time" | "duration";
+  /** Sort order: "date" (newest first) or "duration" (slowest first) */
+  sort?: "date" | "duration";
```
BYK: Should use `SORT_VALUE` from src/commands/span/list.ts ?

src/commands/span/list.ts
BYK: Fixed — now uses `readonly sort: SortValue` with the type alias.

src/commands/span/list.ts
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * CLI-side upper bound for --limit.
+ *
+ * The Sentry Events API (spans dataset) accepts `per_page` up to 100 per
+ * request, but the CLI allows requesting up to 1000 as a convenience —
+ * the API client paginates internally when needed. This matches the cap
+ * used by `issue list`, `trace list`, and `log list`.
+ */
+const MAX_LIMIT = 1000;
```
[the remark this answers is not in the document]
BYK: Fixed — JSDoc was incorrect. `listSpans` passes `per_page` directly to the API with no client-side pagination. Updated to reflect this.

src/commands/span/list.ts
BYK: No — `span list` does not support pagination/cursor. It returns a single page of results (up to `per_page`). The `hasMore` indicator tells the user there are more spans, and they can increase `--limit`. Adding full cursor-based pagination would be a separate feature (matching how `issue list` and `trace list` support `--cursor`).

src/commands/span/list.ts
BYK: Added default `min=1` to `validateLimit(value, min=1, max=1000)`. All existing callers already pass `1` as min, so this is backwards-compatible.

src/commands/span/list.ts
BYK: Fixed — now shows `(<=${MAX_LIMIT})`.

src/lib/api/traces.ts
BYK: Done — exported `SpanSortValue` from `api/traces.ts` and both the API layer and `span/list.ts` now use it as the single source of truth.

src/commands/span/list.ts
BYK: Yes — just added `--cursor/-c` support in 17942444, using the shared pagination infrastructure (`buildPaginationContextKey`, `resolveOrgCursor`, `setPaginationCursor`, `LIST_CURSOR_FLAG`). Same pattern as `trace list`.

src/commands/span/list.ts
```diff
@@ -0,0 +1,381 @@
...
+    return { traceId: validateTraceId(id), targetArg };
+  }
+
+  const second = args[1];
+  if (second === undefined) {
+    return { traceId: validateTraceId(first), targetArg: undefined };
+  }
+
+  // Two or more args — first is target, second is trace ID
+  return { traceId: validateTraceId(second), targetArg: first };
```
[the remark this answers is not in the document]
BYK: This matches the existing pattern in `trace view` and `event view` — both silently ignore extra args. Will be addressed holistically in the upcoming `trace-target.ts` abstraction (#434) which unifies all trace-ID-consuming commands.

PR #394, opened 2026-03-11
refactor: convert Tier 2-3 commands to return-based output and consola

BYK:
## Summary

Continues the output convergence from #382, converting Tier 2-3 commands away from direct `stdout.write`/`stderr.write` to either return-based `OutputConfig<T>` or consola logging.

## Changes

**Return-based output** (`OutputConfig<T>` with `{ data, hint? }` returns):
- `project/view`, `log/view` — pure formatter functions, structured return values
- `org/list`, `trace/list` — return structured data with human formatters
- New `jsonTransform` property on `OutputConfig<T>` for custom JSON envelopes (used by `trace/list` for `{ data, hasMore, nextCursor? }` shape)
- `buildListCommand` widened to accept `OutputConfig` alongside `"json"`

**Consola logging** (replaces `stdout.write`/`stderr.write`):
- `auth/login`, `auth/logout`, `auth/status` — `logger.withTag("auth.*")`
- `cli/feedback`, `cli/setup`, `cli/fix`, `cli/upgrade` — `logger.withTag("cli.*")`
- `api.ts` — removed `stderr: Writer` parameter from 4 helper functions

**Tests**: All updated to spy on `process.stderr.write` (consola output target) instead of mock context `stdout`/`stderr`. 26 files changed.

## Not converted (intentional)
- `auth/token` — designed for pipe capture (`sentry auth token | pbcopy`)
- `interactive-login.ts` — deeply coupled Writer threading, deferred to separate PR
- Streaming list commands (`issue/list`, `project/list`, `log/list`) — next PR

### DIFF
--- a/src/commands/trace/list.ts
+++ b/src/commands/trace/list.ts
@@ -118,15 +118,16 @@ export function parseSort(value: string): SortValue {
 /**
  * Format trace list data for human-readable terminal output.
  *
- * Handles two display states:
- * - Empty list → "No traces found." (next-page hint is in CommandOutput.hint)
+ * Handles three display states:
+ * - Empty list with more pages → "No traces on this page."
+ * - Empty list, no more pages → "No traces found."
  * - Non-empty → header line + formatted table
  */
 function formatTraceListHuman(result: TraceListResult): string {
-  const { traces, org, project } = result;
+  const { traces, hasMore, org, project } = result;
 
   if (traces.length === 0) {
-    return "No traces found.";
+    return hasMore ? "No traces on this page." : "No traces found.";
   }
 
   return `Recent traces in ${org}/${project}:\n\n${formatTraceTable(traces)}`;
@@ -267,7 +268,7 @@ export const listCommand = buildListCommand("trace", {
     // Build footer hint based on result state
     let hint: string | undefined;
     if (traces.length === 0 && hasMore) {
-      hint = `No traces on this page. Try the next page: ${nextPageHint(org, project, flags)}`;
+      hint = `Try the next page: ${nextPageHint(org, project, flags)}`;
     } else if (traces.length > 0) {
       const countText = `Showing ${traces.length} trace${traces.length === 1 ? "" : "s"}.`;
       hint = hasMore


src/commands/log/view.ts
```diff
@@ -376,11 +375,10 @@ export const viewCommand = buildCommand({
...
-    if (flags.json) {
-      writeJson(stdout, logs, flags.fields);
-      return;
-    }
+    const hint = target.detectedFrom
+      ? `Detected from ${target.detectedFrom}`
+      : undefined;
 
-    writeHumanOutput(stdout, logs, target.org, target.detectedFrom);
+    return { data: { logs, orgSlug: target.org }, hint };
```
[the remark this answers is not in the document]
BYK: Fixed in a9e837a1. Added `jsonTransform` to strip `orgSlug` and preserve the original bare-array JSON contract: `jsonTransform: (data, _fields) => data.logs`. Unit tests and E2E test updated accordingly.

src/commands/org/list.ts
```diff
@@ -81,54 +125,29 @@ export const listCommand = buildCommand({
...
-
-    if (orgs.length > flags.limit) {
-      stdout.write(
-        `\nShowing ${flags.limit} of ${orgs.length} organizations\n`
-      );
-    }
+    const hint =
+      orgs.length > flags.limit
+        ? `Showing ${flags.limit} of ${orgs.length} organizations`
+        : "Tip: Use 'sentry org view <slug>' for details";
```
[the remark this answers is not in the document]
BYK: Fixed in a9e837a1. The hint is now conditional on `entries.length > 0` — no tip is shown when the org list is empty.

src/commands/log/view.ts
```diff
@@ -313,7 +316,13 @@ export const viewCommand = buildCommand({
       "within a single argument (handy when piping from other commands).\n\n" +
       "The log ID is the 32-character hexadecimal identifier shown in log listings.",
   },
-  output: "json",
+  output: {
+    json: true,
+    human: formatLogViewHuman,
+    // Preserve original JSON contract: bare array of log entries.
+    // orgSlug exists only for the human formatter (trace URLs).
+    jsonTransform: (data: LogViewData, _fields) => data.logs,
```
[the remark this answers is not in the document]
BYK: Fixed in 8dcba687. The `jsonTransform` now applies `filterFields` per element when `fields` is non-empty, matching the `trace/list` pattern.

src/commands/auth/login.ts
```diff
@@ -119,7 +119,8 @@ export const loginCommand = buildCommand({
       // Non-fatal: cache directory may not exist
     }
 
-    // Device Flow OAuth
+    // Device Flow OAuth — still needs raw stdout/stderr Writers
```
BYK: Why?

src/commands/auth/logout.ts
```diff
@@ -25,24 +27,22 @@ export const logoutCommand = buildCommand({
     flags: {},
   },
   async func(this: SentryContext): Promise<void> {
-    const { stdout } = this;
-
     if (!(await isAuthenticated())) {
-      stdout.write("Not currently authenticated.\n");
+      log.info("Not currently authenticated.");
```
BYK: This is actually an error? Maybe?

src/commands/auth/logout.ts
```diff
@@ -25,24 +27,22 @@ export const logoutCommand = buildCommand({
...
       return;
     }
 
     if (isEnvTokenActive()) {
       const envVar = getActiveEnvVarName();
-      stdout.write(
+      log.info(
```
BYK: Same as above, this is an error?

src/commands/auth/login.ts
BYK: `runInteractiveLogin` takes raw `Writer` params (15+ direct writes for the device flow prompts, QR code, polling status). Converting it is the most complex single-file conversion, so it is deferred to a separate PR. Removed the stale comment — the destructuring is self-explanatory.

src/commands/auth/logout.ts
BYK: Good catch — changed to `log.warn` since these are actionable warnings (you tried to log out but are not authenticated / cannot because of env var). Updated in the latest push.

src/commands/auth/login.ts
BYK: Removed the unnecessary comment in e861376f. The code is self-evident — `runInteractiveLogin` takes Writers as params.

src/commands/auth/logout.ts
BYK: Same — changed to `log.warn` in latest push. Both "not authenticated" and "env var override" are warning-level: the user asked to log out but the action cannot proceed.

src/commands/auth/logout.ts
BYK: Good catch — changed to `log.warn` in e861376f. Both the "not authenticated" and "env token active" messages are now warnings.

src/commands/auth/logout.ts
BYK: Fixed alongside the above — both are `log.warn` now in e861376f.

src/commands/project/view.ts
```diff
@@ -280,35 +286,13 @@ export const viewCommand = buildCommand({
...
+    const entries: ProjectViewEntry[] = projects.map((p, i) => ({
+      ...p,
+      dsn: dsns[i] ?? null,
+      detectedFrom: targets[i]?.detectedFrom,
+    }));
 
-    if (footer) {
-      stdout.write(`\n${footer}\n`);
-    }
+    return { data: entries, hint: footer };
```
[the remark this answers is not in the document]
BYK: Fixed in e861376f. Added `jsonExclude: ["detectedFrom"]` to the `OutputConfig` to strip the human-only field from JSON output.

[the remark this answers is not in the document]
BYK: Already handled — `jsonExclude: ["detectedFrom"]` is set on the `OutputConfig` at line 190. The `renderCommandOutput` function strips excluded keys before JSON serialization. The `detectedFrom` field is only used by `formatProjectViewHuman` for the human-readable "Detected from: ..." line.

src/commands/auth/status.ts
```diff
@@ -61,82 +62,75 @@ function envVarName(source: AuthSource): string {
...
+    log.info("Auto-refresh: enabled");
   } else {
-    stdout.write("Auto-refresh: disabled (no refresh token)\n");
+    log.info("Auto-refresh: disabled (no refresh token)");
   }
 }
 
 /**
- * Write default settings
+ * Log default org/project settings if configured.
```
BYK, line 93: I think this entire output for `auth` commands should be migrated to the new Markdown tables (probably key-value tables)

src/commands/cli/fix.ts
```diff
@@ -464,101 +454,83 @@ type DiagnoseResult = {
...
-    );
-  }
-  stdout.write("\n");
+  const issueList = permIssues
+    .map(
+      (issue) =>
+        `  - ${issue.kind} ${issue.path}: ${formatMode(issue.currentMode)} (expected ${formatMode(issue.expectedMode)})`
+    )
+    .join("\n");
+  log.info(`Found ${permIssues.length} permission issue(s):\n${issueList}`);
```
BYK, line 477: Same with this: should be migrated to markdown rendering pipeline.

src/commands/org/list.ts
```diff
@@ -81,54 +125,32 @@ export const listCommand = buildCommand({
...
+    let hint: string | undefined;
     if (orgs.length > flags.limit) {
-      stdout.write(
-        `\nShowing ${flags.limit} of ${orgs.length} organizations\n`
-      );
+      hint = `Showing ${flags.limit} of ${orgs.length} organizations`;
+    } else if (entries.length > 0) {
+      hint = "Tip: Use 'sentry org view <slug>' for details";
     }
+    // When entries is empty, hint remains undefined — no tip is shown
```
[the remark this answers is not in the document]
BYK: Intentional simplification. When the list is truncated, the truncation notice is more useful than the tip. The hint field on CommandOutput supports one string. Most users have fewer than 25 orgs so the truncated case is rare.

src/commands/auth/status.ts
BYK: Agreed — auth status and cli fix should use the markdown rendering pipeline with key-value tables. This PR focuses on the consola migration (stdout/stderr to logger). Adding the full markdown formatter conversion here would make it too large. Will tackle in a follow-up PR alongside the remaining list command conversions (PR 2).

src/commands/cli/fix.ts
BYK: Same — will migrate cli/fix output to the markdown pipeline in the follow-up PR. These stateful, multi-step outputs (scanning, fixing) are good candidates for structured markdown rendering.

BYK:
All three BugBot LOW findings addressed in 1e6cfd48:

1. **org/list truncated hint**: Changed from `else if` to a `hints[]` array joined with newline — both the truncation notice and the tip footer are now shown when the list is truncated.

2. **cli/setup log shadowing**: Renamed module-level `cmdLog` to `log` (matching every other command) and renamed the local quiet-aware function from `log` to `emit` across all 5 helper function signatures and 22 call sites.

3. **trace/list empty-page hint**: Moved the next-page command back inline with the empty-page message in the hint field, so it reads: "No traces on this page. Try the next page: sentry trace list --cursor last"

PR #397, opened 2026-03-11
feat: add project delete command

MathurAditya724:
## Summary

Add `sentry project delete` subcommand for permanently deleting Sentry projects via the API.

### Changes

- **`src/commands/project/delete.ts`** — New command with safety measures:
  - Requires explicit `<org>/<project>` or `<project>` target (no auto-detect)
  - Interactive confirmation prompt (strict `confirmed !== true` check for Symbol(clack:cancel) gotcha)
  - `--yes`/`-y` flag to skip confirmation for CI/agent usage
  - `--dry-run`/`-n` flag to validate inputs and show what would be deleted without deleting
  - Refuses to run in non-interactive mode without `--yes`
  - Verifies project exists via `getProject()` before prompting
  - 403 errors throw `ApiError` with actionable message (preserves HTTP status for upstream handlers)
- **`src/commands/project/index.ts`** — Registered `delete` in project route map
- **`src/lib/api-client.ts`** — Added `deleteProject()` using `@sentry/api` SDK's `deleteAProject`
- **`src/lib/oauth.ts`** — Added `project:admin` to OAuth scopes (required for project deletion; existing users must re-run `sentry auth login`)
- **`test/commands/project/delete.test.ts`** — 10 unit tests covering happy path, error cases, dry-run, JSON output, and safety checks

### Usage

```
sentry project delete acme-corp/my-app
sentry project delete my-app
sentry project delete acme-corp/my-app --yes
sentry project delete acme-corp/my-app --json --yes
sentry project delete acme-corp/my-app --dry-run
```

### Notes

- Existing tokens do not include `project:admin` — users need to re-authenticate via `sentry auth login` after upgrading
- Local caches (DSN cache, project cache, defaults) are not cleared after deletion — stale entries may persist until TTL expiry (follow-up)


src/lib/oauth.ts
```diff
@@ -51,6 +51,7 @@ function getClientId(): string {
 const SCOPES = [
   "project:read",
   "project:write",
+  "project:admin",
```
betegon, line 55: what would happen if someone doesn't have this permission? would it face trouble signing in?

src/lib/oauth.ts
```diff
@@ -52,6 +52,7 @@ function getClientId(): string {
...
```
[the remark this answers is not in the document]
betegon: not the case.

plugins/sentry-cli/skills/sentry-cli/SKILL.md
```diff
@@ -174,6 +174,16 @@ Create a new project
 - `--json - Output as JSON`
 - `--fields <value> - Comma-separated fields to include in JSON output (dot.notation supported)`
 
+#### `sentry project delete <org/project>`
+
+Delete a project
+
+**Flags:**
+- `-y, --yes - Skip confirmation prompt`
```
BYK, line 188: We should also accept `-f/--force` flag

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  const confirmed = await log.prompt(
+    `Delete project '${project.name}' (${orgSlug}/${project.slug})? This cannot be undone.`,
+    { type: "confirm", initial: false }
+  );
+
+  // consola prompt returns Symbol(clack:cancel) on Ctrl+C — a truthy value.
+  // Strictly check for `true` to avoid deleting on cancel.
+  return confirmed === true;
```
BYK: Let's follow the classical safety measure and ask the user to type out `org/project` like deletions on GitHub UI and other places?

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+async function buildPermissionError(
+  orgSlug: string,
+  projectSlug: string
+): Promise<ApiError> {
+  const label = `'${orgSlug}/${projectSlug}'`;
+  const rolesWithAccess = "Manager, Owner, or Team Admin";
+
+  let orgRole: string | undefined;
+  try {
+    const org = await getOrganization(orgSlug);
```
BYK: If there's a cached version of that, let's make sure to use that

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+      403
+    );
+  }
+
+  if (orgRole && ["manager", "owner", "admin"].includes(orgRole)) {
+    return new ApiError(
+      `Permission denied: You don't have permission to delete ${label}.\n\n` +
+        `Your org role ('${orgRole}') should allow this. ` +
+        "Your auth token may be missing the 'project:admin' scope.\n" +
+        "  Re-authenticate:  sentry auth login",
```
BYK: How would this grant that permission? This can only be the case when they are using a custom oauth token so `sentry auth login` would not help here.

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+      403
+    );
+  }
+
+  return new ApiError(
+    `Permission denied: You don't have permission to delete ${label}.\n\n` +
+      `This requires ${rolesWithAccess} role, or a token with the 'project:admin' scope.\n` +
+      `  Check your role:   sentry org view ${orgSlug}\n` +
```
BYK: Does this command really show the user's role in the org? If not let's make that the case

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+      "  Re-authenticate:   sentry auth login",
```
BYK: Reauth using `sentry auth login` will not resolve any permission issues at all. This could only help when using a custom OAuth token which then they need to regenerate or expand the scope of.

src/commands/project/delete.ts
```diff
@@ -0,0 +1,250 @@
...
+      "Permanently delete a Sentry project. This action cannot be undone.\n\n" +
+      "Requires explicit target — auto-detection is disabled for safety.\n\n" +
+      "Examples:\n" +
+      "  sentry project delete acme-corp/my-app\n" +
+      "  sentry project delete my-app\n" +
+      "  sentry project delete acme-corp/my-app --yes\n" +
+      "  sentry project delete acme-corp/my-app --dry-run",
+  },
+  output: {
+    json: true,
```
BYK: No longer needed (the default)

PR #399, opened 2026-03-12
feat(trial): auto-prompt for Seer trial + sentry trial list/start commands

BYK:
## Summary

When Seer commands (`issue explain`, `issue plan`) fail with a budget or enablement error in an interactive terminal, the CLI now checks for an available Seer product trial, prompts the user, and retries the command on success. Additionally, new `sentry trial list` and `sentry trial start` commands give users proactive control over product trials.

## Auto-prompt flow

Mirrors the existing `executeWithAutoAuth` middleware pattern:

```
main() → executeWithAutoAuth() → executeWithSeerTrialPrompt() → runCommand()
```

1. Seer command fails with `SeerError` (`no_budget` or `not_enabled`)
2. `isTrialEligible()` checks: eligible reason + org slug available + interactive TTY
3. Fetches trial availability via `GET /api/0/customers/{org}/` (control silo)
4. Finds an unstarted Seer trial (prefers `seerUsers`, falls back to `seerAutofix`)
5. Prompts user with consola confirm
6. Starts trial via `PUT /api/0/customers/{org}/product-trial/`
7. Retries the original command through the full middleware chain

Non-eligible errors (`ai_disabled`, missing org slug, non-TTY, API failures) pass through unchanged.

### `sentry trial list [org]`
Lists all product trials for an org with status indicators:
- ○ Available (cyan) — not yet started
- ● Active (green) — started, with days remaining
- − Expired (muted) — ended

Supports `--json` and `--fields` for machine consumption.

### `sentry trial start <name> [org]`
Starts a named product trial. Supported names: `seer`, `replays`, `performance`, `spans`, `profiling`, `logs`.

Features smart argument swap detection — `sentry trial start my-org seer` works the same as `sentry trial start seer my-org`.

## Key design decisions

- **`isTrialEligible()` accepts `unknown`** — does the `instanceof SeerError` check internally so callers don't need to narrow first
- **Control silo routing** — `/customers/` is a billing endpoint, uses `getControlSiloUrl()` not region URLs
- **Consola logger** — all user-facing output uses `log.info()`/`log.warn()`/`log.success()` per project conventions (no raw `stderr.write`)
- **Graceful degradation** — API failures during trial check silently fall through to the normal error display
- **`ai_disabled` excluded** — admin's explicit choice to disable AI; trial wouldn't override it
- **UTC timezone** — `getTrialStatus()` uses `setUTCHours()` for consistent date comparison
- **Soft trial hints in error messages** — `SeerError.format()` says "You may be eligible" + `sentry trial list` (not a direct "start" since trial may not exist)

## New files

| File | Description |
|------|-------------|
| `src/lib/seer-trial.ts` | `isTrialEligible()` + `promptAndStartTrial()` |
| `src/lib/trials.ts` | Trial name mapping, status helpers, `findAvailableTrial()` |
| `src/commands/trial/index.ts` | Route map for trial subcommands |
| `src/commands/trial/list.ts` | `sentry trial list` command |
| `src/commands/trial/start.ts` | `sentry trial start` command |
| `src/lib/arg-parsing.ts` | `detectSwappedTrialArgs()` |

## Modified files

| File | Changes |
|------|---------|
| `src/lib/api-client.ts` | `getProductTrials()` + `startProductTrial()` using control silo |
| `src/lib/errors.ts` | Soft trial hint in `SeerError.format()` |
| `src/bin.ts` | `executeWithSeerTrialPrompt()` middleware |
| `src/app.ts` | Register trial routes + `trials` alias |
| `src/types/sentry.ts` | `ProductTrialSchema`, `CustomerTrialInfoSchema` |

## Tests

- **20 unit tests** for trial prompt flow (`seer-trial.test.ts`)
- **10 unit tests** for trial API client (`api-client.seer-trial.test.ts`)
- **30 unit tests** for trial helpers (`trials.test.ts`)
- **4 property-based tests** for eligibility invariants (`seer-trial.property.test.ts`)
- **9 unit tests** for `trial list` command (`trial/list.test.ts`)
- **10 unit tests** for `trial start` command (`trial/start.test.ts`)
- **4 unit tests** for `detectSwappedTrialArgs` (`arg-parsing.test.ts`)
- 87 trial-related tests total, all passing

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -30,32 +30,38 @@ function handleStreamError(err: NodeJS.ErrnoException): void {
 process.stdout.on("error", handleStreamError);
 process.stderr.on("error", handleStreamError);
 
-/** Run CLI command with telemetry wrapper */
-async function runCommand(args: string[]): Promise<void> {
-  await withTelemetry(async (span) =>
-    run(app, args, buildContext(process, span))
-  );
-}
+/**
+ * Error-recovery middleware for the CLI.
+ *
+ * Each middleware wraps command execution and may intercept specific errors
+ * to perform recovery actions (e.g., login, start trial) then retry.
+ *
+ * Middlewares are applied innermost-first: the last middleware in the array
+ * wraps the outermost layer, so it gets first crack at errors. This means
+ * auth recovery (outermost) can catch errors from both the command AND
+ * the trial prompt retry.
+ *
+ * @param next - The next function in the chain (command or inner middleware)
+ * @param args - CLI arguments for retry
+ * @returns A function with the same signature, with error recovery added
+ */
+type ErrorMiddleware = (
+  next: (argv: string[]) => Promise<void>,
+  args: string[]
+) => Promise<void>;
 
 /**
- * Execute command with automatic Seer trial prompt.
+ * Seer trial prompt middleware.
  *
- * If the command fails with a trial-eligible SeerError in an interactive TTY,
- * checks for available trial, prompts user, starts trial, and retries.
- *
- * Shows a brief context message (not the full error format with URLs) before
- * the trial prompt. If the trial isn't available or the user declines, the
- * full error is re-thrown so the outer handler in main() displays it normally.
- *
- * @throws Re-throws the original error when trial is unavailable or declined
+ * Catches trial-eligible SeerErrors and offers to start a free trial.
+ * On success, retries the original command. On failure/decline, re-throws
+ * so the outer error handler displays the full error with upgrade URL.
  */
-async function executeWithSeerTrialPrompt(args: string[]): Promise<void> {
+const seerTrialMiddleware: ErrorMiddleware = async (next, args) => {
   try {
-    await runCommand(args);
+    await next(args);
   } catch (err) {
-    // isTrialEligible handles instanceof SeerError check + reason + orgSlug + TTY
     if (isTrialEligible(err)) {
-      // isTrialEligible narrows err to SeerError with defined orgSlug
       const started = await promptAndStartTrial(
         // biome-ignore lint/style/noNonNullAssertion: isTrialEligible guarantees orgSlug is defined
         err.orgSlug!,
@@ -64,31 +70,25 @@ async function executeWithSeerTrialPrompt(args: string[]): Promise<void> {
 
       if (started) {
         process.stderr.write("\nRetrying command...\n\n");
-        await runCommand(args);
+        await next(args);
         return;
       }
-
-      // Trial not started (unavailable, declined, or failed) — re-throw
-      // so the outer error handler in main() displays the full error
-      // with the upgrade/settings URL
     }
     throw err;
   }
-}
+};
 
 /**
- * Execute command with automatic authentication.
+ * Auto-authentication middleware.
  *
- * If the command fails due to missing authentication and we're in a TTY,
- * automatically run the interactive login flow and retry the command.
- *
- * @throws Re-throws any non-authentication errors from the command
+ * Catches auth errors (not_authenticated, expired) in interactive TTYs
+ * and runs the login flow. On success, retries through the full middleware
+ * chain so inner middlewares (e.g., trial prompt) also apply to the retry.
  */
-async function executeWithAutoAuth(args: string[]): Promise<void> {
+const autoAuthMiddleware: ErrorMiddleware = async (next, args) => {
   try {
-    await executeWithSeerTrialPrompt(args);
+    await next(args);
   } catch (err) {
-    // Auto-login for auth errors in interactive TTY environments
     // Use isatty(0) for reliable stdin TTY detection (process.stdin.isTTY can be undefined in Bun)
     // Errors can opt-out via skipAutoAuth (e.g., auth status command)
     if (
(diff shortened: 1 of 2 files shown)


src/lib/seer-trial.ts
```diff
@@ -0,0 +1,112 @@
...
+  );
+
+  // Symbol(clack:cancel) is truthy — strict equality check
+  if (confirmed !== true) {
+    return false;
+  }
+
+  // 3. Start trial using the category from the available trial
+  try {
+    stderr.write("\nStarting Seer trial...\n");
```
BYK: We should not be using `stderr` or `stdout` directly for these. See the recent patches on `main`: either use `log.info()` etc, or use the return-based output system with markdown.

src/lib/seer-trial.ts
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    trial = await getSeerTrialStatus(orgSlug);
+  } catch {
+    // Can't check trial status — degrade gracefully
+    return false;
+  }
+
+  if (!trial) {
+    // No trial available — fall through to normal error
+    return false;
```
BYK, line 92: I'd say we should tell them their trial has expired/finished and encourage them to buy it if they like it? :)

src/lib/seer-trial.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  // 3. Start trial using the category from the available trial
+  try {
+    stderr.write("\nStarting Seer trial...\n");
+    await startSeerTrial(orgSlug, trial.category);
+    stderr.write(`${success("✓")} Seer trial activated!\n`);
+    return true;
+  } catch {
+    stderr.write(
+      "Failed to start trial. Please try again or visit your Sentry settings.\n"
```
BYK: A link to their Sentry settings would be very useful here. Also a brief mention of contacting support may be?

src/bin.ts
```diff
@@ -36,6 +42,45 @@ async function runCommand(args: string[]): Promise<void> {
...
+ * the trial prompt. If the trial isn't available or the user declines, the
+ * full error is re-thrown so the outer handler in main() displays it normally.
+ *
+ * @throws Re-throws the original error when trial is unavailable or declined
+ */
+async function executeWithSeerTrialPrompt(args: string[]): Promise<void> {
+  try {
+    await runCommand(args);
+  } catch (err) {
+    if (err instanceof SeerError && isTrialEligible(err)) {
```
BYK: Why not move the `instanceof` check into `isTrialEligible()` too?

src/bin.ts
```diff
@@ -46,7 +91,7 @@ async function runCommand(args: string[]): Promise<void> {
  */
 async function executeWithAutoAuth(args: string[]): Promise<void> {
   try {
-    await runCommand(args);
+    await executeWithSeerTrialPrompt(args);
```
BYK: Looks like we are re-inventing middleware for the CLI? Does Stricli have anything built-in for this? If not should we look into a proper middleware solution?

src/lib/seer-trial.ts
BYK: Done — replaced all `stderr.write()` with `log.info()`/`log.warn()`/`log.success()` via `logger.withTag("seer")`. Removed the `stderr` parameter entirely.

src/lib/seer-trial.ts
BYK: Good call! The code now shows an explicit message when no trial is available: "No Seer trial available. If you've already used your trial, consider upgrading your plan to continue using Seer" followed by the billing URL.

src/lib/seer-trial.ts
BYK: Added both! The billing URL is now included in the failure message, and I added a `support@sentry.io` contact mention as a fallback.

src/bin.ts
BYK: Done — `isTrialEligible()` now accepts `unknown` and does the `instanceof SeerError` check internally. The caller in bin.ts just passes the raw caught error.

src/bin.ts
BYK: Good question! Stricli doesn't have built-in middleware support — it focuses on command definition and argument parsing. The pattern here follows the existing `executeWithAutoAuth()` wrapper that was already in `bin.ts`, so it's consistent with what we already do for auth error recovery.

A proper middleware abstraction could be nice as we add more of these, but for now with just 2 wrappers (auth retry + trial prompt), the nesting is manageable. Happy to extract a generic middleware runner if we add a third one.

test/lib/seer-trial.test.ts
```diff
@@ -0,0 +1,293 @@
...
+    startProductTrialSpy.mockRejectedValue(new Error("API error"));
+
+    const result = await promptAndStartTrial("test-org", "no_budget");
+
+    expect(result).toBe(false);
+    expect(logWarnCalls.some((m) => m.includes("Failed to start trial"))).toBe(
+      true
+    );
+    // Should include a link to billing/settings
+    expect(logWarnCalls.some((m) => m.includes("sentry.io"))).toBe(true);
```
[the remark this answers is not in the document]
BYK: False positive — this is a test assertion checking that log output contains `sentry.io` as part of the expected billing URL. This is not URL validation or sanitization; it's verifying that `buildBillingUrl()` output appears in the logged warnings.

src/lib/trials.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/**
+ * Check if a string is a known CLI-friendly trial name.
+ *
+ * @param value - String to check
+ * @returns true if it's a valid trial name
+ */
+export function isTrialName(value: string): boolean {
+  return value in TRIAL_NAMES;
```
BYK: Might be better to make `TRIAL_NAMES` a `Map()` or use `Object.keys().includes()` here as `in` check can return true for built-in object properties (unless we use Object.create(null) I guess?)

src/bin.ts
BYK: It would be confusing when someone wants to add a third middleware. I'd just make this an array-based middleware system (simple) for now to give structure for future builders.

BYK: Done — refactored to an array-based `ErrorMiddleware` system in 7a6e13c9. New middlewares just get appended to the `errorMiddlewares` array, and `buildExecutor()` composes them automatically.

src/lib/trials.ts
BYK: Good catch — changed to `Object.hasOwn(TRIAL_NAMES, value)` in 7a6e13c9.

src/lib/trials.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (!(trial.isStarted && trial.endDate)) {
+    return null;
+  }
+
+  const end = new Date(trial.endDate);
+  const now = new Date();
+  const diffMs = end.getTime() - now.getTime();
+  const diffDays = Math.ceil(diffMs / (1000 * 60 * 60 * 24));
+  return Math.max(0, diffDays);
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 7a6e13c9 — `getDaysRemaining` now applies `end.setUTCHours(23, 59, 59, 999)` matching the same end-of-day UTC treatment used in `getTrialStatus`.

test/lib/seer-trial.test.ts
```diff
@@ -0,0 +1,297 @@
...
```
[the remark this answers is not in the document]
BYK: False positive — same as the other CodeQL alert. This is a test assertion verifying that log output contains the expected support email address, not URL sanitization. The assertion has been narrowed to check for `support@sentry` (without the TLD) in b9647b5a to avoid triggering this rule.

PR #408, opened 2026-03-12
docs: update credential storage docs and remove stale config.json references

betegon:
## Summary

Auth credentials were migrated to SQLite (`cli.db`) but docs and code comments still referenced `~/.sentry/config.json`. This updates everything to match reality:

- Rewrites the auth command docs "Credential Storage" section to note that credentials live in `cli.db` with restricted permissions, and points users to `sentry auth token` / `sentry auth status` instead of exposing the internal schema
- Updates the configuration page's credential storage section to name `cli.db`, mention WAL permissions, and link to the auth docs
- Removes DB schema details (column/type table), `sqlite3` query examples, and "Direct Database Access" subsection — users shouldn't read the DB directly
- Removes sqlite3 examples from the SKILL.md agent skill file
- Fixes three stale comments/strings in `logout.ts`, `oauth.ts`, and `dsn/types.ts`

## Test Plan

- `bun run typecheck` — passes (pre-existing trial module errors only)
- `bun run lint` — clean
- `bun run test:unit` — no new failures
- Verified no `sqlite3` or schema content remains in docs

### DIFF
--- a/docs/src/content/docs/commands/auth.md
+++ b/docs/src/content/docs/commands/auth.md
@@ -94,8 +94,8 @@ Use `sentry auth token` to retrieve your current access token, or `sentry auth s
 
 The CLI checks for auth tokens in the following order, using the first one found:
 
-1. `SENTRY_AUTH_TOKEN` environment variable
-2. `SENTRY_TOKEN` environment variable (legacy)
+1. `SENTRY_AUTH_TOKEN` environment variable (legacy)
+2. `SENTRY_TOKEN` environment variable
 3. The stored token in the SQLite database
 
 When a token comes from an environment variable, the CLI skips expiry checks and automatic refresh.


BYK (changes requested):
We should not be documenting our schema or encouraging people to read the db directly. We already have `sentry auth token` command for this

docs/src/content/docs/commands/auth.md
```diff
@@ -84,18 +84,18 @@ sentry auth refresh
...
-    "expiresAt": "2024-12-31T00:00:00Z"
-  }
-}
-```
+### Environment Variable Precedence
+
+The CLI checks for auth tokens in the following order, using the first one found:
+
+1. `SENTRY_AUTH_TOKEN` environment variable
+2. `SENTRY_TOKEN` environment variable (legacy)
```
BYK: It's the other way around, SENTRY_TOKEN is the new one, inspired by GITHUB_TOKEN

PR #409, opened 2026-03-12
feat: support SENTRY_HOST as alias for SENTRY_URL

betegon:
## Summary

Adds `SENTRY_HOST` as a supported env var for configuring the Sentry instance URL. It takes precedence over `SENTRY_URL` so both work, but users who expect `SENTRY_HOST` (common in other Sentry tooling) get the behavior they want out of the box.

Precedence: `SENTRY_HOST` > `SENTRY_URL` > default (`https://sentry.io`)

## Changes

A single `getConfiguredSentryUrl()` helper in `constants.ts` replaces all 8 `process.env.SENTRY_URL` reads across 6 source files. `applySentryUrlContext()` now sets/deletes both env vars so `SENTRY_HOST` can't shadow a freshly-set `SENTRY_URL`.

## Test plan

- New tests for `SENTRY_HOST` alone and precedence over `SENTRY_URL` in region, code-scanner, URL parser, and URL builder tests
- All 146 tests across the 4 modified test files pass
- `bun run typecheck` and `bun run lint` clean

### DIFF
--- a/test/commands/event/view.test.ts
+++ b/test/commands/event/view.test.ts
@@ -158,13 +158,16 @@ describe("parsePositionalArgs", () => {
     });
   });
 
-  // URL integration tests — applySentryUrlContext may set SENTRY_URL as a side effect
+  // URL integration tests — applySentryUrlContext may set SENTRY_HOST/SENTRY_URL as a side effect
   describe("Sentry URL inputs", () => {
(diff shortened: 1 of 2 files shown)


betegon:
@BYK no lol, i agree with you. but mcp was there first 😆

PR #416, opened 2026-03-12
refactor: unify commands as generators with HumanRenderer factory, remove stdout plumbing

BYK:
## Summary

Unifies all command functions as async generators, introduces a `HumanRenderer` factory for stateful rendering, and removes direct stdout/stderr plumbing from commands. This is Phase 6b of the output convergence plan.

### 1. Unify all command functions as async generators

Every command `func` is now an `async *func` generator. Commands yield `CommandOutput<T>` values instead of writing to stdout directly.

- Commands that produce output yield via `commandOutput(data)`
- Void generators (e.g. `auth/token`) are valid no-op generators
- The `buildCommand` wrapper iterates the generator and routes output through the `OutputConfig` pipeline

### 2. Brand `CommandOutput` with Symbol discriminant

Replace duck-typing (`"data" in value`) with a `COMMAND_OUTPUT_BRAND` Symbol. Prevents false positives from raw API responses that have a `data` property.

- `commandOutput<T>(data)` factory creates branded values
- `isCommandOutput()` checks the Symbol instead of structural shape
- All command files migrated to use `commandOutput()`

### 3. Move hints from yield to generator return value

Hints (footer text like "Detected from .env.local") are no longer part of the yielded `CommandOutput`. Generators `return { hint }` after their final yield.

- New `CommandReturn` type: `{ hint?: string }`
- Wrapper uses manual `.next()` iteration to capture the return value
- Hints are suppressed in JSON mode

### 4. `HumanRenderer` factory pattern for `OutputConfig.human`

`OutputConfig.human` is now a factory function `() => HumanRenderer<T>` instead of a plain `(data: T) => string`. This enables stateful rendering (e.g., streaming tables) without module-level singletons.

- `HumanRenderer<T>` has `render(data: T): string` and optional `finalize(hint?: string): string`
- `stateless(fn)` helper wraps plain formatters for commands that don't need state
- `createLogRenderer()` replaces the old module-level `streamingTable`/`streamingHeaderEmitted` singletons in `log/list.ts`
- The `buildCommand` wrapper resolves the factory once per invocation, passes the renderer through iteration, and calls `finalize()` after the generator completes

### 5. Remove stdout/stderr plumbing from commands

Commands no longer receive or pass `stdout`/`stderr` Writer references. All diagnostic/interactive output routes through consola logger (→ stderr), keeping stdout reserved for structured command output.

- `HandlerContext` and `DispatchOptions` no longer carry `stdout`
- `runInteractiveLogin` uses logger instead of Writer params
- Follow-mode banners, diagnostics, and QR code display all use logger
- `displayTraceLogs` → `formatTraceLogs` (returns string)
- `formatFooter()` helper extracted from `writeFooter()`

### 6. `log/list.ts` streaming refactor

The most complex command — 685 lines with follow-mode streaming, trace filtering, and JSONL output — is fully converted:

- `createLogRenderer()` factory manages streaming table state per-invocation
- `render()` emits table header on first non-empty batch, rows per batch
- `finalize(hint?)` closes the table footer and appends hint text
- Empty-state hints render as primary text (not muted footer)
- `jsonl` flag on `LogListResult` controls JSONL vs array serialization
- `executeSingleFetch`/`executeTraceSingleFetch` return `FetchResult` with separate `result` and `hint` fields

### 7. `auth/login` yield pattern

Interactive OAuth login now yields `LoginResult` values through the generator instead of writing to stdout/stderr directly:

- Token-based login: yields success result, returns hint
- OAuth flow: yields QR code + URL, progress dots, and final result
- `runInteractiveLogin` returns `LoginResult` values consumed by the generator

**Remaining stdout usage (intentional):**
- `auth/token`: raw stdout for pipe compatibility (`sentry auth token | pbcopy`)
- `help.ts`: help text to stdout (like `git --help`)
- `trace/logs.ts`: `process.stdout.write` (pending OutputConfig migration)

## Test plan

- 1656 tests pass, 0 fail, 15753 assertions across 55 files
- Typecheck clean, lint clean (375 files)
- E2E tests pass (103 pass, 3 skip)
- All CI checks green including Seer Code Review and Cursor BugBot

### DIFF
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -19,7 +19,7 @@ export const helpCommand = buildCommand({
       "Display help information. Run 'sentry help' for an overview, " +
       "or 'sentry help <command>' for detailed help on a specific command.",
   },
-  output: { human: (s: string) => s },
+  output: { human: (s: string) => s.trimEnd() },
   parameters: {
     flags: {},
     positional: {
--- a/src/commands/trace/logs.ts
+++ b/src/commands/trace/logs.ts
@@ -59,7 +59,7 @@ function formatTraceLogsHuman(data: TraceLogsData): string {
   const countText = `Showing ${data.logs.length} log${data.logs.length === 1 ? "" : "s"} for trace ${data.traceId}.`;
   const tip = hasMore ? " Use --limit to show more." : "";
   parts.push(formatFooter(`${countText}${tip}`));
-  return parts.join("");
+  return parts.join("").trimEnd();
 }
 
 /** Maximum allowed value for --limit flag */


src/commands/auth/login.ts
```diff
@@ -12,13 +12,38 @@ import {
...
+  );
+  if (result.user) {
+    lines.push(`  Logged in as: ${formatUserIdentity(result.user)}`);
+  }
+  lines.push(`  Config saved to: ${result.configPath}`);
+  if (result.expiresIn) {
+    lines.push(`  Token expires in: ${formatDuration(result.expiresIn)}`);
+  }
+  lines.push(""); // trailing newline
+  return lines.join("\n");
```
BYK, line 44: Should we not make this part a table?

src/commands/auth/login.ts
```diff
@@ -139,46 +165,41 @@ export const loginCommand = buildCommand({
...
           email: user.email,
           username: user.username,
           name: user.name,
         });
+        result.user = {
+          name: user.name,
+          email: user.email,
+          username: user.username,
+          id: user.id,
+        };
```
BYK: Why not just `result.user = user`?

src/commands/auth/login.ts
```diff
@@ -139,46 +165,41 @@ export const loginCommand = buildCommand({
...
-    const loginSuccess = await runInteractiveLogin(
-      stdout,
-      stderr,
-      process.stdin,
-      {
-        timeout: flags.timeout * 1000,
-      }
-    );
+    // OAuth device flow
+    const result = await runInteractiveLogin(process.stdin, {
```
BYK: `process.stdin` is a global so no chance of _not_ passing it in as an argument? Guessing this is for testability but I was hoping we could somehow mock this? Especially if we were using Consola's `prompt` helper?

src/commands/auth/logout.ts
```diff
@@ -32,15 +33,17 @@ export const logoutCommand = buildCommand({
...
+  async *func(this: SentryContext) {
     if (!(await isAuthenticated())) {
-      return {
-        data: { loggedOut: false, message: "Not currently authenticated." },
-      };
+      yield commandOutput({
+        loggedOut: false,
+        message: "Not currently authenticated.",
+      });
+      return;
```
BYK: Curious if we can do `return yield ...` for a one liner?

src/commands/cli/upgrade.ts
```diff
@@ -417,7 +418,7 @@ export const upgradeCommand = buildCommand({
       "  sentry cli upgrade --force      # Force re-download even if up to date\n" +
       "  sentry cli upgrade --method npm # Force using npm to upgrade",
   },
-  output: { json: true, human: formatUpgradeResult },
+  output: { json: true, human: stateless(formatUpgradeResult) },
```
BYK: I wonder if we can drop `json` or at least default it to `true` as our goal is for _all_ commands to have `json` output, no matter what.

src/commands/trace/logs.ts
```diff
@@ -209,16 +206,17 @@ export const logsCommand = buildCommand({
...
-      logs,
-      traceId,
-      limit: flags.limit,
-      asJson: flags.json,
-      fields: flags.fields,
-      emptyMessage:
-        `No logs found for trace ${traceId} in the last ${flags.period}.\n\n` +
-        `Try a longer period: sentry trace logs --period 30d ${traceId}\n`,
-    });
+    process.stdout.write(
```
BYK: What? This should be using `yield commandOutput()`, right?

src/commands/trial/start.ts
```diff
@@ -160,38 +163,36 @@ export const startCommand = buildCommand({
...
+async function promptOpenBillingUrl(url: string): Promise<boolean> {
   const log = logger.withTag("trial");
 
-  stdout.write(`\n  ${url}\n\n`);
+  log.log(`\n  ${url}\n`);
 
   // Show QR code so mobile/remote users can scan
   const qr = await generateQRCode(url);
-  stdout.write(`${qr}\n`);
+  log.log(qr);
```
BYK: No no no, these should be `yield commandOutput()`

src/commands/trial/start.ts
```diff
@@ -160,38 +163,36 @@ export const startCommand = buildCommand({
...
+ * structured command output.
+ *
  * @returns true if browser was opened, false otherwise
  */
-async function promptOpenBillingUrl(
-  url: string,
-  stdout: { write: (s: string) => unknown }
-): Promise<boolean> {
+async function promptOpenBillingUrl(url: string): Promise<boolean> {
   const log = logger.withTag("trial");
```
BYK: I think we should rename this `const log` to `const logger` and import `logger` under a different name

src/commands/help.ts
```diff
@@ -30,12 +30,11 @@ export const helpCommand = buildCommand({
...
   // biome-ignore lint/complexity/noBannedTypes: Stricli requires empty object for commands with no flags
-  async func(this: SentryContext, _flags: {}, ...commandPath: string[]) {
-    const { stdout } = this;
-
+  // biome-ignore lint/correctness/useYield: void generator — delegates to Stricli help system
+  async *func(this: SentryContext, _flags: {}, ...commandPath: string[]) {
     // No args: show branded help
     if (commandPath.length === 0) {
-      await printCustomHelp(stdout);
+      process.stdout.write(await printCustomHelp());
```
BYK: No `process.stdout.write` calls in commands. Use `yield commandOutput()`

src/commands/init.ts
BYK, line 79: We should be `yield`ing from `runWizard()`?

src/lib/formatters/output.ts
```diff
@@ -105,99 +162,227 @@ export type OutputConfig<T> = {
...
 
 /**
- * Return type for commands with {@link OutputConfig}.
+ * Unique brand for {@link CommandOutput} objects.
+ *
+ * Using a Symbol instead of duck-typing (`"data" in v`) prevents false
+ * positives when a command accidentally yields a raw API response that
+ * happens to have a `data` property.
+ */
+export const COMMAND_OUTPUT_BRAND: unique symbol = Symbol.for(
```
BYK: Just a thought: would using classes and `instance of` checks be better? Or just a strict `type`/`class` so we can rely on type checking?

src/lib/formatters/output.ts
```diff
@@ -105,99 +162,227 @@ export type OutputConfig<T> = {
...
- * Called by the `buildCommand` wrapper when a command with `output: { ... }`
- * returns data. In JSON mode the data is serialized as-is (with optional
- * field filtering); in human mode the config's `human` formatter is called.
+ * Use this in `jsonTransform` when a streaming command yields batches
+ * that should be expanded to one line per item.
+ */
+type JsonlLines = {
+  readonly [JSONL_BRAND]: true;
+  readonly items: readonly unknown[];
+};
```
BYK: Why do we need this? We should just be able to `yield` each item instead of grouping them together??

src/lib/clipboard.ts
```diff
@@ -100,8 +101,7 @@ export function setupCopyKeyListener(
       const text = getText();
       const copied = await copyToClipboard(text);
       if (copied && active) {
-        stdout.write(CLEAR_LINE);
-        stdout.write(success("Copied!"));
+        log.success("Copied!");
```
BYK, line 104: Why did we have the `CLEAR_LINE` thing and why is it gone now? 

Also, not sure `log.success()` is the best way to output this?

src/lib/command.ts
```diff
@@ -366,9 +388,39 @@ export function buildCommand<
...
+  function writeFinalization(
+    stdout: Writer,
+    hint: string | undefined,
+    json: unknown,
+    // biome-ignore lint/suspicious/noExplicitAny: Renderer type mirrors erased OutputConfig<T>
+    renderer?: HumanRenderer<any>
+  ): void {
+    if (json) {
+      return;
+    }
```
BYK, line 368: Shall we output the `hint` in JSON outputs too?

src/lib/interactive-login.ts
```diff
@@ -67,39 +79,35 @@ export async function runInteractiveLogin(
...
-            stdout.write("Opening in browser...\n\n");
+            log.info("Opening in browser...");
           } else {
             // Show QR code as fallback when browser can't open
-            stdout.write("Scan this QR code or visit the URL below:\n\n");
+            log.info("Scan this QR code or visit the URL below:");
             const qr = await generateQRCode(verificationUriComplete);
-            stdout.write(qr);
-            stdout.write("\n");
+            log.log(qr);
```
BYK, line 86: I don't think these should be `log.info` and instead should be `yield commandOutput()`

src/lib/interactive-login.ts
```diff
@@ -67,39 +79,35 @@ export async function runInteractiveLogin(
...
-          stdout.write(`URL: ${verificationUri}\n`);
-          stdout.write(`Code: ${userCode}\n\n`);
+          log.info(`URL: ${verificationUri}`);
+          log.info(`Code: ${userCode}`);
           const copyHint = stdin.isTTY ? ` ${muted("(c to copy)")}` : "";
-          stdout.write(
-            `Browser didn't open? Use the url above to sign in${copyHint}\n\n`
+          log.info(
+            `Browser didn't open? Use the url above to sign in${copyHint}`
           );
```
BYK, line 95: Same as above

src/lib/interactive-login.ts
```diff
@@ -67,39 +79,35 @@ export async function runInteractiveLogin(
...
-            stdin,
-            () => urlToCopy,
-            stdout
-          );
+          keyListener.cleanup = setupCopyKeyListener(stdin, () => urlToCopy);
         },
         onPolling: () => {
-          stdout.write(".");
+          // Dots append on the same line without newlines — logger can't do this
+          process.stderr.write(".");
```
BYK: Yeah that's why it should be a spinner or `yield`. Spinner is probably better at this point?

src/commands/log/list.ts
```diff
@@ -49,13 +53,23 @@ type ListFlags = {
...
-  /** Human-readable hint (e.g., "Showing 100 logs. Use --limit to show more.") */
-  hint?: string;
   /** Trace ID, present for trace-filtered queries */
   traceId?: string;
+  /**
+   * When true, JSON output uses JSONL (one object per line) instead
+   * of a JSON array. Set for follow-mode batches where output is
+   * consumed incrementally. Does not affect human rendering.
+   */
+  jsonl?: boolean;
```
BYK: I don't think we need this at all. We should just `yield` each log item one by one for following mode (JSONL natively) and `yield` an array of log items once for JSON mode.

src/commands/log/list.ts
```diff
@@ -478,25 +476,22 @@ function createLogRenderer(): HumanRenderer<LogListResult> {
...
-    return result.jsonl ? undefined : [];
+    return;
   }
 
-  const mapped = result.logs.map(applyFields);
-  return result.jsonl ? jsonlLines(mapped) : mapped;
+  const applyFields = (log: LogLike) =>
+    fields && fields.length > 0 ? filterFields(log, fields) : log;
+
+  return result.logs.map(applyFields);
```
BYK: You can inline `applyFields` here?

src/lib/formatters/output.ts
```diff
@@ -161,51 +164,39 @@ export type OutputConfig<T> = {
...
  *
  * @example
  * ```ts
  * yield commandOutput(myData);
  * ```
  */
 export function commandOutput<T>(data: T): CommandOutput<T> {
-  return { [COMMAND_OUTPUT_BRAND]: true, data };
+  return new CommandOutput(data);
 }
```
BYK: We definitely don't need a helper to do `new CommandOutput()`

src/lib/formatters/output.ts
```diff
@@ -122,8 +122,11 @@ export function stateless<T>(fn: (data: T) => string): () => HumanRenderer<T> {
  *   and serialized as-is to JSON)
  */
 export type OutputConfig<T> = {
-  /** Enable `--json` and `--fields` flag injection */
-  json: true;
+  /**
+   * Enable `--json` and `--fields` flag injection.
+   * Defaults to `true` — can be omitted for brevity.
+   */
+  json?: true;
```
BYK: I didn't see you updating commands that set `json: true` everywhere. Also, if there's nothing else setting `json: false` it's better to remove this altogether

src/lib/command.ts
```diff
@@ -321,19 +320,14 @@ export function buildCommand<
...
   /**
-   * Check if a value is a branded {@link CommandOutput} object.
+   * Check if a value is a {@link CommandOutput} instance.
    *
-   * Uses the {@link COMMAND_OUTPUT_BRAND} Symbol instead of duck-typing
-   * on `"data" in v`, preventing false positives from raw API responses
-   * or other objects that happen to have a `data` property.
+   * Uses `instanceof` instead of duck-typing on `"data" in v`,
+   * preventing false positives from raw API responses or other objects
+   * that happen to have a `data` property.
```
BYK: Not sure if the comment really adds value

src/lib/interactive-login.ts
```diff
@@ -89,25 +88,32 @@ export async function runInteractiveLogin(
...
+              stdin as NodeJS.ReadStream & { fd: 0 },
+              () => urlToCopy
+            );
+          }
         },
         onPolling: () => {
-          // Dots append on the same line without newlines — logger can't do this
-          process.stderr.write(".");
+          // Spinner handles the visual feedback — no-op here
         },
```
BYK: Shall we remove the `onPolling` callback altogether then?

src/lib/command.ts
```diff
@@ -319,13 +307,6 @@ export function buildCommand<
...
-  /**
-   * Check if a value is a {@link CommandOutput} instance.
-   *
-   * Uses `instanceof` instead of duck-typing on `"data" in v`,
-   * preventing false positives from raw API responses or other objects
-   * that happen to have a `data` property.
-   */
   function isCommandOutput(v: unknown): v is CommandOutput<unknown> {
     return v instanceof CommandOutput;
   }
```
BYK: Don't think we need this helper?

src/lib/command.ts
```diff
@@ -281,11 +271,9 @@ export function buildCommand<
   builderArgs: LocalCommandBuilderArguments<FLAGS, ARGS, CONTEXT>
 ): Command<CONTEXT> {
   const originalFunc = builderArgs.func;
-  const rawOutput = builderArgs.output;
-  /** Resolved output config (object form), or undefined if no auto-rendering */
-  const outputConfig = typeof rawOutput === "object" ? rawOutput : undefined;
+  const outputConfig = builderArgs.output;
   /** Whether to inject --json/--fields flags */
-  const hasJsonOutput = rawOutput === "json" || typeof rawOutput === "object";
+  const hasJsonOutput = outputConfig !== undefined;
```
BYK: Do we have _any_ command that doesn't define an `outputConfig`? If not, we can get rid of this check and `human` optionality.

src/lib/interactive-login.ts
```diff
@@ -105,9 +105,6 @@ export async function runInteractiveLogin(
             );
           }
         },
-        onPolling: () => {
```
BYK, line 94: We should also remove this from `performDeviceFlow` then?

src/commands/auth/login.ts
```diff
@@ -182,7 +182,7 @@ export const loginCommand = buildCommand({
         // Non-fatal: user info is supplementary. Token remains stored and valid.
       }
 
-      yield commandOutput(result);
+      yield new CommandOutput(result);
       return;
```
BYK: Let's make sure all code paths doing `yield new CommandOutput(..); return;` become `return yield new CommandOutput(..);`

src/commands/cli/setup.ts
```diff
@@ -56,6 +54,31 @@ type SetupFlags = {
...
+  version: string;
+};
+
+/** Format setup result for human-readable output */
+function formatSetupResult(result: SetupResult): string {
+  const lines: string[] = [...result.messages];
+  if (result.warnings.length > 0) {
+    for (const w of result.warnings) {
+      lines.push(`⚠ ${w}`);
+    }
```
BYK: Why are we not using `logger.warn()` for the warnings??

src/lib/formatters/output.ts
```diff
@@ -58,25 +58,72 @@ type WriteOutputOptions<T> = {
...
- *    AND auto-renders the command's return value. Commands return
- *    `{ data }` or `{ data, hint }` objects.
+ * @example
+ * ```ts
+ * output: {
+ *   human: stateless(formatMyData),
+ * }
+ * ```
+ */
+export function stateless<T>(fn: (data: T) => string): () => HumanRenderer<T> {
```
BYK: Do you think we can somehow get rid of this `stateless` helper and just use `render: ourOldRenderFunction`? 	would that make things worse?

src/commands/cli/setup.ts
```diff
@@ -441,12 +457,16 @@ export const setupCommand = buildCommand({
...
+  async *func(this: SentryContext, flags: SetupFlags) {
     const { process, homeDir } = this;
 
+    const log = logger.withTag("cli.setup");
+    const messages: string[] = [];
+
     const emit: Logger = (msg: string) => {
       if (!flags.quiet) {
-        log.info(msg);
+        messages.push(msg);
```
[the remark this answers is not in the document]
BYK: We should just `yield` each message as they come instead of `push`ing them to an array.

src/lib/formatters/output.ts
```diff
@@ -58,25 +58,81 @@ type WriteOutputOptions<T> = {
...
+ * Disambiguation: a function with `.length === 0` is treated as a factory.
+ */
+export function resolveRenderer<T>(human: HumanOutput<T>): HumanRenderer<T> {
+  // Factory: zero-arg function that returns a renderer
+  if (human.length === 0) {
+    return (human as () => HumanRenderer<T>)();
+  }
+  // Plain formatter: wrap in a stateless renderer
+  return { render: human as (data: T) => string };
+}
```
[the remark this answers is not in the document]
BYK: This is fine.

PR #420, opened 2026-03-13
fix(dsn): make code scanner monorepo-aware and extend --fresh to bypass DSN cache

betegon:
## Summary

The DSN code scanner missed files inside monorepo packages because `MAX_SCAN_DEPTH=2` couldn't reach past `packages/<name>/src/`. For example, in the spotlight monorepo only 1 of 3 DSNs was detected — `packages/website/sentry.client.config.mjs` at depth 2 was found, but `packages/spotlight/src/instrument.ts` at depth 3 was not.

Even after fixing the scanner, stale cached results survive because mtime validation only checks files that were *previously found* — files missed by the old depth limit have no cache entries. Users had no way to force a re-scan.

## Changes

**Scanner fix:**
- Bump `MAX_SCAN_DEPTH` from 2 to 3 (also helps non-monorepo projects with deeper configs like `src/lib/config/sentry.ts`)
- Reset depth to 0 when entering a monorepo package directory (`packages/*`, `apps/*`, `libs/*`, `services/*`, `modules/*`), giving each package its own depth-3 scanning budget
- Uses the existing `MONOREPO_ROOTS` constant from `types.ts`, matching how the env file scanner already handles monorepos

**Cache bypass (`--fresh`):**
- Add `disableDsnCache()`/`enableDsnCache()` following the same pattern as `disableResponseCache()`
- Wire into `applyFreshFlag()` so `--fresh` now bypasses both HTTP response cache and DSN detection cache
- Cache writes still proceed when disabled, so the re-scanned result gets stored for next time

## Test plan

- [x] Added test: finds DSNs in monorepo packages deeper than MAX_SCAN_DEPTH
- [x] Added test: finds DSNs from multiple monorepo packages with correct `packagePath`
- [x] Added tests: `getCachedDsn`/`getCachedDetection` return undefined when cache disabled, writes persist
- [x] All code-scanner and dsn-cache tests pass
- [x] Typecheck passes
- [x] Lint passes
- [ ] Manual test: run `sentry issue list --fresh` from spotlight monorepo, verify multiple projects detected

---

### DIFF
--- a/test/lib/dsn/code-scanner.test.ts
+++ b/test/lib/dsn/code-scanner.test.ts
@@ -462,6 +462,40 @@ describe("Code Scanner", () => {
       expect(beResult?.packagePath).toBe("packages/backend");
     });
 
+    test("finds DSNs deeply nested in monorepo packages", async () => {
+      // packages/spotlight/src/electron/main/index.ts is depth 5 from root,
+      // but after monorepo reset at packages/spotlight/, it's depth 3 —
+      // exactly at MAX_SCAN_DEPTH. This was a specific failing case.
+      mkdirSync(join(testDir, "packages/spotlight/src/electron/main"), {
+        recursive: true,
+      });
+      writeFileSync(
+        join(testDir, "packages/spotlight/src/electron/main/index.ts"),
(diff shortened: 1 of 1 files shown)


BYK (approved):
🚀 let's ship this

PR #432, opened 2026-03-16
feat: add --json flag to help command for agent introspection

BYK:
Phase 1 of #346: Add structured JSON output to the `help` command so AI agents can discover CLI commands at runtime.

## What

Three modes of `sentry help --json`:

- **Full tree**: `sentry help --json` → all routes, commands, flags, positional params as JSON
- **Route group**: `sentry help --json issue` → group metadata with subcommands
- **Specific command**: `sentry help --json issue list` → single command's full metadata

Framework-injected flags (`help`, `helpAll`, `log-level`, `verbose`) are stripped from JSON output so agents see only user-facing flags.

## Architecture

- **`src/lib/introspect.ts`** (new) — shared route-tree introspection module with types, type guards, extraction functions, and JSON cleaning utilities. Used at runtime by `help --json` and at build time by `generate-skill.ts`.
- **`src/commands/help.ts`** — adds `output: "json"` and `--json` flag. Dynamic imports for the route tree and introspect module to avoid circular deps.
- **`src/lib/help.ts`** — consolidated to import shared types from `introspect.ts` instead of duplicating them.
- **`script/generate-skill.ts`** — imports shared types and functions from `introspect.ts`, removing ~100 lines of duplication. No functional change to output.

## Tests

- 33 unit tests for introspection functions
- 12 property-based tests (fast-check) for invariants
- 10 integration tests for `help --json` end-to-end

### DIFF
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -3,23 +3,34 @@
  *
  * Provides help information for the CLI.
  * - `sentry help` or `sentry` (no args): Shows branded help with banner
- * - `sentry help <command>`: Shows Stricli's detailed help (--helpAll) for that command
+ * - `sentry help <command>`: Shows detailed help for that command
+ * - `sentry help --json`: Emits full command tree as structured JSON
+ * - `sentry help --json <command>`: Emits specific command/group metadata as JSON
  */
 
-import { run } from "@stricli/core";
 import type { SentryContext } from "../context.js";
 import { buildCommand } from "../lib/command.js";
+import { OutputError } from "../lib/errors.js";
 import { CommandOutput } from "../lib/formatters/output.js";
-import { printCustomHelp } from "../lib/help.js";
+import {
+  formatHelpHuman,
+  introspectAllCommands,
+  introspectCommand,
+  printCustomHelp,
+} from "../lib/help.js";
 
 export const helpCommand = buildCommand({
   docs: {
     brief: "Display help for a command",
     fullDescription:
       "Display help information. Run 'sentry help' for an overview, " +
-      "or 'sentry help <command>' for detailed help on a specific command.",
+      "or 'sentry help <command>' for detailed help on a specific command. " +
+      "Use --json for machine-readable output suitable for AI agents.",
   },
-  output: { human: (s: string) => s.trimEnd() },
+  output: {
+    human: formatHelpHuman,
+    jsonExclude: ["_banner"] as const,
+  },
   parameters: {
     flags: {},
     positional: {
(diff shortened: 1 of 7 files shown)


src/commands/help.ts
```diff
@@ -31,8 +35,48 @@ export const helpCommand = buildCommand({
...
-  // biome-ignore lint/complexity/noBannedTypes: Stricli requires empty object for commands with no flags
-  async *func(this: SentryContext, _flags: {}, ...commandPath: string[]) {
+  async *func(
+    this: SentryContext,
+    flags: { json?: boolean },
+    ...commandPath: string[]
+  ) {
+    if (flags.json) {
+      // Dynamic import to avoid circular dependency (app.ts imports helpCommand)
+      const { routes } = await import("../app.js");
```
BYK: We should find a way to avoid the circular import here altogether. There _must_ be a way.

src/commands/help.ts
```diff
@@ -31,8 +35,48 @@ export const helpCommand = buildCommand({
...
+        const resolved = resolveCommandPath(routeMap, commandPath);
+        if (!resolved) {
+          writeJson(this.stdout, {
+            error: `Command not found: ${commandPath.join(" ")}`,
+          });
+          return;
+        }
+
+        if (resolved.kind === "command") {
+          writeJson(this.stdout, cleanCommandInfoForJson(resolved.info));
```
BYK: Why are we not using the standard `--json`, `--fields` output system based on `yield`ing? We should not use `writeJson` anywhere

src/commands/help.ts
```diff
@@ -31,15 +38,25 @@ export const helpCommand = buildCommand({
...
     },
   },
-  // biome-ignore lint/complexity/noBannedTypes: Stricli requires empty object for commands with no flags
-  async *func(this: SentryContext, _flags: {}, ...commandPath: string[]) {
+  async *func(
+    this: SentryContext,
+    flags: { json?: boolean },
+    ...commandPath: string[]
+  ) {
+    if (flags.json) {
```
BYK: We should implement a human renderer and stop using the `flags.json` check at all.

src/lib/introspect.ts
```diff
@@ -0,0 +1,402 @@
...
+// ---------------------------------------------------------------------------
+// Flag Names Injected by the Framework
+// ---------------------------------------------------------------------------
+
+/**
+ * Flag names injected by buildCommand or Stricli that should be hidden
+ * from JSON introspection output. These are framework-level concerns,
+ * not user-facing per-command flags.
+ */
+const FRAMEWORK_FLAGS = new Set(["help", "helpAll", "log-level", "verbose"]);
```
BYK: I disagree, these should all be visible.

src/lib/introspect.ts
```diff
@@ -0,0 +1,402 @@
...
+ *
+ * @param info - Route info with all flags on commands
+ * @returns Route info with framework flags removed from all commands
+ */
+export function cleanRouteInfoForJson(info: RouteInfo): RouteInfo {
+  return {
+    ...info,
+    commands: info.commands.map(cleanCommandInfoForJson),
+  };
+}
```
BYK: If we remove the filtering above, I think these would bo obsolete

src/commands/help.ts
```diff
@@ -4,22 +4,52 @@
...
+ * Human renderer for help output.
+ *
+ * The command yields structured introspection data (for --json) but in
+ * human mode we display the branded banner for the full tree, or
+ * delegate to Stricli for specific commands.
+ *
+ * For the full tree case, this renders the pre-built branded banner.
+ * For specific command/group introspection data, it renders as
+ * indented JSON (the primary use case is --json mode anyway).
+ */
```
BYK: No, human readable output should not emit JSON

src/commands/help.ts
```diff
@@ -4,22 +4,52 @@
...
+function formatHelpHuman(data: HelpJsonResult): string {
+  // The full tree carries the pre-rendered banner for human display
+  if ("_banner" in data && typeof data._banner === "string") {
+    return data._banner.trimEnd();
+  }
+  return JSON.stringify(data, null, 2);
```
BYK: This has no place in a human-readable output.

PR #437, opened 2026-03-16
feat: add `sentry schema` command for API introspection

BYK:
Phase 2 of #346: Add `sentry schema` command for runtime API introspection by AI agents.

## What

Enables AI agents to discover Sentry API endpoints at runtime without pre-loaded documentation:

```bash
sentry schema                        # List all API resources with endpoint counts
sentry schema issues                 # Endpoints for 'issues' resource
sentry schema issues retrieve        # Detailed endpoint info (method, path, params, description)
sentry schema issues retrieve --json # Machine-readable output
sentry schema -q replay              # Search across all endpoints
sentry schema --all                  # Flat list of all 215 endpoints
sentry schema 'monitor*'             # Glob pattern matching on resources
```

## Architecture

**Build-time generation** — `script/generate-api-schema.ts` fetches the dereferenced OpenAPI spec from [`getsentry/sentry-api-schema`](https://github.com/getsentry/sentry-api-schema) (version-tagged to match the installed `@sentry/api` package) and extracts:
- HTTP methods and URL templates from the OpenAPI paths
- Human-readable descriptions and deprecation status
- Query parameter names from OpenAPI parameter definitions
- SDK function name mapping from the `@sentry/api` JS bundle
- Resource names derived from URL paths (last non-parameter segment)

Output: `src/generated/api-schema.json` (215 endpoints, ~117KB) — **gitignored**, generated at build/dev/test time via `bun run generate:schema`.

**Runtime** — `src/lib/api-schema.ts` provides query functions:
- `getResourceSummaries()` — grouped resource overview
- `getEndpointsByResource()` — filter by resource
- `getEndpoint()` — find specific endpoint (prefers operationId prefix match)
- `searchEndpoints()` — full-text search across all fields
- Glob pattern support via regex conversion for resource matching

The `sentry schema` command yields `CommandOutput` through the standard output system with `mdKvTable` + `renderMarkdown` for endpoint detail views.

## Other changes
- **Updated `@sentry/api`** from 0.21.0 → 0.54.0
- **CI workflow** updated to run `generate:schema` before lint, typecheck, and tests
- **`src/generated/`** added to `.gitignore`

## Tests

19 unit tests for schema query functions covering all query modes, edge cases, and data invariants.


script/generate-api-schema.ts
```diff
@@ -0,0 +1,293 @@
...
+ *   2. sdk.gen.d.ts — JSDoc descriptions, deprecation status
+ *
+ * Usage:
+ *   bun run script/generate-api-schema.ts
+ *
+ * Output:
+ *   src/generated/api-schema.json
+ */
+
+const API_PKG = "node_modules/@sentry/api/dist";
```
BYK: Shouldn't we use `require.resolve('@sentry/api')` or something?

script/generate-api-schema.ts
```diff
@@ -0,0 +1,293 @@
...
+ */
+async function parseJsBundle(): Promise<
+  Map<string, { method: string; url: string }>
+> {
+  const js = await Bun.file(`${API_PKG}/index.js`).text();
+  const results = new Map<string, { method: string; url: string }>();
+
+  // Match function declarations with method calls
+  const funcPattern =
+    /var (\w+) = \(options\S*\) => \(options\S*client \?\? client\)\.(\w+)\(/g;
```
BYK, line 87: This is quite fragile we should use a proper JS parser.

That being said, I think the better path is to get the version of the `@sentry/api` package, then just use the raw https://github.com/getsentry/sentry-api-schema/blob/main/openapi-derefed.json or https://github.com/getsentry/sentry-api-schema/blob/main/rpc_method_schema.json files there for generation.

src/commands/schema.ts
```diff
@@ -0,0 +1,287 @@
...
+      return `${method}${path}  ${label}${deprecated}`;
+    })
+    .join("\n");
+}
+
+/** Format a single endpoint in detail */
+function formatEndpointDetail(endpoint: ApiEndpoint): string {
+  const lines: string[] = [];
+
+  lines.push(`${formatMethod(endpoint.method)} ${bold(endpoint.path)}`);
```
BYK: Should we not use the markdown key-value table helpers here?

src/commands/schema.ts
```diff
@@ -0,0 +1,287 @@
...
+  docs: {
+    brief: "Browse the Sentry API schema",
+    fullDescription:
+      "Browse and search the Sentry API schema. Shows available resources, " +
+      "operations, and endpoint details. Use with --json for machine-readable output.\n\n" +
+      "Examples:\n" +
+      "  sentry schema                      List all API resources\n" +
+      "  sentry schema issues                Show endpoints for a resource\n" +
+      "  sentry schema issues list            Show details for one endpoint\n" +
+      "  sentry schema --list                Flat list of all endpoints\n" +
```
BYK: Shall we call this `--all`?

src/commands/schema.ts
```diff
@@ -0,0 +1,287 @@
...
+      "  sentry schema --search monitor      Search endpoints by keyword",
```
BYK, line 273: Can we do something like `sentry schema monitor*` for searching? At the very least we should add `-q` as a shorthand for `--search`

src/lib/api-schema.ts
```diff
@@ -0,0 +1,121 @@
...
+  return schema.find(
+    (e) =>
+      e.resource.toLowerCase() === lowerResource &&
+      (e.operationId.toLowerCase().includes(lowerQuery) ||
+        e.fn.toLowerCase().includes(lowerQuery))
+  );
+}
+
+/** Search endpoints by query string (matches fn, path, description, resource, operationId) */
+export function searchEndpoints(query: string): ApiEndpoint[] {
```
BYK, line 92: I think we already have fuzzy search / levenshtein distance helper somewhere in the code. If it's there, why not use that?

PR #445, opened 2026-03-17
fix(dsn): prevent hang during DSN auto-detection in repos with test fixtures

BYK:
Running `sentry issues` without explicit flags in repos containing test fixture DSNs (e.g., the CLI repo itself) caused the process to hang. PR #420 bumped scan depth from 2→3, causing `test/` directories to be scanned — finding ~86 fake DSNs that were all resolved via unbounded `Promise.all`, triggering 190+ concurrent API requests, rate limiting (429), and retry storms.

## Changes

- **Skip test directories in scanner** — Add `test/`, `tests/`, `__mocks__/`, `fixtures/`, `__fixtures__/` to `ALWAYS_SKIP_DIRS`. Test directories contain fixture DSNs, not real Sentry configuration.
- **Deduplicate DSNs before resolution** — Group by `(orgId, projectId)` or `publicKey`, resolving only one representative per unique pair. Reduces ~86 DSNs to ~10-20 unique API calls.
- **Add concurrency limit** — `p-limit(5)` on DSN resolution API calls to prevent overwhelming the Sentry API.
- **Add 15s timeout** — Race `Promise.all` against a deadline, returning partial results on timeout so the CLI never hangs indefinitely.
- **In-flight dedup for `resolveOrgRegion`** — Concurrent calls for the same orgSlug share a single HTTP request via a module-level promise cache.
- **Fix `getDirMtime` dynamic import** — Replace wasteful `await import("node:fs/promises")` (called 75+ times) with a static import.

Relates to #420 and #414.

### DIFF
--- a/src/lib/resolve-target.ts
+++ b/src/lib/resolve-target.ts
@@ -595,15 +595,15 @@ const DSN_RESOLVE_TIMEOUT_MS = 15_000;
 /**
  * Resolve DSNs with a concurrency limit and overall timeout.
  *
- * Uses p-limit's `map` helper for concurrency control and
- * `AbortSignal.timeout` for a self-cleaning deadline. Queued tasks
- * check the abort signal before doing work (same pattern as
- * code-scanner's earlyExit flag from PR #414).
+ * Uses p-limit's `map` helper for concurrency control and races it
+ * against `AbortSignal.timeout` so the CLI never hangs indefinitely.
+ * Queued tasks check the abort signal before doing work (same pattern
+ * as code-scanner's earlyExit flag from PR #414). In-flight tasks that
+ * already started are abandoned on timeout — their individual HTTP
+ * timeouts (30s in sentry-client.ts) bound them independently.
  *
- * Note: Cannot use `limit.clearQueue()` — as documented in
- * code-scanner.ts, it causes cleared promises to never settle,
- * hanging the map forever. Instead, queued tasks return null when
- * the signal is aborted.
+ * Results are written to a shared array so that tasks completing before
+ * the deadline are captured even when the overall operation times out.
  *
  * @param dsns - Deduplicated DSNs to resolve
  * @returns Array of resolved targets (null for failures/timeouts)
@@ -614,14 +614,31 @@ async function resolveDsnsWithTimeout(
   const limit = pLimit(DSN_RESOLVE_CONCURRENCY);
   const signal = AbortSignal.timeout(DSN_RESOLVE_TIMEOUT_MS);
 
-  const results = await limit.map(dsns, (dsn) => {
+  // Shared results array — tasks write their result as they complete,
+  // so partial results survive timeout.
+  const results: (ResolvedTarget | null)[] = new Array(dsns.length).fill(null);
+
+  const mapDone = limit.map(dsns, (dsn, i) => {
     if (signal.aborted) {
       return Promise.resolve(null);
     }
-    return resolveDsnToTarget(dsn);
+    return resolveDsnToTarget(dsn).then((target) => {
+      results[i] = target;
+      return target;
+    });
   });
 
-  if (signal.aborted) {
+  // Race limit.map against the abort signal so in-flight tasks
+  // don't block the timeout.
+  const aborted = new Promise<"timeout">((resolve) => {
+    signal.addEventListener("abort", () => resolve("timeout"), { once: true });
+  });
+  const raceResult = await Promise.race([
+    mapDone.then(() => "done" as const),
+    aborted,
+  ]);
+
+  if (raceResult === "timeout") {
     log.warn(
       `DSN resolution timed out after ${DSN_RESOLVE_TIMEOUT_MS / 1000}s, returning partial results`
     );


src/lib/region.ts
```diff
@@ -13,21 +13,50 @@ import { withAuthGuard } from "./errors.js";
...
 import { getSentryBaseUrl, isSentrySaasUrl } from "./sentry-urls.js";
 
+/**
+ * In-flight promise cache for org region resolution, keyed by orgSlug.
+ *
+ * When multiple DSNs share an orgId, concurrent calls to resolveOrgRegion
+ * deduplicate into a single HTTP request. A resolved promise returns
+ * instantly on `await`, so this also serves as a warm in-memory cache.
+ */
+const regionPromises = new Map<string, Promise<string>>();
```
BYK:
```suggestion
const regionCache = new Map<string, Promise<string>>();
```

src/lib/resolve-target.ts
```diff
@@ -578,11 +579,62 @@ export async function fetchProjectId(
...
+ * @param dsns - Deduplicated DSNs to resolve
+ * @returns Array of resolved targets (null for failures/timeouts)
+ */
+async function resolveDsnsWithTimeout(
+  dsns: DetectedDsn[]
+): Promise<(ResolvedTarget | null)[]> {
+  const limit = pLimit(DSN_RESOLVE_CONCURRENCY);
+  const results: (ResolvedTarget | null)[] = new Array(dsns.length).fill(null);
+
+  const resolvePromises = dsns.map((dsn, i) =>
```
BYK: pLimit has it's own `map` helper for this

src/lib/resolve-target.ts
```diff
@@ -578,11 +579,62 @@ export async function fetchProjectId(
...
+      results[i] = await resolveDsnToTarget(dsn);
+    })
+  );
+
+  const allDone = Promise.all(resolvePromises).then(() => "done" as const);
+  const timeout = new Promise<"timeout">((resolve) =>
+    setTimeout(() => resolve("timeout"), DSN_RESOLVE_TIMEOUT_MS)
+  );
+
+  const raceResult = await Promise.race([allDone, timeout]);
```
BYK: This would not work unless we cancel/abort pending other promises.

src/lib/region.ts
BYK: Applied — renamed to `regionCache`.

src/lib/resolve-target.ts
BYK: Switched to `limit.map` — much cleaner.

src/lib/resolve-target.ts
BYK: Replaced with `AbortSignal.timeout()` + abort check in the mapper — same pattern as code-scanner's `earlyExit` flag from PR #414. `AbortSignal.timeout` is self-cleaning (no manual `clearTimeout` needed) and queued tasks return `null` when the signal is aborted.

src/lib/region.ts
```diff
@@ -13,21 +13,50 @@ import { withAuthGuard } from "./errors.js";
...
-  // 1. Check cache first
+export function resolveOrgRegion(orgSlug: string): Promise<string> {
+  const existing = regionPromises.get(orgSlug);
+  if (existing) {
+    return existing;
+  }
+
+  const promise = resolveOrgRegionUncached(orgSlug);
+  regionPromises.set(orgSlug, promise);
+  return promise;
```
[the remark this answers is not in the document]
BYK: 👍 Good catch. Added `.catch(() => regionCache.delete(orgSlug))` to evict rejected promises from the cache. Non-auth errors already resolve to `baseUrl` fallback via `withAuthGuard`, so only `AuthError` re-throws can leave a rejected promise in the cache.

src/lib/resolve-target.ts
```diff
@@ -578,11 +579,62 @@ export async function fetchProjectId(
...
+  const resolvePromises = dsns.map((dsn, i) =>
+    limit(async () => {
+      results[i] = await resolveDsnToTarget(dsn);
+    })
+  );
+
+  const allDone = Promise.all(resolvePromises).then(() => "done" as const);
+  const timeout = new Promise<"timeout">((resolve) =>
+    setTimeout(() => resolve("timeout"), DSN_RESOLVE_TIMEOUT_MS)
+  );
```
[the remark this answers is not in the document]
BYK: Replaced `setTimeout` + `Promise.race` with `AbortSignal.timeout()` which is self-cleaning — no manual `clearTimeout` needed, does not hold the event loop.

src/lib/region.ts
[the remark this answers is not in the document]
BYK: Same as the Seer comment above — added `.catch` handler to evict rejected promises from `regionCache`.

src/lib/resolve-target.ts
```diff
@@ -578,11 +579,62 @@ export async function fetchProjectId(
...
+  dsns: DetectedDsn[]
+): Promise<(ResolvedTarget | null)[]> {
+  const limit = pLimit(DSN_RESOLVE_CONCURRENCY);
+  const signal = AbortSignal.timeout(DSN_RESOLVE_TIMEOUT_MS);
+
+  const results = await limit.map(dsns, (dsn) => {
+    if (signal.aborted) {
+      return Promise.resolve(null);
+    }
+    return resolveDsnToTarget(dsn);
```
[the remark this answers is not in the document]
BYK: 👍 Valid point. Now racing `limit.map` against the abort signal via `Promise.race` so the function returns on timeout even if in-flight tasks are still running. Results are written to a shared array so partial results survive. Individual HTTP requests are already bounded by the 30s `REQUEST_TIMEOUT_MS` in sentry-client.ts. Threading the abort signal into `resolveDsnToTarget` → `getProject` → SDK fetch would require plumbing changes across the entire API layer — a deeper change best done separately.

src/lib/resolve-target.ts
```diff
@@ -578,11 +579,62 @@ export async function fetchProjectId(
...
+  });
+
+  if (signal.aborted) {
+    log.warn(
+      `DSN resolution timed out after ${DSN_RESOLVE_TIMEOUT_MS / 1000}s, returning partial results`
+    );
+  }
+
+  return results;
+}
```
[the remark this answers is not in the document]
BYK: Fixed — now racing `limit.map` against the abort signal via `Promise.race`, so the function returns on timeout even if started tasks are still in-flight. See reply to the Seer comment above for details.

PR #454, opened 2026-03-17
fix: improve CLI output for auth login and upgrade flows

### LINKED ISSUE #452: Improve CLI output for auth login flow
## Summary

The `sentry auth login` flow has a few rough edges worth cleaning up. The main issues are: CLI output formatting is ugly (no spacing between emojis, and the emojis themselves are off), and the clickable copy-paste link doesn't include the auth code, making it unreliable in environments where the browser doesn't auto-open.

## Background

During the auth login flow, the CLI outputs a link for users to open in a browser. In environments where `xdg-open`/`open` doesn't work (e.g. SSH, CI, some Linux setups), the user must manually copy the link — but the code isn't embedded in it, so they're stuck. The current output also has formatting issues: emojis are jammed together with no spacing, and the emoji choices are a bit rough.

Reference: https://github.com/getsentry/sentry/pull/110909

## Scope

- Fix spacing/formatting in `auth login` CLI output
- Use better emoji choices (or none) for the output lines
- Embed the auth code in the copy-paste URL so it works in no-browser environments
- Out of scope: changes to the server-side auth approval flow

## Implementation

1. Audit the output strings in the auth login command
2. Fix spacing and emoji formatting
3. Update the generated URL to include the user code as a query param (confirm server accepts it)

Action taken on behalf of David Cramer.

BYK:
Migrate diagnostic output from cluttered stderr `log.info()` calls to the markdown-rendering pipeline for cleaner, more consistent formatting.

## Auth login (`interactive-login.ts`)
- Show `verification_uri_complete` (with `user_code` embedded) instead of bare `verification_uri` so the URL works in headless environments (SSH, CI)
- Use `log.log()` for URL/code detail lines instead of `log.info()` to avoid repeated ℹ icon prefixes
- Only show copy hint when browser didn't open
- Render user code as inline code span via `renderInlineMarkdown()`

## CLI upgrade (`upgrade.ts` + `human.ts`)
- Demote method/version/channel/latest `log.info()` calls to `log.debug()`
- Add metadata key-value table (method, channel) to `formatUpgradeResult` via `mdKvTable()`, rendering through the standard markdown pipeline
- Keep `log.info()` only for progress indicators that fire before the result is ready (`Upgrading to...`, nightly migration)

Closes #452

### DIFF
--- a/test/lib/interactive-login.test.ts
+++ b/test/lib/interactive-login.test.ts
@@ -82,38 +82,4 @@ describe("buildDeviceFlowDisplay", () => {
     // Without browser: extra copy-hint line + blank line
     expect(withoutBrowser.length).toBeGreaterThan(withBrowser.length);
   });
-
-  test("includes user code as inline code span", () => {
-    const lines = buildDeviceFlowDisplay(CODE, URL, true, false);
-    const joined = lines.join("\n");
-    expect(joined).toContain(`\`${CODE}\``);
-  });
-
-  test("omits copy hint when browser opened", () => {
-    const lines = buildDeviceFlowDisplay(CODE, URL, true, true);
-    const joined = lines.join("\n");
-    expect(joined).not.toContain("Copy the URL above");
-    expect(joined).not.toContain("to copy URL");
-  });
-
-  test("shows copy hint when browser did not open (TTY)", () => {
-    const lines = buildDeviceFlowDisplay(CODE, URL, false, true);
-    const joined = lines.join("\n");
-    expect(joined).toContain("Copy the URL above to sign in.");
-    expect(joined).toContain("to copy URL");
-  });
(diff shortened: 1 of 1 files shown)


BYK:
Addressed both review comments in af24db30:

**Seer (URL escaping):** Good catch. Wrapped `verificationUriComplete` with `escapeMarkdownInline()` before passing to `renderInlineMarkdown()`. Added a test with underscore-containing URLs to verify.

**Cursor BugBot (TTY-dependent assertions):** Set `SENTRY_PLAIN_OUTPUT=1` in the upgrade test mock context (`createMockContext`) so `formatUpgradeResult` always renders raw markdown tables with ASCII pipes, regardless of whether the test runner is in a TTY.

PR #459, opened 2026-03-18
fix: clean up upgrade output and hide empty table headers

BYK:
Clean up the ugly `cli upgrade` output and fix a systemic issue with
empty header rows in box-drawing tables across the codebase.

## Upgrade output

Before:
```
Binary: Installed to /home/byk/.local/bin/sentry
✓ Upgraded to  0.18.0-dev.1773836960
0.18.0-dev.1773794950 → 0.18.0-dev.1773836960

╭─────────┬─────────╮
│         │         │
├─────────┼─────────┤
│ Method  │ curl    │
│ Channel │ nightly │
╰─────────┴─────────╯
```

After:
```
✓ Upgraded to 0.18.0-dev.1773836960 (from 0.18.0-dev.1773794950)
  Method: curl · Channel: nightly
```

Changes:
- Replace metadata table with a compact muted inline line
- Remove redundant version arrow; fold "from" version into success line
- Add `--quiet` to `cli setup` subprocess to suppress leaked stdout

## Systemic table fix

`mdKvTable()` without a heading produces `| | |` empty header cells,
which rendered as an ugly empty bordered row in box-drawing tables.
This affected 12 call sites: auth status, issue/org/project details,
trace/span/log views, and more.

Fix: add `hideHeaders` option to `renderTextTable` and auto-detect
all-empty headers in `renderTableToken` to skip the header row while
still using it for column width measurement.

### DIFF
--- a/src/lib/formatters/text-table.ts
+++ b/src/lib/formatters/text-table.ts
@@ -106,9 +106,13 @@ export function renderTextTable(
     minWidths = [],
     shrinkable = [],
     truncate = false,
-    hideHeaders = options.hideHeaders ?? headers.every((h) => h.trim() === ""),
     rowSeparator = false,
   } = options;
+
+  // Auto-detect empty headers when not explicitly set by the caller.
+  // Extracted from destructuring because `??` preserves explicit `false`.
+  const hideHeaders =
+    options.hideHeaders ?? headers.every((h) => h.trim() === "");
 
   const border = BorderChars[borderStyle];
   const colCount = headers.length;


src/lib/formatters/markdown.ts
```diff
@@ -571,7 +575,11 @@ function renderTableToken(table: Tokens.Table): string {
     return "left";
   });
 
-  return renderTextTable(headers, rows, { alignments });
+  // Hide the header row when all cells are empty — avoids rendering a
+  // visually empty bordered row above the data (common in mdKvTable output).
+  const hideHeaders = headers.every((h) => h.trim() === "");
```
BYK: Shouldn't this logic live in `renderTextTable` instead?

BYK: Good call — moved the auto-detection into `renderTextTable` itself. It now auto-hides headers when all cells are empty/whitespace, with an explicit `hideHeaders: true/false` override available. `renderTableToken` no longer needs any special handling.

src/lib/formatters/text-table.ts
```diff
@@ -94,6 +106,7 @@ export function renderTextTable(
     minWidths = [],
     shrinkable = [],
     truncate = false,
+    hideHeaders = options.hideHeaders ?? headers.every((h) => h.trim() === ""),
```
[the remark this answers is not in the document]
BYK: Good catch. Extracted `hideHeaders` from the destructuring into a separate `const` using `??` so that explicit `false` from callers is correctly preserved.

PR #465, opened 2026-03-18
feat: dynamic cache-backed shell completions with fuzzy matching

BYK:
## Summary

Add a hybrid static + dynamic shell completion system that suggests cached org slugs, project slugs, and project aliases alongside existing command/subcommand names — with fuzzy matching for typo tolerance.

## Approach: Hybrid Static + Dynamic

- **Static** (0ms): command/subcommand names, flag names, and enum flag values are embedded in the shell script
- **Dynamic** (<50ms): positional arg values are completed by calling `sentry __complete` at runtime, which reads the SQLite cache with fuzzy matching
- **Lazy fetch** (one-time ~200-500ms): if no projects are cached for an org, fetches and caches them on first tab-complete

## New Files

- `src/lib/fuzzy.ts` — Shared Levenshtein distance + `fuzzyMatch()` with tiered scoring (exact > prefix > contains > Levenshtein distance)
- `src/lib/complete.ts` — Completion engine: handles the `__complete` fast-path with context parsing, cache querying, and lazy project fetching
- `test/lib/fuzzy.test.ts` — Property-based + unit tests for fuzzy matching
- `test/lib/complete.test.ts` — Tests for the completion engine

## Key Changes

- **`src/bin.ts`**: `__complete` fast-path at top of `main()` — skips all middleware, telemetry, and auth
- **`src/lib/completions.ts`**: Extended `extractCommandTree()` with flag metadata; updated bash/zsh/fish generators with flag completion + dynamic `__complete` callback
- **`src/lib/db/project-cache.ts`**: Added `getAllCachedProjects()` and `getCachedProjectsForOrg()` for completion queries
- **`src/lib/platforms.ts`**: Uses shared `levenshtein()` from `fuzzy.ts` instead of local copy

## How Tab Completion Works

```
sentry <TAB>          → commands (static, instant)
sentry issue <TAB>    → subcommands (static, instant)
sentry issue list --<TAB>     → flags (static, instant)
sentry issue list <TAB>       → org slugs with "/" (dynamic, from cache)
sentry issue list myorg/<TAB> → projects for that org (dynamic, lazy-fetches if cold)
sentry issue list senry/<TAB> → "sentry/" via fuzzy match (Levenshtein distance 1)
```


src/lib/complete.ts
```diff
@@ -0,0 +1,273 @@
...
+      completeAliases(partial),
+    ]);
+    return [...orgCompletions, ...aliasCompletions];
+  }
+
+  // Has slash — complete project within the org
+  const orgPart = partial.slice(0, slashIdx);
+  const projectPart = partial.slice(slashIdx + 1);
+
+  return completeProjectSlugs(projectPart, orgPart);
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in 0cfe02ed — added `fuzzyResolveOrg()` that fuzzy-matches the org part against cached slugs before querying for projects. Now `senry/` correctly resolves to `sentry` and shows its projects. Added tests for both the happy path and unresolvable org case.

src/lib/completions.ts
```diff
@@ -146,38 +366,57 @@ complete -F _${binaryName}_completions ${binaryName}
...
+_${binaryName}_dynamic() {
+  local -a completions
+  local line
+  while IFS=$'\\t' read -r value desc; do
+    if [[ -n "$desc" ]]; then
+      completions+=("\${value}:\${desc}")
+    else
+      completions+=("\${value}")
+    fi
+  done < <(${binaryName} __complete \${words[2,-1]} 2>/dev/null)
```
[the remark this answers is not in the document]
BYK: Excellent find! Fixed in 0cfe02ed — the zsh completion function now:
1. Declares `local line` at the top of the main function scope (where `_arguments -C` sets it)
2. Uses `$line[1]` and `$line[2]` (the parsed command and subcommand) to pass full context to `__complete`
3. Removed the separate `_dynamic()` function — the completion logic is now inline in the `args` state where `$line` is available

This correctly passes the command/subcommand context even after `_arguments` resets `$words` in the `args` state.

src/lib/complete.ts
```diff
@@ -0,0 +1,300 @@
...
+    precedingWords.length >= 2
+      ? `${precedingWords[0]} ${precedingWords[1]}`
+      : "";
+
+  // Check if this is a flag value position (previous word is a flag)
+  const lastWord = precedingWords.at(-1);
+  if (lastWord?.startsWith("--")) {
+    // We're completing a flag value — don't provide org/project completions
+    return [];
+  }
```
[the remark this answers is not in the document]
BYK: Good catch! Fixed in 83de0c97 — removed the `--` prefix heuristic entirely. Boolean flags like `--verbose` and `--json` don't consume the next word, so suppressing completions after any `--` flag was wrong. The shell script already handles flag value completion statically via the enum variable lookup, so the dynamic completer doesn't need to second-guess.

src/lib/complete.ts
```diff
@@ -0,0 +1,300 @@
...
+  const slugs = orgs.map((o) => o.slug);
+  const matched = fuzzyMatch(partial, slugs);
+
+  const nameMap = new Map(orgs.map((o) => [o.slug, o.name]));
+
+  return matched.map((slug) => ({
+    value: `${slug}/`,
+    description: nameMap.get(slug),
+  }));
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 83de0c97 — extracted `matchOrgSlugs(partial, formatValue)` as a shared helper parameterized by a value formatter function. Both `completeOrgSlugs` (bare slugs for org-only commands) and `completeOrgSlugsWithSlash` (with trailing `/` for org/project commands) are now one-liners that delegate to it.

src/lib/completions.ts
```diff
@@ -29,96 +88,223 @@ type CommandGroup = {
...
   return { groups, standalone };
 }
 
+/**
+ * Sanitize a string for use in a shell variable name.
+ * Replaces hyphens and dots with underscores.
+ */
+function shellVarName(s: string): string {
+  return s.replace(/[-. ]/g, "_");
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 74ac8e6f — added a `__sentry_varname()` bash helper that uses `${1//[-. ]/_}` to match the same replacement pattern as `shellVarName()` in TypeScript. All runtime variable lookups now go through this helper instead of the inline `${var//-/_}`.

src/lib/completions.ts
```diff
@@ -244,12 +498,34 @@ export function generateFishCompletion(binaryName: string): string {
 # Disable file completion by default
 complete -c ${binaryName} -f
 
+# Dynamic completion for positional values
+function __${binaryName}_complete_dynamic
+  set -l tokens (commandline -opc)
+  ${binaryName} __complete $tokens[2..] (commandline -ct) 2>/dev/null | while read -l line
+    echo $line
+  end
+end
```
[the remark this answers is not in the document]
BYK: Fixed in 74ac8e6f — split the fish function into `set -l preceding (commandline -opc)` and `set -l current (commandline -ct)` as separate variables, then pass `$preceding[2..] $current` to `__complete`. This correctly separates the preceding words from the partial token.

src/lib/db/project-cache.ts
```diff
@@ -137,6 +137,30 @@ export async function setCachedProjectByDsnKey(
...
+ */
+export async function getCachedProjectsForOrg(
+  orgSlug: string
+): Promise<{ projectSlug: string; projectName: string }[]> {
+  const db = getDatabase();
+  const rows = db
+    .query(
+      "SELECT DISTINCT project_slug, project_name FROM project_cache WHERE org_slug = ?"
+    )
+    .all(orgSlug) as Pick<ProjectCacheRow, "project_slug" | "project_name">[];
```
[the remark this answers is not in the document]
BYK: Fixed in 6cd74f8f — changed `SELECT DISTINCT project_slug, project_name` to `GROUP BY project_slug ORDER BY cached_at DESC` so the same project cached under different keys with slightly different names only appears once, using the most recently cached name.

src/lib/completions.ts
```diff
@@ -217,24 +462,35 @@ export function generateFishCompletion(binaryName: string): string {
...
+        .flatMap((s) =>
+          s.flags.map(
+            (f) =>
+              `complete -c ${binaryName} -n "__fish_seen_subcommand_from ${g.name}; and __fish_seen_subcommand_from ${s.name}" -l "${f.name}" -d "${escapeDblQuote(f.brief)}"`
+          )
         )
         .join("\n");
-      return `\n# ${g.name} subcommands\n${lines}`;
+
+      return `\n# ${g.name} subcommands\n${cmdLines}${flagLines ? `\n${flagLines}` : ""}`;
```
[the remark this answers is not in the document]
BYK: Fixed in 6cd74f8f — added a `standaloneFlagLines` section to the fish generator that emits `complete -c ... -l flagname -d description` for each standalone command's flags.

src/lib/completions.ts
```diff
@@ -244,12 +514,38 @@ export function generateFishCompletion(binaryName: string): string {
 # Disable file completion by default
 complete -c ${binaryName} -f
 
+# Dynamic completion for positional values
+function __${binaryName}_complete_dynamic
+  # commandline -opc: all tokens before cursor (excludes current partial)
+  # commandline -ct: the current token being completed (the partial)
+  set -l preceding (commandline -opc)
+  set -l current (commandline -ct)
+  ${binaryName} __complete $preceding[2..] $current 2>/dev/null | while read -l line
```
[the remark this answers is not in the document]
BYK: Fixed in 9c0c61eb — quoted `$current` in the fish function so an empty partial (TAB after space) is correctly passed as `""` to `__complete` instead of being silently dropped.

src/lib/completions.ts
```diff
@@ -29,104 +88,237 @@ type CommandGroup = {
...
+  const enumCaseEntries = uniqueEnumFlags.map((f) => `"${f}"`).join("|");
+
+  // Build the enum flag test only if there are any enum flags
+  const enumBranch = enumCaseEntries
+    ? `elif [[ "\${prev}" == @(${enumCaseEntries}) ]]; then
+        # Enum value completion (static)
+        local val_var="$(__${binaryName}_varname "\${cmd}")_$(__${binaryName}_varname "\${subcmd}")_$(__${binaryName}_varname "\${prev#--}")_values"
+        if [[ -n "\${!val_var+x}" ]]; then
+          COMPREPLY=($(compgen -W "\${!val_var}" -- "\${cur}"))
+        fi`
```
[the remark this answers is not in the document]
BYK: Fixed in 9c0c61eb — added a standalone fallback for the enum value lookup. When `COMP_WORDS[2]` is a flag (not a subcommand), the `cmd_subcmd_flag_values` lookup fails, so we now fall back to `cmd_flag_values`.

src/lib/db/project-cache.ts
```diff
@@ -137,6 +137,33 @@ export async function setCachedProjectByDsnKey(
...
+): Promise<{ projectSlug: string; projectName: string }[]> {
+  const db = getDatabase();
+  // Group by project_slug to deduplicate — take the most recently cached name
+  // in case the same project appears under different cache keys with slightly
+  // different names (e.g., after a rename between caching events).
+  const rows = db
+    .query(
+      "SELECT project_slug, project_name FROM project_cache WHERE org_slug = ? GROUP BY project_slug ORDER BY cached_at DESC"
+    )
+    .all(orgSlug) as Pick<ProjectCacheRow, "project_slug" | "project_name">[];
```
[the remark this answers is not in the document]
BYK: Fixed in bd4778ec — replaced `GROUP BY ... ORDER BY cached_at DESC` with `MAX(cached_at)` aggregate. SQLite guarantees that non-aggregated columns come from the row that produced the MAX/MIN value, so `project_name` is deterministically picked from the most recently cached row.

src/lib/complete.ts
```diff
@@ -0,0 +1,294 @@
...
+  // Has slash — complete project within the org.
+  // Fuzzy-resolve the org slug first so that "senry/" still finds "sentry" projects.
+  const orgPart = partial.slice(0, slashIdx);
+  const projectPart = partial.slice(slashIdx + 1);
+
+  const resolvedOrg = await fuzzyResolveOrg(orgPart);
+  if (!resolvedOrg) {
+    return [];
+  }
+
```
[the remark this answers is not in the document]
BYK: Fixed in d95bf369 — added early return when `orgPart` is empty (bare `/` input). Prevents `fuzzyResolveOrg("")` from matching the first alphabetical org.

src/lib/completions.ts
```diff
@@ -196,10 +428,26 @@ ${subArrays}
...
+      # In the args state, $line[1] and $line[2] hold the parsed command
+      # and subcommand. Pass them to __complete for context detection.
+      local -a completions
+      while IFS=$'\\t' read -r value desc; do
+        if [[ -n "$desc" ]]; then
+          completions+=("\${value}:\${desc}")
+        else
+          completions+=("\${value}")
+        fi
+      done < <(${binaryName} __complete "$line[1]" "$line[2]" "\${words[@]}" 2>/dev/null)
```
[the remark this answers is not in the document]
BYK: Fixed in 5d272c94 — when `$words` is empty (no positional args yet), we now explicitly pass an empty `""` as the partial. When `$words` has content, it's passed as-is since the last element naturally serves as the partial.

src/bin.ts
```diff
@@ -160,11 +160,18 @@ function buildExecutor(): (args: string[]) => Promise<void> {
 const executeCommand = buildExecutor();
 
 async function main(): Promise<void> {
+  const args = process.argv.slice(2);
+
+  // Fast-path: shell completion (no telemetry, no middleware, no upgrade check)
+  if (args[0] === "__complete") {
+    const { handleComplete } = await import("./lib/complete.js");
```
BYK: Do we need a dynamic import here? Wouldn't a static import be faster?

src/bin.ts
```diff
@@ -160,11 +160,18 @@ function buildExecutor(): (args: string[]) => Promise<void> {
 const executeCommand = buildExecutor();
 
 async function main(): Promise<void> {
+  const args = process.argv.slice(2);
+
+  // Fast-path: shell completion (no telemetry, no middleware, no upgrade check)
```
BYK: How much overhead would we get from adding telemetry here as ensuring fast autocompletions would be important for a good UX?

src/lib/completions.ts
```diff
@@ -18,8 +26,59 @@ export type CompletionLocation = {
   created: boolean;
 };
 
-/** A command with its description */
-type CommandEntry = { name: string; brief: string };
+/**
+ * Check if a routing target is a route map (has subcommands).
+ * Works with untyped Stricli route targets (avoids generic type constraints).
+ */
+function isRouteMap(target: unknown): target is {
```
BYK: Shouldn't these live in `introspect.js`?

src/lib/completions.ts
```diff
@@ -29,104 +88,243 @@ type CommandGroup = {
...
  */
-function isRouteMap(target: unknown): target is {
+function extractFlagEntries(
+  flags: Record<string, FlagDef> | undefined
+): FlagEntry[] {
+  if (!flags) {
+    return [];
+  }
+
+  return Object.entries(flags)
```
BYK: For small number of flags, not sure if this makes a big difference but using a `for .. of` loop here would be more efficient both in terms of memory and processing

src/lib/completions.ts
```diff
@@ -29,104 +88,243 @@ type CommandGroup = {
...
-    target !== null &&
-    typeof (target as Record<string, unknown>).getAllEntries === "function"
-  );
+}): CommandEntry[] {
+  const subcommands: CommandEntry[] = [];
+  for (const sub of routeMap.getAllEntries()) {
+    if (!sub.hidden) {
+      subcommands.push(buildCommandEntry(sub.name.original, sub.target));
+    }
+  }
```
BYK, line 121: Why not `.filter()` here? Alternatively, if `getAllEntries` can be a generator, we can save on some memory?

src/lib/completions.ts
```diff
@@ -29,104 +88,243 @@ type CommandGroup = {
...
   return { groups, standalone };
 }
 
+/**
+ * Sanitize a string for use in a shell variable name.
+ * Replaces hyphens and dots with underscores.
+ */
+function shellVarName(s: string): string {
+  return s.replace(/[-. ]/g, "_");
+}
```
BYK, line 167: How safe is this? Don't we have a built-in helper in Bun or node stdlib?

src/lib/completions.ts
```diff
@@ -29,104 +88,243 @@ type CommandGroup = {
...
+        else
+          # Fallback for standalone commands (subcmd is a flag, not a subcommand)
+          local standalone_val_var="$(__${binaryName}_varname "\${cmd}")_$(__${binaryName}_varname "\${prev#--}")_values"
+          if [[ -n "\${!standalone_val_var+x}" ]]; then
+            COMPREPLY=($(compgen -W "\${!standalone_val_var}" -- "\${cur}"))
+          fi
+        fi`
+    : "";
+
   return `# bash completion for ${binaryName}
```
BYK, line 282: I wonder if we should have a template string tag function to make sure the substitutions we make here are safe (escaped)?

Also wondering if these shell scripts should live in their dedicated template files, outside the code?

src/lib/complete.ts
```diff
@@ -0,0 +1,299 @@
...
+    .map((c) => (c.description ? `${c.value}\t${c.description}` : c.value))
+    .join("\n");
+
+  if (output) {
+    process.stdout.write(`${output}\n`);
+  }
+}
+
+/** Commands that accept org/project positional args. */
+const ORG_PROJECT_COMMANDS = new Set([
```
BYK: Can we not dynamically derive this list? Okay doing this at build time too but a manually curated list like this would go out of sync easily and quickly

src/lib/complete.ts
```diff
@@ -0,0 +1,299 @@
...
+  "span list",
+  "span view",
+  "event view",
+  "log list",
+  "log view",
+  "dashboard list",
+]);
+
+/** Commands that accept only an org slug (no project). */
+const ORG_ONLY_COMMANDS = new Set([
```
BYK: Same as above, re dynamic generation

src/lib/complete.ts
```diff
@@ -0,0 +1,299 @@
...
+): Promise<Completion[]> {
+  const orgs = await getCachedOrganizations();
+  if (orgs.length === 0) {
+    return [];
+  }
+
+  const slugs = orgs.map((o) => o.slug);
+  const matched = fuzzyMatch(partial, slugs);
+
+  const nameMap = new Map(orgs.map((o) => [o.slug, o.name]));
```
BYK, line 173: Looks like we can generate this map earlier and completely bypass creating that separate `slugs` array?

src/lib/complete.ts
```diff
@@ -0,0 +1,299 @@
...
+/**
+ * Complete organization slugs with fuzzy matching.
+ *
+ * Queries the org_regions cache for all known org slugs and matches
+ * them against the partial input.
+ *
+ * @param partial - Partial org slug to match
+ * @returns Completions with org names as descriptions
+ */
+export function completeOrgSlugs(partial: string): Promise<Completion[]> {
```
BYK: Is this helper really useful? Also I'm curious about the identity function we pass to `matchOrgSlugs()`. Is it not avoiable?

src/bin.ts
BYK: Fixed in 9e86526b — `db/index.ts` no longer imports telemetry at module level, so it's safe as a static import. Changed to `import { disableDbTracing } from "./lib/db/index.js"`.

src/bin.ts
BYK: Benchmarked: `Sentry.init()` adds ~30ms, `client.flush()` is async (fires after stdout). After the lazy-import fix, completions are ~60ms dev / ~190ms binary. Adding telemetry would add ~30ms user-perceived. We could do it as a follow-up — the import chain fix in this PR is the prerequisite.

src/lib/completions.ts
BYK: Fixed in 9e86526b — widened `isRouteMap`/`isCommand` in `introspect.ts` to accept `unknown`, then removed the duplicate type guards from `completions.ts` and imported the shared versions.

src/lib/completions.ts
BYK: Fixed in 9e86526b — replaced `filter().map()` with a `for..of` loop.

src/lib/completions.ts
BYK: Fixed in 9e86526b — replaced the manual for loop with `.filter().map()`.

src/lib/completions.ts
BYK: Fixed in 9e86526b — switched to a whitelist: `replace(/[^a-zA-Z0-9_]/g, "_")`. Input is Stricli route/flag names (alphanumeric + hyphens), but the whitelist guards against anything unexpected. No stdlib helper exists for shell variable sanitization. The bash `__varname` helper was updated to match.

src/lib/completions.ts
BYK: Both are great ideas for a follow-up! A tagged template function for shell escaping would make the generation safer and more readable. Moving templates to dedicated files would also improve maintainability. Filed as future improvements — keeping this PR focused on the completion engine itself.

src/lib/complete.ts
BYK: Cannot derive at runtime — `app.ts` imports `@sentry/bun` at line 2, and `complete.ts` is the fast-path that avoids that (~280ms). Instead added a drift-detection test in `completions.property.test.ts` that walks the route tree and verifies every command with an org-related placeholder appears in one of the sets. Also exported the sets for testability.

src/lib/complete.ts
BYK: Same approach — covered by the drift-detection test. See above.

src/lib/complete.ts
BYK: Fixed in 9e86526b — build slugs directly from `orgs.map(o => o.slug)` and pass to `fuzzyMatch`, then build nameMap separately. Eliminated `[...nameMap.keys()]` intermediate.

src/lib/complete.ts
BYK: Fixed in 9e86526b — merged `matchOrgSlugs` into `completeOrgSlugs` with an optional `suffix` parameter (default `""`). `completeOrgSlugsWithSlash` now just calls `completeOrgSlugs(partial, "/")`. No identity function, no separate helper.

src/lib/completions.ts
```diff
@@ -29,104 +88,243 @@ type CommandGroup = {
...
+__${binaryName}_varname() { echo "\${1//[-. ]/_}"; }
+
 _${binaryName}_completions() {
   local cur prev words cword
   _init_completion || return
 
   local commands="${allTopLevel.join(" ")}"
 ${subVars}
+${flagVarLines.join("\n")}
+${enumVarLines.join("\n")}
```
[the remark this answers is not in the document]
BYK: Fixed in 9e86526b — added `*)` fallback in the bash `COMP_CWORD=2` case for standalone command flags and dynamic completions. For zsh, the `subcommand` state now uses `;;&` (test-next-pattern) fallthrough so standalone commands get dynamic completion from the `args` handler.

src/lib/completions.ts
```diff
@@ -18,8 +26,59 @@ export type CompletionLocation = {
...
+    flags?: Record<string, FlagDef>;
+  };
+} {
+  return (
+    typeof target === "object" &&
+    target !== null &&
+    "parameters" in target &&
+    !("getAllEntries" in target)
+  );
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 9e86526b — removed the duplicate type guards from completions.ts and imported `isRouteMap`/`isCommand` from introspect.ts. Widened the introspect.ts guards to accept `unknown`.

src/lib/complete.ts
```diff
@@ -0,0 +1,299 @@
...
+]);
+
+/** Commands that accept only an org slug (no project). */
+const ORG_ONLY_COMMANDS = new Set([
+  "org view",
+  "team list",
+  "repo list",
+  "trial list",
+  "trial start",
+]);
```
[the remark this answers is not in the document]
BYK: Addressed in 9e86526b — exported the sets and added drift-detection tests in `completions.property.test.ts` that verify: (1) every entry in the hardcoded sets maps to a real route, (2) every command with an org-related positional is in at least one set, (3) no overlaps between sets.

src/lib/completions.ts
```diff
@@ -244,12 +514,40 @@ export function generateFishCompletion(binaryName: string): string {
...
 
+/** Escape single quotes for zsh completion descriptions. */
+function escapeSingleQuote(s: string): string {
+  return s.replace(/'/g, "'\\''");
+}
+
+/** Escape double quotes and backslashes for fish/bash completion descriptions. */
+function escapeDblQuote(s: string): string {
+  return s.replace(/\\/g, "\\\\").replace(/"/g, '\\"');
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 1000bdcb — added `$` escaping to `escapeDblQuote()`. Now escapes `\`, `$`, and `"` in that order.

src/lib/completions.ts
```diff
@@ -196,9 +413,34 @@ ${subArrays}
...
       ;;
     subcommand)
-      case "$words[1]" in
+      case "$line[1]" in
 ${caseBranches}
+        *)
+          # Standalone command — delegate to dynamic completion
+          ;;
       esac
+      ;;&
```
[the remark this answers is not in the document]
BYK: Fixed in b71cd4fa — changed `;;&` to `;|` which is the canonical zsh syntax for testing remaining case patterns. Both work in zsh, but `;|` is the idiomatic form.

src/lib/complete.ts
BYK: I still see `const slugs` on line 158 above?

src/lib/complete.ts
```diff
@@ -63,8 +63,16 @@ export async function handleComplete(args: string[]): Promise<void> {
...
+/**
+ * Commands that accept org/project positional args.
+ *
+ * Hardcoded for fast-path performance — cannot derive from the route map
+ * at runtime because `app.ts` imports `@sentry/bun`. A property test in
+ * `completions.property.test.ts` verifies this set stays in sync.
+ *
+ * @internal Exported for testing only.
+ */
+export const ORG_PROJECT_COMMANDS = new Set([
```
BYK, line 90: Crazy-ish idea: import these from a file that we generate at build time?

src/lib/completions.ts
```diff
@@ -148,13 +113,10 @@ function extractSubcommands(routeMap: {
...
-  for (const sub of routeMap.getAllEntries()) {
-    if (!sub.hidden) {
-      subcommands.push(buildCommandEntry(sub.name.original, sub.target));
-    }
-  }
-  return subcommands;
+  return routeMap
+    .getAllEntries()
+    .filter((sub) => !sub.hidden)
+    .map((sub) => buildCommandEntry(sub.name.original, sub.target));
```
BYK: Why the change here? I think for-of loop was more efficient?

src/lib/completions.ts
```diff
@@ -191,10 +153,14 @@ export function extractCommandTree(): CommandTree {
 
 /**
  * Sanitize a string for use in a shell variable name.
- * Replaces hyphens and dots with underscores.
+ *
+ * Uses a whitelist approach — replaces any character that is not a valid
```
BYK: s/whitelist/allowlist

src/lib/complete.ts
BYK: Fixed in 6d97421f — now builds `nameMap` first and passes `[...nameMap.keys()]` to `fuzzyMatch`. No separate `slugs` array.

src/lib/complete.ts
BYK: Interesting idea! A build-time generated file would let us derive the sets from the route map without the runtime import cost. The drift-detection test catches mismatches today, but codegen would eliminate the manual lists entirely. Filed as a follow-up — would need changes to `script/build.ts` and a new generated module.

src/lib/completions.ts
BYK: You're right — reverted to `for..of` in 6d97421f. I misread the original comment as a request to change to `.filter()`. The for-of loop is more efficient since it avoids creating an intermediate filtered array.

src/lib/completions.ts
BYK: Fixed in 6d97421f — changed to "allowlist".

src/lib/complete.ts
```diff
@@ -155,10 +155,9 @@ export async function completeOrgSlugs(
     return [];
   }
 
-  const slugs = orgs.map((o) => o.slug);
-  const matched = fuzzyMatch(partial, slugs);
-
   const nameMap = new Map(orgs.map((o) => [o.slug, o.name]));
+  const matched = fuzzyMatch(partial, [...nameMap.keys()]);
```
BYK: Use `Array.from(nameMap.keys())` which is more efficient than spreading?

BYK: Fixed in 22324604 — changed to `Array.from(nameMap.keys())`.

src/lib/completions.ts
```diff
@@ -244,12 +517,40 @@ export function generateFishCompletion(binaryName: string): string {
...
 
+/** Escape single quotes for zsh completion descriptions. */
+function escapeSingleQuote(s: string): string {
+  return s.replace(/'/g, "'\\''");
+}
+
+/** Escape characters special in double-quoted shell strings (bash/fish). */
+function escapeDblQuote(s: string): string {
+  return s.replace(/\\/g, "\\\\").replace(/\$/g, "\\$").replace(/"/g, '\\"');
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 7ee4e82c — added `\(` and `\)` escaping to `escapeDblQuote()`. Fish treats `()` as command substitution inside double quotes.

src/lib/fuzzy.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Handles empty strings, equal strings, and arbitrary Unicode input.
+ *
+ * @param a - First string
+ * @param b - Second string
+ * @returns The minimum number of single-character edits (insert, delete, replace)
+ */
+export function levenshtein(a: string, b: string): number {
+  const m = a.length;
+  const n = b.length;
+  const dp = new Map<number, number>();
```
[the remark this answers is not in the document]
BYK: Acknowledged — the `levenshtein` function was extracted as-is from the existing `platforms.ts` (line 220). Switching to a flat array or two-row optimization would be a worthwhile follow-up but is out of scope for this PR since it's a pre-existing pattern.

src/lib/completions.ts
```diff
@@ -196,9 +416,34 @@ ${subArrays}
...
+        if [[ -n "$desc" ]]; then
+          completions+=("\${value}:\${desc}")
+        else
+          completions+=("\${value}")
+        fi
+      done < <(
+        if (( \${#words} == 0 )); then
+          ${binaryName} __complete "$line[1]" "$line[2]" "" 2>/dev/null
+        else
+          ${binaryName} __complete "$line[1]" "$line[2]" "\${words[@]}" 2>/dev/null
```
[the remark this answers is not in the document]
BYK: In zsh, `_arguments -C` with `1: :->command` and `2: :->subcommand` populates `$line[1]` and `$line[2]` with the parsed values from those positions. In the `args` state (position 3+), both should be populated. For standalone commands entering via `;|` fall-through, `$line[1]` has the command name and `$line[2]` may be the first positional arg — `handleComplete()` gracefully handles this by returning empty when the cmdPath doesn't match any known command. Not ideal for standalone completions in zsh, but not broken.

src/lib/complete.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  const slugs = projects.map((p) => p.projectSlug);
+  const matched = fuzzyMatch(projectPartial, slugs);
+
+  const nameMap = new Map(projects.map((p) => [p.projectSlug, p.projectName]));
+
+  return matched.map((slug) => ({
+    value: `${orgSlug}/${slug}`,
+    description: nameMap.get(slug),
+  }));
```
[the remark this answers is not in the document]
BYK: Fixed in 0aebaf4f — same consolidation as `completeOrgSlugs`: build `nameMap` first, pass `Array.from(nameMap.keys())` to `fuzzyMatch`.

src/lib/completions.ts
```diff
@@ -196,9 +416,34 @@ ${subArrays}
...
       ;;
     subcommand)
-      case "$words[1]" in
+      case "$line[1]" in
 ${caseBranches}
+        *)
+          # Standalone command — delegate to dynamic completion
+          ;;
       esac
+      ;|
```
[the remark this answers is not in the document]
BYK: Fixed in 36347416 — changed `args)` to `args|subcommand)` so the dynamic completion handler matches when `$state` is `"subcommand"` (via the `;|` fall-through from the `*` branch).

src/bin.ts
```diff
@@ -200,4 +225,14 @@ async function main(): Promise<void> {
...
+// Dispatch: check argv before any heavy imports
+// ---------------------------------------------------------------------------
+
+const args = process.argv.slice(2);
+
+if (args[0] === "__complete") {
+  runCompletion(args.slice(1));
+} else {
+  runCli(args);
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 2b4fd24c — added `.catch()` to both `runCompletion()` and `runCli()` calls. Completions silently swallow errors (return no results), CLI prints the error and sets exit code 1.

src/lib/complete.ts
```diff
@@ -0,0 +1,294 @@
...
+ */
+export async function getCompletions(
+  precedingWords: string[],
+  partial: string
+): Promise<Completion[]> {
+  // Build the command path from preceding words (e.g., "issue list")
+  const cmdPath =
+    precedingWords.length >= 2
+      ? `${precedingWords[0]} ${precedingWords[1]}`
+      : "";
```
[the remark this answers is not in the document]
BYK: Acknowledged but this is a non-issue in practice — Stricli handles global flags before they reach the completion engine. When the shell invokes `__complete`, it passes `COMP_WORDS[1:]` which doesn't include global flags that Stricli already consumed. The completion protocol only sees the command path tokens.

src/lib/db/completion-telemetry.ts
```diff
@@ -0,0 +1,79 @@
...
+    return rows.map((row) => ({
+      id: row.id,
+      commandPath: row.command_path,
+      durationMs: row.duration_ms,
+      resultCount: row.result_count,
+    }));
+  } catch {
+    return [];
+  }
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 5c881c1d — replaced the separate SELECT + DELETE with a single atomic `DELETE FROM completion_telemetry_queue RETURNING id, command_path, duration_ms, result_count`. No race window between read and delete.

PR #470, opened 2026-03-19
fix: accept nullable user fields in OAuth token response

### LINKED ISSUE #468: bug: auth login fails with 'Unexpected response from token endpoint' despite receiving valid token
### Environment

- sentry CLI v0.18.0 (installed via `curl https://cli.sentry.dev/install -fsS | bash`)
- macOS Sonoma 14.3
- Auth target: sentry.io (SaaS, not self-hosted)

### Steps to Reproduce

1. Install sentry CLI v0.18.0 via the install script
2. Run `sentry auth login`
3. Authorize in the browser when prompted

### Expected Result

CLI stores the token and confirms successful authentication.

### Actual Result

CLI receives a valid OAuth token response from the server but rejects it:

```
◐ Waiting for authorization...
✖ Authorization failed

 ERROR  Unexpected response from token endpoint
  {"access_token":"...","refresh_token":"...","expires_in":2591999,"expires_at":"2026-04-18T09:03:59.747189Z","token_type":"Bearer","scope":"event:read event:write member:read org:read project:admin project:read project:write team:read","user":{"id":"48168","name":null,"email":"..."}}
```

The token response contains all expected OAuth2 fields (`access_token`, `refresh_token`, `expires_in`, `token_type`, `scope`) yet the CLI treats it as an error. Reproduced twice in a row. Possibly the response schema validation is too strict or doesn't expect the `user` field.

BYK:
The Sentry API can return `null` for `user.name` in the OAuth token response. The Zod schema had `name: z.string()` which rejected null, causing `safeParse()` to fail and login to throw 'Unexpected response from token endpoint'.

## Changes

- Make `name` and `email` nullable in `TokenResponseSchema` (root cause)
- Make `name` nullish in `SentryUserSchema` for consistency
- Convert null → undefined at the UserInfo/LoginResult boundary in `interactive-login.ts`, `auth/login.ts`, and `auth/whoami.ts`
- Add focused schema tests for nullable user fields

Closes #468


src/lib/interactive-login.ts
```diff
@@ -31,6 +31,24 @@ export type LoginResult = {
...
+  email?: string | null;
+  username?: string | null;
+}): NonNullable<LoginResult["user"]> {
+  return {
+    id: user.id,
+    ...(user.name ? { name: user.name } : undefined),
+    ...(user.email ? { email: user.email } : undefined),
+    ...(user.username ? { username: user.username } : undefined),
+  };
+}
```
BYK: Oh boy, why not just:

```
return Object.fromEntries(Object.entries().filter([k, v]) => Boolean(v) || k === 'id'));
```

src/types/sentry.ts
```diff
@@ -215,7 +215,7 @@ export const SentryUserSchema = z
     id: z.string(),
     email: z.string().optional(),
     username: z.string().optional(),
-    name: z.string().optional(),
+    name: z.string().nullish(),
```
BYK, line 226: Should we change other stuff to `nullish()` here too?

Also, why are these types not coming from `@sentry/api` module?

src/lib/interactive-login.ts
BYK: Done — rewrote with `Object.fromEntries` + `Boolean(v)` filter as suggested.

src/types/sentry.ts
BYK: Done — made `email` and `username` also `.nullish()` for consistency.

Re `@sentry/api`: the `/auth/` endpoint is undocumented and not in the OpenAPI spec, so `@sentry/api` doesn't export types for it. Added a JSDoc note explaining this.

src/lib/interactive-login.ts
```diff
@@ -31,6 +31,24 @@ export type LoginResult = {
...
+  name?: string | null;
+  email?: string | null;
+  username?: string | null;
+}): NonNullable<LoginResult["user"]> {
+  return {
+    id: user.id,
+    ...(user.name ? { name: user.name } : undefined),
+    ...(user.email ? { email: user.email } : undefined),
+    ...(user.username ? { username: user.username } : undefined),
+  };
```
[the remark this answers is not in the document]
BYK: Intentional — an empty-string name or email is semantically absent for display purposes (`formatUserIdentity` already treats them as falsy). The JSDoc on `toLoginUser` now documents this explicitly.

PR #486, opened 2026-03-19
perf(resolve): carry project data through resolution to eliminate redundant getProject calls

BYK:
### Summary

`resolveProjectBySlug()` and `resolveOrgProjectTarget()` call `findProjectsBySlug()` which fetches the full `SentryProject` via `getProject()` per org — but both discard that data, returning only `{ org, project }` slugs. Three commands then call `getProject()` a second time.

This is the same class of redundancy that #473 fixed for the issue short-ID path, generalized to all project-search resolution consumers.

### Changes

**Core types** (`src/lib/resolve-target.ts`):
- `resolveProjectBySlug()` now returns `projectData: SentryProject` alongside slugs
- `ResolvedOrgProject` gains optional `projectData?` (populated in project-search path)
- `ResolvedTarget` gains optional `projectData?` for downstream consumers

**Fixed callers** (each saves ~500-800ms in the project-search path):
- `dashboard/create.ts` — uses `projectData.id` instead of `fetchProjectId()`
- `project/delete.ts` — uses `projectData` instead of re-calling `getProject()`
- `project/view.ts` — skips `getProject()` when `projectData` already available

Non-redundant callers (`event/view`, `log/view`, `init`, `trace/span/log list`) are unchanged — they only destructure `{ org, project }` for scoping.

### Test plan

- `bun run typecheck` passes
- `bun run lint` passes
- All affected test suites pass (298 tests across 12 files)
- Tests updated: `toEqual` → `toMatchObject` where the new `projectData` field is present

### DIFF
--- a/src/commands/event/view.ts
+++ b/src/commands/event/view.ts
@@ -229,7 +229,8 @@ export async function resolveEventTarget(
         `sentry event view <org>/${parsed.projectSlug} ${eventId}`
       );
       return {
-        ...resolved,
+        org: resolved.org,
+        project: resolved.project,
         orgDisplay: resolved.org,
         projectDisplay: resolved.project,
       };


BYK:
Good catch. Fixed in 22cf036 — both  and `resolveOrgProjectTarget()` now destructure out `orgSlug` from the `ProjectWithOrg` before assigning to `projectData`, so it's a clean `SentryProject` and won't leak into JSON output when callers spread it.

BYK:
Fixed in be6735f — replaced `...resolved` spread with explicit field mapping in `event/view.ts`, consistent with how `project/view.ts` was handled.

PR #490, opened 2026-03-20
perf: HTTP latency optimizations — diagnostics, cache warming, concurrency limits

BYK:
## Summary

Four targeted optimizations based on investigation showing Bun already uses HTTP/2 multiplexing and keep-alive (transport is not the bottleneck), but sequential resolution chains and unbounded fan-outs cause observable slowness.

### 1. HTTP timing diagnostics (`sentry-client.ts`)
Debug-level logging in the authenticated fetch showing method, URL path, status, timing (ms), cache hits, and retry attempts. Visible with `SENTRY_LOG_LEVEL=debug` or `--verbose`.

### 2. Post-login cache warming (`auth/login.ts`)
After successful login, fire-and-forget `listOrganizationsUncached()` to pre-populate the org + region SQLite cache, eliminating the ~800ms cold-start on the first command.

### 3. Concurrency limits on org fan-outs
Add `p-limit(5)` to all unbounded `Promise.all()` patterns that fan out across organizations or regions:
- `findProjectsBySlug()`, `findProjectsByPattern()`, `findProjectByDsnKey()` in `api/projects.ts`
- `findEventAcrossOrgs()` in `api/events.ts`
- `resolveProjectSearch()` in `commands/issue/utils.ts`

### 4. Faster org resolution for normal slugs (`region.ts`)
`resolveEffectiveOrg()` now uses `resolveOrgRegion()` (1 API call) for normal slugs instead of `listOrganizationsUncached()` (1+N requests). The expensive fan-out is reserved for DSN numeric IDs (`oNNNNN`) that need ID→slug mapping.

## Test plan

- `bun run typecheck` — clean
- `bun run lint` — clean
- 293 tests across affected modules pass

### DIFF
--- a/src/commands/issue/utils.ts
+++ b/src/commands/issue/utils.ts
@@ -4,6 +4,7 @@
  * Common functionality used by explain, plan, view, and other issue commands.
  */
 
+import pLimit from "p-limit";
 import {
   findProjectsBySlug,
   getAutofixState,
@@ -13,6 +14,7 @@ import {
   type IssueSort,
   listIssuesPaginated,
   listOrganizations,
+  ORG_FANOUT_CONCURRENCY,
   triggerRootCauseAnalysis,
   tryGetIssueByShortId,
 } from "../../lib/api-client.js";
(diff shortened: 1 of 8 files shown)


src/commands/issue/utils.ts
```diff
@@ -238,12 +239,16 @@ async function resolveProjectSearch(
   // 3. Fast path: try resolving the short ID directly across all orgs.
   //    The shortid endpoint validates both project existence and issue existence
   //    in a single call, eliminating the separate getProject() round-trip.
+  //    Concurrency-limited to avoid overwhelming the API for enterprise users.
   const fullShortId = expandToFullShortId(suffix, projectSlug);
   const orgs = await listOrganizations();
 
+  const limit = pLimit(5);
```
BYK: This "5 concurrent requests" seems common enough so maybe abstract that number `5` into a global constant?

PR #560, opened 2026-03-25
feat(dashboard): add pagination and glob filtering to dashboard list

BYK:
## Summary

Add cursor-based pagination and client-side glob filtering to `dashboard list`, plus bare-org slug support across all dashboard commands.

### API Layer
- Replace `listDashboards()` with `listDashboardsPaginated()` returning `PaginatedResponse` with cursor support via Link header parsing

### Dashboard List Command
- Switch to `buildListCommand` with auto-injected `--cursor`/`-c` and `--fresh`/`-f` flags
- Add positional name argument with title filtering: bare words, glob patterns (`'Error*'`, `'*API*'`), and multi-word unquoted titles all work
- Case-insensitive matching via `Bun.Glob`
- Support `org/project/name` as single arg (splits on second `/`)
- Extended cursor encoding (`serverCursor|dashboardId`) for mid-page resume when filtering fills `--limit` partway through a server page
- Auto-paginate when `--limit` exceeds `API_MAX_PER_PAGE` (100)
- Fuzzy suggestions (via `fuzzyMatch()`) when filter matches nothing
- JSON output uses `{ data, hasMore, nextCursor? }` paginated envelope

### Dashboard Resolve
- Bare org slug normalization: `sentry dashboard list my-org 'Error*'` and `sentry dashboard view my-org 12345` now work (auto-appends `/` when first arg has no `/`)
- `resolveDashboardId()` paginates through all dashboards and uses `fuzzyMatch()` for error suggestions instead of overwhelming full listings

### AGENTS.md
- Codify auto-pagination rule for `--limit > API_MAX_PER_PAGE`
- Add fuzzy suggestion guidance for resolution errors

## Examples
```bash
sentry dashboard list                         # auto-detect org
sentry dashboard list my-org/                 # explicit org
sentry dashboard list 'Error*'               # glob filter
sentry dashboard list my-org '*API*'          # bare org + filter
sentry dashboard list my-org/cli CLI          # org/project + filter
sentry dashboard list sentry/cli/CLI          # org/project/name single arg
sentry dashboard list my-org CLI Health       # multi-word unquoted filter
sentry dashboard list -c last                 # next page
sentry dashboard list --json                  # paginated JSON envelope
sentry dashboard view my-org 'My Dashboard'  # bare org in view
```

## Test Plan
- 81 tests across `list.test.ts` and `resolve.test.ts` covering pagination, glob filtering, cursor encoding/decoding, bare-org normalization, multi-word args, fuzzy suggestions, and `org/project/name` splitting

### DIFF
--- a/src/commands/dashboard/list.ts
+++ b/src/commands/dashboard/list.ts
@@ -119,9 +119,16 @@ export function decodeCursor(cursor: string): {
 function formatDashboardListHuman(result: DashboardListResult): string {
   if (result.dashboards.length === 0) {
     if (result.titleFilter && result.allTitles && result.allTitles.length > 0) {
-      const similar = fuzzyMatch(result.titleFilter, result.allTitles, {
-        maxResults: 5,
-      });
+      // Strip glob metacharacters before fuzzy matching so '*', '?', '['
+      // don't inflate Levenshtein distances (e.g. "Error*" → "Error").
+      const stripped = result.titleFilter.replace(/[*?[\]]/g, "");
+      const similar = fuzzyMatch(
+        stripped || result.titleFilter,
+        result.allTitles,
+        {
+          maxResults: 5,
+        }
+      );
       if (similar.length > 0) {
         return `No dashboards matching '${result.titleFilter}'. Did you mean:\n${similar.map((t) => `  • ${t}`).join("\n")}`;
       }


plugins/sentry-cli/skills/sentry-cli/references/dashboards.md
```diff
@@ -11,14 +11,15 @@ requires:
 
 Manage Sentry dashboards
 
-### `sentry dashboard list <org/project>`
+### `sentry dashboard list <args...>`
```
BYK: We should have something more descriptive than `args...` here. Something like `title|org|grep`

src/commands/dashboard/resolve.ts
```diff
@@ -100,17 +108,106 @@ export function parseDashboardPositionalArgs(args: string[]): {
...
+ * doesn't apply — `resolveOrgFromTarget` ignores the slug anyway.
+ * To specify an org, use `org/` or pass two args: `org 'filter'`.
+ *
+ * When two args are provided and the first is a bare slug (no `/`), it is
+ * normalized to `slug/` so `parseOrgProjectArg` treats it as an org-all target.
+ *
+ * @param args - Raw positional arguments
+ * @returns Target arg for org resolution and optional title filter glob
+ */
+export function parseDashboardListArgs(args: string[]): {
```
BYK, line 140: Could this be generalized and used by other commands too?

src/commands/dashboard/resolve.ts
```diff
@@ -100,17 +108,106 @@ export function parseDashboardPositionalArgs(args: string[]): {
...
+/**
+ * Split a slash-containing single arg into target and optional title filter.
+ *
+ * - `org/` or `org/project` (≤1 slash) → target only, no filter
+ * - `org/project/name` (2+ slashes) → target is `org/project`, filter is the rest
+ *
+ * This lets users type `sentry dashboard list my-org/my-project/CLI` as a
+ * single arg instead of requiring two separate args.
+ */
+function splitOrgProjectName(arg: string): {
```
BYK, line 182: Don't we have a shared helper for this already?

src/commands/dashboard/resolve.ts
```diff
@@ -124,27 +221,51 @@ export async function resolveDashboardId(
...
+    // Match by ID/slug first (e.g. "default-overview"), then fall back to title
+    const match =
+      data.find((d) => d.id.toLowerCase() === lowerRef) ??
+      data.find((d) => d.title.toLowerCase() === lowerRef);
+    if (match) {
+      return match.id;
+    }
+
+    for (const d of data) {
+      allTitles.push(d.title);
```
BYK, line 243: This entire logic also seems like it can be generalized and used by other `list` commands?

src/commands/dashboard/list.ts
```diff
@@ -79,33 +162,203 @@ function formatDashboardListHuman(result: DashboardListResult): string {
...
+    }
+    serverCursor = nextCursor;
+  }
+
+  return { dashboards: results, cursorToStore, allTitles };
+}
+
+// ---------------------------------------------------------------------------
+// Command
+// ---------------------------------------------------------------------------
```
BYK: Let's get rid of all these useless and noisy separator commmands

plugins/sentry-cli/skills/sentry-cli/references/dashboards.md
BYK: Fixed — added `placeholder: "org/title-filter"` to the array positional parameter. SKILL.md now generates `<org/title-filter...>` instead of `<args...>`.

src/commands/dashboard/resolve.ts
BYK: Good question. This function is dashboard-specific — it disambiguates single args between title filters (bare words, globs, spaces) and org/project targets, which is unique to dashboard list (other list commands don't have positional name filters). The multi-word join logic (`args.slice(1).join(" ")`) is also specific to dashboard titles. If another command gets a similar positional filter, we could extract a shared helper then, but for now the disambiguation rules are dashboard-specific.

src/commands/dashboard/resolve.ts
BYK: The existing `parseSlashSeparatedArg` splits on the *last* slash (designed for hex IDs like `org/project/abc123`). This function splits on the *second* slash because dashboard titles can contain anything — we need `org/project` as target and everything after as the name filter. Different split strategies for different use cases. Could be generalized into a shared helper with a configurable split strategy in a follow-up if other commands need it.

src/commands/dashboard/resolve.ts
BYK: Agreed the paginate-match-fuzzy pattern could be extracted. Each entity has different match logic though (dashboards match by ID + title, issues by short ID, etc.) so the generic version would need a configurable matcher. Worth doing as a follow-up when another command needs the same pattern.

src/commands/dashboard/list.ts
BYK: Removed all `// ---------------------------------------------------------------------------` separator dividers. Kept the simple section name comments (e.g. `// Command`) as they're useful navigational markers without being noisy.

src/commands/dashboard/list.ts
```diff
@@ -134,18 +377,75 @@ export const listCommand = buildCommand({
       return;
     }
 
-    const dashboards = await withProgress(
+    // Resolve pagination cursor (handles "last" magic value)
+    const contextKey = buildPaginationContextKey("dashboard", orgSlug, {
+      ...(titleFilter && { q: titleFilter }),
+    });
```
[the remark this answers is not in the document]
BYK: Fixed — lowercased the filter in `buildPaginationContextKey` to match the case-insensitive glob matching.

src/commands/dashboard/resolve.ts
```diff
@@ -124,27 +221,51 @@ export async function resolveDashboardId(
...
+      perPage: API_MAX_PER_PAGE,
+      cursor,
+    });
+    // Match by ID/slug first (e.g. "default-overview"), then fall back to title
+    const match =
+      data.find((d) => d.id.toLowerCase() === lowerRef) ??
+      data.find((d) => d.title.toLowerCase() === lowerRef);
+    if (match) {
+      return match.id;
+    }
```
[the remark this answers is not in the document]
BYK: The per-page ID→title search already prioritizes ID matches within each page via `find(id) ?? find(title)`. The cross-page scenario (ID on page 2, title on page 1) is theoretically possible but effectively impossible with Sentry dashboard IDs (numeric or `default-overview`) — a title match on page 1 would only shadow an ID match if a dashboard title exactly equals another dashboard's numeric ID. Accepting this as a negligible theoretical risk.

src/commands/dashboard/list.ts
```diff
@@ -48,6 +118,15 @@ type DashboardListResult = {
  */
 function formatDashboardListHuman(result: DashboardListResult): string {
   if (result.dashboards.length === 0) {
+    if (result.titleFilter && result.allTitles && result.allTitles.length > 0) {
+      const similar = fuzzyMatch(result.titleFilter, result.allTitles, {
+        maxResults: 5,
```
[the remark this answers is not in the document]
BYK: Fixed — glob metacharacters (`*?[]`) are now stripped before passing to `fuzzyMatch`, so `Error*` fuzzy-matches against `Error Rate` correctly.

src/commands/dashboard/list.ts
```diff
@@ -79,33 +165,198 @@ function formatDashboardListHuman(result: DashboardListResult): string {
...
+      }
+      break;
+    }
+
+    if (!nextCursor) {
+      cursorToStore = undefined;
+      break;
+    }
+    serverCursor = nextCursor;
+  }
```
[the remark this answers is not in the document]
BYK: The `MAX_PAGINATION_PAGES` (50) safety limit is the same one used by `listIssuesAllPages` — it's intentional to prevent runaway requests. An org with 5000+ dashboards hitting the 50-page limit with a selective glob is an extreme edge case. The behavior matches `issue list` and other paginated commands. Users can override via `SENTRY_MAX_PAGINATION_PAGES` env var if needed.

PR #611, opened 2026-03-31
feat(auth): enforce auth by default in buildCommand

betegon:
## Summary

All commands now require authentication by default. `buildCommand` checks `isAuthenticated()` before invoking the command function and throws `AuthError("not_authenticated")` if no token exists — this feeds into the existing auto-auth middleware that prompts login and retries.

Previously each command had to manually guard auth or rely on an API call failing deep in execution. `sentry init` in particular had no auth check at all and would proceed into the wizard flow before eventually failing.

## Changes

Added `auth?: boolean` to `buildCommand` options (defaults `true`). The check runs before the generator, so the auto-auth middleware in `cli.ts` catches it and kicks off the interactive login flow on unauthenticated runs.

Commands that intentionally work without a token opt out with `auth: false`:
- `auth login`, `auth logout`, `auth refresh`, `auth status`
- `help`, `schema`
- `cli fix`, `cli setup`, `cli upgrade`, `cli feedback`

With this in place, the redundant explicit `isAuthenticated()` checks in `auth whoami` and `auth token` were removed.

## Test Plan

- `sentry init` (unauthenticated) → should immediately prompt "Authentication required. Starting login flow..."
- `sentry auth login` (unauthenticated) → should work normally
- `sentry auth status` (unauthenticated) → should show "not authenticated" cleanly, no login prompt
- `sentry auth logout` (unauthenticated) → should return "Not currently authenticated."
- `SENTRY_AUTH_TOKEN=xxx sentry init` → no auth prompt, proceeds normally

### DIFF
--- a/src/commands/auth/token.ts
+++ b/src/commands/auth/token.ts
@@ -8,6 +8,7 @@
 import type { SentryContext } from "../../context.js";
 import { buildCommand } from "../../lib/command.js";
 import { getAuthToken } from "../../lib/db/auth.js";
+import { AuthError } from "../../lib/errors.js";
 import { CommandOutput } from "../../lib/formatters/output.js";
 
 export const tokenCommand = buildCommand({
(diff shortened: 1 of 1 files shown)


BYK (approved):
🚀 

The only concern I have is `sentry issues --help` commands. I'm guessing they will be fine but please double check before merging.

PR #623, opened 2026-04-01
feat(traces): expose custom span attributes and improve agent guidance

BYK:
## Summary

Addresses feedback from an AI agent field report about Sentry CLI usability. The critical gap was that custom span attributes (like `gen_ai.usage.input_tokens`, `pi.session.id`) were completely invisible via CLI — this PR fixes that across all three span/trace commands.

### Custom attribute access (critical)

- **`span view`**: Fetches full attributes from the `trace-items/{itemId}/` endpoint (same as Sentry frontend's span detail sidebar). All custom OTEL attributes now appear in both JSON and human output with no configuration needed.
- **`trace view`**: `--fields` values that aren't standard output fields are forwarded as `additional_attributes` query params to the trace detail API.
- **`span list`**: `--fields` now drives API field selection — unknown field names are requested as additional `field` params from the events API. Includes `gen_ai` group alias that expands to common AI observability fields.

### Agent guidance improvements

- Added `--since` as an alias for `--period` on all list commands (the most natural flag name for time filtering)
- Added "Quick Reference" section to agent guidance docs covering correct patterns for time filtering, org/project scoping, trace ID usage, and valid dataset names
- Regenerated SKILL.md

### Quality of life

- Filter zero-valued web vitals measurements from `trace view` JSON output (~40% size reduction for typical traces)
- Added `EVENTS_API_DATASETS` shared constant and documented valid dataset names in api command docs
- Widened `TraceSpan` type to include all fields the trace detail API actually returns (`measurements`, `additional_attributes`, `sdk_name`, `errors`, `occurrences`, etc.)

### DIFF
--- a/src/lib/formatters/json.ts
+++ b/src/lib/formatters/json.ts
@@ -10,17 +10,32 @@ import type { Writer } from "../../types/index.js";
 /**
  * Get a nested value from an object using a dot-notated path.
  *
+ * First checks for the path as a literal property name (e.g.,
+ * `"gen_ai.usage.input_tokens"` as a flat key), then falls back to
+ * dot-separated nested traversal (e.g., `"contexts.trace.traceId"`).
+ * This ensures custom span attributes with dotted names are found.
+ *
  * Returns `{ found: true, value }` when the path resolves, even if the
  * leaf value is `undefined` or `null`. Returns `{ found: false }` when
  * any intermediate segment is not an object (or is missing).
  *
  * @param obj - Source object to traverse
- * @param path - Dot-separated key path (e.g. `"contexts.trace.traceId"`)
+ * @param path - Key path: literal property name or dot-separated nesting
  */
 function getNestedValue(
   obj: unknown,
   path: string
 ): { found: true; value: unknown } | { found: false } {
+  if (obj === null || obj === undefined || typeof obj !== "object") {
+    return { found: false };
+  }
+
+  // Fast path: literal property name (handles dotted keys like "gen_ai.usage.input_tokens")
+  if (Object.hasOwn(obj, path)) {
+    return { found: true, value: (obj as Record<string, unknown>)[path] };
+  }
+
+  // Fall back to dot-separated nested traversal
   let current: unknown = obj;
   for (const segment of path.split(".")) {
     if (
@@ -41,15 +56,25 @@ function getNestedValue(
 /**
  * Set a nested value in an object, creating intermediate objects as needed.
  *
+ * When `literalKey` is true, the path is used as a literal property name
+ * (no dot splitting). This preserves dotted attribute names like
+ * `"gen_ai.usage.input_tokens"` as flat keys in the output.
+ *
  * @param target - Target object to write into (mutated in place)
- * @param path - Dot-separated key path
+ * @param path - Key path: literal name or dot-separated nesting
  * @param value - Value to set at the leaf
+ * @param literalKey - When true, skip dot splitting
  */
 function setNestedValue(
   target: Record<string, unknown>,
   path: string,
-  value: unknown
+  value: unknown,
+  literalKey = false
 ): void {
+  if (literalKey) {
+    target[path] = value;
+    return;
+  }
   const segments = path.split(".");
   let current: Record<string, unknown> = target;
   for (let i = 0; i < segments.length - 1; i++) {
@@ -104,7 +129,11 @@ export function filterFields<T>(data: T, fields: string[]): Partial<T> {
   for (const field of fields) {
     const lookup = getNestedValue(data, field);
     if (lookup.found) {
-      setNestedValue(result, field, lookup.value);
+      // Use literal key when the field name exists as a direct property
+      // (e.g., "gen_ai.usage.input_tokens" as a flat key)
+      const isLiteral =
+        typeof data === "object" && data !== null && Object.hasOwn(data, field);
+      setNestedValue(result, field, lookup.value, isLiteral);
     }
   }
 


src/commands/span/list.ts
```diff
@@ -57,6 +57,63 @@ type ListFlags = {
   readonly fields?: string[];
 };
 
+/** Maps output field names to their API field equivalents in SPAN_FIELDS */
+const OUTPUT_TO_API_FIELD: Record<string, string> = {
```
BYK: Why do we need this mapping? Remove if we can, document in comments with the reason if we cannot.

src/commands/span/list.ts
```diff
@@ -57,6 +57,63 @@ type ListFlags = {
...
+    } else {
+      expanded.push(f);
+    }
+  }
+
+  const knownOutputNames = new Set(Object.keys(OUTPUT_TO_API_FIELD));
+  const knownApiNames = new Set(Object.values(OUTPUT_TO_API_FIELD));
+  const extra = expanded.filter(
+    (f) => !(knownOutputNames.has(f) || knownApiNames.has(f))
+  );
```
BYK: Can we not make `expanded` a `Set` too and just perform set operations here? If we don't have native set methods available in latest Node or Bun, we can use this library here: https://npmx.dev/package/superset (prefer native methods though)

src/commands/span/list.ts
```diff
@@ -57,6 +57,63 @@ type ListFlags = {
...
+  return extra.length > 0 ? [...new Set(extra)] : undefined;
```
BYK: `Array.from(set)` should be way more efficient than spreading.

src/commands/span/view.ts
```diff
@@ -38,6 +41,73 @@ import {
...
+  "span.duration",
+  "span.self_time",
+  "span.op",
+  "span.name",
+  "span.description",
+  "span.category",
+  "parent_span",
+  "transaction",
+  "transaction.op",
+  "transaction.event_id",
```
BYK: How are these low value???

src/commands/span/view.ts
```diff
@@ -38,6 +41,73 @@ import {
...
+/**
+ * Convert a trace-items attribute array into a key-value dict,
+ * optionally filtering out internal attributes.
+ */
+function attributesToDict(
+  attributes: TraceItemDetail["attributes"],
+  filterInternal: boolean
+): Record<string, unknown> {
+  const dict: Record<string, unknown> = {};
+  for (const attr of attributes) {
```
BYK: Can we not use `Object.fromEntries()` here somehow?

src/commands/span/view.ts
```diff
@@ -216,12 +316,49 @@ function formatSpanViewHuman(data: SpanViewData): string {
...
+  org: string,
+  project: string,
+  traceId: string
+): Promise<Map<string, TraceItemDetail>> {
+  const limit = pLimit(SPAN_DETAIL_CONCURRENCY);
+  const details = new Map<string, TraceItemDetail>();
+
+  await Promise.all(
+    results.map((r) =>
+      limit(async () => {
```
BYK: I think `pLimit` has a built-in `map` helper?

src/lib/api/traces.ts
```diff
@@ -197,6 +257,10 @@ export async function listSpans(
   const projectFilter = isNumericProject ? "" : `project:${projectSlug}`;
   const fullQuery = [projectFilter, options.query].filter(Boolean).join(" ");
 
+  const fields = options.extraFields?.length
+    ? [...SPAN_FIELDS, ...options.extraFields]
```
BYK: Why not `Array.from()` again or `SPAN_FIELDS.concat()` or something?

src/lib/formatters/trace.ts
```diff
@@ -533,3 +589,69 @@ export function formatSpanDetails(
...
 }
+
+/** Max custom attributes to show in human output before truncation */
+const MAX_DISPLAY_ATTRS = 10;
+
+/**
+ * Attribute names that duplicate fields already shown in `buildSpanKvRows`
+ * or are EAP storage internals with no diagnostic value.
+ */
+const HIDDEN_HUMAN_ATTRS = new Set([
```
BYK: Is this list repeated across traces and spans or is it me?

src/types/sentry.ts
```diff
@@ -432,6 +432,28 @@ export type TraceSpan = {
   event_id?: string;
   /** Nested child spans */
   children?: TraceSpan[];
+  /** Span name (often same as op) */
```
BYK, line 435: Why don't we have these types from `@sentry/api`? While you are at it, make sure we are on the latest version of `@sentry/api` to not miss any new types.

src/commands/span/list.ts
BYK: Replaced with a flat `KNOWN_SPAN_FIELDS` set that lists both API names and their output aliases. Much simpler — just a membership check, no mapping needed.

src/commands/span/list.ts
BYK: Done — `expanded` is now a `Set` from the start, and filtering uses `Set.delete()` in a loop over `KNOWN_SPAN_FIELDS`.

src/commands/span/list.ts
BYK: Fixed — using `Array.from(expanded)` now.

src/commands/span/view.ts
BYK: Fair point — these are not low-value, they're redundant (already shown in the standard span fields above). Renamed the set to `REDUNDANT_DETAIL_ATTRS` with category comments: "timing/storage internals" vs "already shown in standard span fields". The `user.*` entries were also removed from this set since they're user context, not duplicates.

src/commands/span/view.ts
BYK: Done — rewrote to use `Object.fromEntries()` with a `.filter().map()` chain.

src/commands/span/view.ts
BYK: Yes! Switched to `limit.map(results, async (r) => { ... })`.

src/lib/api/traces.ts
BYK: Fixed — using `SPAN_FIELDS.concat(options.extraFields)` now.

src/lib/formatters/trace.ts
BYK: It was! Extracted to a single `REDUNDANT_DETAIL_ATTRS` constant in `src/lib/api/traces.ts`, shared by both `span/view.ts` and `formatters/trace.ts`.

src/types/sentry.ts
BYK: Checked — `@sentry/api` (v0.54.0, latest is v0.91.0) does not export any `TraceSpan` type. The `/trace/{traceId}/` endpoint is private (not in the public OpenAPI schema), so the `TraceSpan` type is hand-written by necessity. Same for `TraceItemDetail` — the `/projects/{org}/{project}/trace-items/{itemId}/` endpoint is also private. This is documented in the AGENTS.md lore entry.

Updating `@sentry/api` to latest is a good idea but orthogonal to this PR — will file separately.

src/lib/list-command.ts
```diff
@@ -299,7 +299,7 @@ export const LIST_PERIOD_FLAG = {
  * Exported separately from `LIST_BASE_ALIASES` because not all list commands
  * need a period flag, and some commands already use `-t` for other purposes.
  */
-export const PERIOD_ALIASES = { t: "period" } as const;
+export const PERIOD_ALIASES = { t: "period", since: "period" } as const;
```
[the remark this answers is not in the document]
BYK: Good catch. Reverted — Stricli only supports single-character aliases (`AvailableAlias = LowercaseLetter | UppercaseLetter`). Updated docs to reference `--period` and `-t` only.

src/lib/api/datasets.ts
```diff
@@ -0,0 +1,18 @@
...
+export const EVENTS_API_DATASETS = [
+  "spans",
+  "transactions",
+  "logs",
+  "errors",
+  "discover",
+] as const;
+
+/** Type for Events API dataset values */
+export type EventsApiDataset = (typeof EVENTS_API_DATASETS)[number];
```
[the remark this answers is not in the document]
BYK: Intentional — this is a shared constant + type for documentation and future validation use (the plan calls for commands that accept dataset params to validate against it). Similar pattern to other constants in the codebase that exist for reference before their first consumer lands.

src/lib/api/traces.ts
```diff
@@ -205,7 +309,7 @@ export async function listSpans(
     {
       params: {
         dataset: "spans",
-        field: SPAN_FIELDS,
+        field: fields,
```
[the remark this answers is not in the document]
BYK: False positive — `SpanListItemSchema` already uses `.passthrough()` (line 796 of `src/types/sentry.ts`). Extra fields from the API survive Zod parsing.

src/lib/api/datasets.ts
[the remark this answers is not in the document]
BYK: Removed in 054af1f7.

src/commands/span/list.ts
```diff
@@ -311,14 +390,17 @@ async function handleTraceMode(
...
         statsPeriod: flags.period,
+        extraFields: extraApiFields,
       })
   );
 
   // Update pagination state (handles both advance and truncation)
   advancePaginationState(PAGINATION_KEY, contextKey, direction, nextCursor);
   const hasPrev = hasPreviousPage(PAGINATION_KEY, contextKey);
 
-  const flatSpans = spanItems.map(spanListItemToFlatSpan);
```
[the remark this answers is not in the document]
BYK: Good catch. Fixed in d92865cb — `getNestedValue` now checks for the path as a literal property name first (handles flat dotted keys like `gen_ai.usage.input_tokens`), then falls back to dot-separated traversal. `setNestedValue` also preserves literal keys in the output.

PR #721, opened 2026-04-13
feat(cli): add `sentry cli defaults` command for persistent settings

### LINKED ISSUE #304: feat: support setting and using a default organization
## Summary

Currently, when org resolution fails, the error hint says `"Set a default: sentry org view <org>"` — but there's no mechanism to actually set a default org.

## Proposal

- Implement `sentry org default <org>` (or similar) to persist a default org slug, and use it as a fallback when no org is specified or detected from DSN.
- Update the error hint in `src/commands/project/create.ts:319` to reference the correct command once implemented.

## Context

Reference:[ PR review comment](https://github.com/getsentry/cli/pull/237#discussion_r2840143148) from @BYK suggesting this as a follow-up feature.

BYK:
## Summary

Implements `sentry cli defaults` — a command for viewing and managing persistent CLI defaults. Closes #304.

- **Show defaults**: `sentry cli defaults` displays all current settings
- **Set defaults**: `sentry cli defaults org my-org`, `sentry cli defaults telemetry off`, etc.
- **Clear specific**: `sentry cli defaults org --clear`
- **Clear all**: `sentry cli defaults --clear --yes`

Supported keys: `org`, `project`, `telemetry` (on/off/yes/no/true/false/1/0), `url` (for self-hosted).

## Storage consolidation

The `defaults` single-row table was never written by production code. This PR:
- Migrates all defaults to the `metadata` KV table (keys: `defaults.org`, `defaults.project`, `defaults.telemetry`, `defaults.url`)
- Adds schema migration 13 that copies any existing data and drops the table
- Updates the JSON migration to write to `metadata` instead of `defaults`

Public getters (`getDefaultOrganization()`, `getDefaultProject()`) keep the same signatures — all consumers work unchanged.

## Telemetry preference

Adds persistent telemetry opt-out with a 4-level priority chain:
1. `SENTRY_CLI_NO_TELEMETRY=1` env var (highest)
2. `DO_NOT_TRACK=1` env var ([consoledonottrack.com](https://consoledonottrack.com))
3. SQLite persistent preference via `sentry cli defaults telemetry off`
4. Default: enabled

## URL default

Adds persistent URL default for self-hosted instances, applied in `preloadProjectContext()` after `.sentryclirc` env shim using the same `env.SENTRY_URL` mechanism.

## New utility

`parseBoolValue()` in `src/lib/parse-bool.ts` — human-friendly boolean parser adapted from Sentry JS SDK's `envToBool`. Handles on/off, yes/no, true/false, 1/0, t/f, y/n.

## Tests

- Property-based tests for `parseBoolValue` (8 properties, 50 runs each)
- Unit tests for defaults storage layer and `isTelemetryEnabled()` priority chain (26 tests)
- Updated 4 existing test files that used the removed `setDefaults()` function


src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,373 @@
...
+    newValue: string | boolean | null;
+  };
+};
+
+/**
+ * Compute the effective telemetry state considering env var overrides.
+ * Used for display in show mode to help users understand when env vars
+ * override their stored preference.
+ */
+function computeTelemetryEffective(): DefaultsResult["telemetryEffective"] {
```
BYK: Shouldn't this live in a generic helper library for `Sentry.init()` system to use? I'm also concerned about how we track auto-completions. I'm assuming they'd still honor this and drop the temp entries from the DB without sending them?

src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,373 @@
...
+  if (pref !== undefined) {
+    return { enabled: pref, source: "preference" };
+  }
+  return { enabled: true, source: "default" };
+}
+
+/**
+ * Read the current value of a single default for display or change tracking.
+ */
+function getCurrentValue(key: DefaultKey): string | boolean | null {
```
BYK: I feel like a map of key to handlers would be easier to maintain compared to a big switch statement?

src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,373 @@
...
+  "project",
+  "telemetry",
+  "url",
+]);
+
+/** Canonical key name (normalize "organization" → "org" for display) */
+type DefaultKey = "organization" | "project" | "telemetry" | "url";
+
+/** Map user-facing key names to canonical DefaultsState keys */
+function normalizeKey(key: string): DefaultKey | null {
```
BYK: Should we instead have a global `ALIASES` const was a map `org` -> `organization` and just use `key.toLowerCase()` and that map instead of a switch?

src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,373 @@
...
+};
+
+/**
+ * Compute the effective telemetry state considering env var overrides.
+ * Used for display in show mode to help users understand when env vars
+ * override their stored preference.
+ */
+function computeTelemetryEffective(): DefaultsResult["telemetryEffective"] {
+  const env = getEnv();
+  if (env.SENTRY_CLI_NO_TELEMETRY === "1") {
```
BYK: These ENV variable names should probably be in some constant?

src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,373 @@
...
+      return getDefaultUrl() ?? null;
+    default:
+      return null;
+  }
+}
+
+/**
+ * Set a default value. Validates and stores the value.
+ */
+function setDefault(key: DefaultKey, value: string): void {
```
BYK: Same as above: `name` -> `handler` pattern rather than a switch?

src/commands/cli/defaults.ts
BYK: `isTelemetryEnabled()` already lives in `src/lib/telemetry.ts` (the generic helper) which `withTelemetry()` calls. `computeTelemetryEffective()` stays in the command file because it is display-only — it computes the *source* string for the human formatter, not just the boolean.

Re completions: they honor this correctly. The completion fast-path sets `SENTRY_CLI_NO_TELEMETRY=1` (highest priority in `isTelemetryEnabled()`). Deferred queue entries are only drained when `enabled=true` in `withTelemetry()`, so if the persistent preference is off, queued entries are never sent.

src/commands/cli/defaults.ts
BYK: Done — replaced all three switch functions (`getCurrentValue`, `setDefault`, `clearDefault`) with a single `DEFAULTS_REGISTRY: Record<DefaultKey, DefaultHandler>` map. Each key maps to `{ get, set, clear }` handlers. The command dispatch just does `handler.get()` / `handler.set(value)` / `handler.clear()`.

src/commands/cli/defaults.ts
BYK: Done — replaced with a `KEY_ALIASES: Record<string, DefaultKey>` map. `normalizeKey()` is now a one-liner: `KEY_ALIASES[key.toLowerCase()] ?? null`.

src/commands/cli/defaults.ts
BYK: Done — extracted `TELEMETRY_ENV_VAR` and `DO_NOT_TRACK_ENV_VAR` as exported constants in `src/lib/telemetry.ts`. Used in both `isTelemetryEnabled()` and `computeTelemetryEffective()`. The formatter also simplified to just check `source.startsWith("env:")` and slice the env var name.

src/commands/cli/defaults.ts
BYK: Done — same `DEFAULTS_REGISTRY` handler map covers this too. See reply on the other handler map thread.

src/commands/cli/defaults.ts
```diff
@@ -0,0 +1,334 @@
...
+const DEFAULTS_REGISTRY: Record<DefaultKey, DefaultHandler> = {
+  organization: {
+    get: getDefaultOrganization,
+    set: setDefaultOrganization,
+    clear: () => setDefaultOrganization(null),
+  },
+  project: {
+    get: getDefaultProject,
+    set: setDefaultProject,
+    clear: () => setDefaultProject(null),
```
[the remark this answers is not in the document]
BYK: Good catch. Added a `validateSlug()` guard in the `organization` and `project` set handlers that rejects empty/whitespace-only strings with a `ValidationError`. The `url` handler already had this via `normalizeUrl`, and `telemetry` via `parseBoolValue`.

src/lib/formatters/human.ts
```diff
@@ -2405,3 +2405,90 @@ export function formatWidgetEdited(result: {
...
+ * Describe the effective telemetry source for display.
+ * Shows when an env var overrides the stored preference.
+ */
+function telemetryOverrideNote(
+  effective: DefaultsResult["telemetryEffective"]
+): string {
+  if (!effective?.source.startsWith("env:")) {
+    return "";
+  }
+  const envVar = effective.source.slice(4);
```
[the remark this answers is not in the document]
BYK: Fixed — both `.slice(4)` calls replaced with `.slice("env:".length)` to match the codebase convention.

PR #723, opened 2026-04-13
fix(upgrade): detect npm install method from node_modules path

BYK:
## Summary

- Add `detectPackageManagerFromPath()` — a fast path-based check that detects npm/pnpm installation by examining `process.argv[1]` for a `node_modules` segment
- Wire it into the detection cascade after Homebrew (`/Cellar/`) and before the DB lookup, mirroring the same authoritative-override pattern
- Persists the detected method to SQLite so subsequent runs skip even this cheap check

## Problem

When the CLI is installed via npm and run on Windows (especially via NVM/Laravel Herd), spawning `npm`, `pnpm`, `bun`, `yarn` to detect the install method fails with ENOENT because `.cmd` files aren't found by `spawn()` without `shell: true`. This left the method as `"unknown"`, blocking `sentry cli upgrade` entirely — even though the CLI was clearly running from a `node_modules` directory.

## Approach

Rather than fixing the Windows spawn issue (which would still be slow and fragile), add a fast path check that examines the script path for a `node_modules` segment — the same pattern used by the Homebrew `/Cellar/` check. Cross-platform (splits on both `/` and `\`), handles Windows 8.3 short paths, and distinguishes pnpm via its `.pnpm` directory.

Fixes CLI-Y1


src/lib/upgrade.ts
```diff
@@ -68,6 +68,9 @@ const NPM_REGISTRY_URL = "https://registry.npmjs.org/sentry";
 /** Regex to strip 'v' prefix from version strings */
 export const VERSION_PREFIX_REGEX = /^v/;
 
+/** Splits a file path on both forward slashes and backslashes (Unix + Windows) */
+const PATH_SEPARATOR_RE = /[\\/]/;
```
BYK: Why not use `path.sep`?

src/lib/upgrade.ts
```diff
@@ -200,6 +203,44 @@ function isHomebrewInstall(): boolean {
...
+  }
+
+  // pnpm uses a distinctive .pnpm directory inside node_modules
+  if (segments.includes(".pnpm")) {
+    return "pnpm";
+  }
+
+  // Default to npm for other node_modules installations
+  // (npm, yarn classic, and bun global all use the same flat layout)
+  return "npm";
```
BYK, line 241: Shall we not try to detect `bun` at least?

BYK:
Addressed both review comments in 387e8016:

1. **`path.sep` instead of regex**: Replaced `PATH_SEPARATOR_RE` with `scriptPath.split(sep)` — `sep` was already imported from `node:path`. Tests now use `join()` for platform-native path construction.

2. **Bun detection**: Added `.bun` segment check — bun global installs live under `~/.bun/install/global/node_modules/`. Detection order is now: `.pnpm` → pnpm, `.bun` → bun, default → npm.

PR #747, opened 2026-04-14
feat(setup): install agent skills for detected roots

betegon:
## Summary

This updates the existing review branch with the smaller change built from `main` instead of the earlier unconditional `.agents` installer. It keeps the Claude-style flow, adds shared `~/.agents` support only when that root already exists, and keeps setup output aligned with the file that was actually created.

## Changes

- detect existing `~/.agents` and `~/.claude` roots instead of creating a top-level shared root
- reuse the existing skill path helper for both targets and keep each install independent
- update setup and unit tests for no-agent silence, detected `~/.agents` installs, and fresh-path reporting

## Test Plan

- [x] `bun test test/lib/agent-skills.test.ts`
- [x] `bun test test/commands/cli/setup.test.ts`
- [x] `bun run typecheck`
- [x] `bun run lint` (passes with an existing unrelated warning in `src/lib/formatters/markdown.ts`)

### DIFF
--- a/src/lib/agent-skills.ts
+++ b/src/lib/agent-skills.ts
@@ -93,8 +93,7 @@ async function writeSkillFiles(
  *
  * Checks supported roots and writes the embedded skill files to each detected
  * location. The installer never creates top-level agent roots. Their presence
- * is the detection signal that the user already has a compatible agent
- * installed, matching the original Claude-only skip behavior.
+ * is the detection signal that the user already has a compatible agent installed.
  *
  * If any target is freshly created, the returned `path` points to that new
  * installation so setup output matches the file that was actually added.
--- a/src/lib/agent-skills.ts
+++ b/src/lib/agent-skills.ts
@@ -1,8 +1,8 @@
 /**
  * Agent skill installation for AI coding assistants.
  *
- * Detects supported AI coding agents (currently Claude Code) and installs
- * the Sentry CLI skill files so the agent can use CLI commands effectively.
+ * Detects supported agent roots and installs the embedded Sentry CLI
+ * skill files without creating new top-level agent directories.
  *
  * Skill file contents are embedded at build time (via a generated module
  * produced by script/generate-skill.ts), so no network fetch is needed.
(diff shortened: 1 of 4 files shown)


src/lib/agent-skills.ts
```diff
@@ -34,49 +38,43 @@ export function detectClaudeCode(homeDir: string): boolean {
...
+ * @param skillDir - Directory to write skill files into (e.g. ~/.agents/skills/sentry-cli)
+ * @param parentDir - Parent directory to check for writability (e.g. ~/.agents or ~/.claude)
+ * @param createParent - Whether to create the parent directory if it doesn't exist
  */
-export async function installAgentSkills(
-  homeDir: string
+async function writeSkillFiles(
+  skillDir: string,
+  parentDir: string,
+  createParent: boolean
```
BYK: We should not create the parent at all as it should be there if any of these agents are installed?

src/lib/agent-skills.ts
```diff
@@ -105,3 +103,53 @@ export async function installAgentSkills(
...
+ * @returns Location info if installed to at least one target, null otherwise
+ */
+export async function installAgentSkills(
+  homeDir: string
+): Promise<AgentSkillLocation | null> {
+  let agentsResult: AgentSkillLocation | null = null;
+  let claudeResult: AgentSkillLocation | null = null;
+
+  // 1. Always install to ~/.agents/skills/sentry-cli/ (cross-agent standard)
+  agentsResult = await writeSkillFiles(
```
BYK: This should not be unconditional still -- we should try to see if there is at least one agent installed?

src/lib/agent-skills.ts
```diff
@@ -105,3 +103,53 @@ export async function installAgentSkills(
...
+    join(homeDir, ".agents", "skills", "sentry-cli"),
```
BYK: Logic here and for Claude is repeated. It was nicely abstracted away in `getSkillInstallPath` in the previous version of the code. Can use similar tidy up.

src/lib/agent-skills.ts
```diff
@@ -34,49 +38,43 @@ export function detectClaudeCode(homeDir: string): boolean {
...
+  if (!existsSync(parentDir)) {
+    if (!createParent) {
+      return null;
+    }
+    try {
+      mkdirSync(parentDir, { recursive: true, mode: 0o755 });
+    } catch {
+      return null;
+    }
   }
```
BYK: I'd say this logic should be (if we decide to keep the `createParent` argument):

```
if (createParent) { await mkdir(parentDir, {recursive: true, mode: 0o755}); }
```

the `accessSync` below would catch the rest. Speaking of which, I think it can also be converted to async.

BYK (approved):
Love it!

We may wanna consider making the `.claude` one a symlink if both exists but that's a very minor and hard-to-get optimization. We can consider if there are more targets in the future (hope not! :D)

PR #748, opened 2026-04-14
feat(setup): install agent skills for detected roots

betegon:
## Summary

This follows up on #747 with a smaller change built from `main` instead of carrying forward the unconditional `.agents` installer. It keeps the existing Claude-only shape, adds shared `~/.agents` support only when that root already exists, and keeps setup output aligned with the file that was actually created.

## Changes

- detect existing `~/.agents` and `~/.claude` roots instead of creating a top-level shared root
- reuse the existing skill path helper for both targets and keep each install independent
- update setup and unit tests for no-agent silence, detected `~/.agents` installs, and fresh-path reporting

## Test Plan

- [x] `bun test test/lib/agent-skills.test.ts`
- [x] `bun test test/commands/cli/setup.test.ts`
- [x] `bun run typecheck`
- [x] `bun run lint` (passes with an existing unrelated warning in `src/lib/formatters/markdown.ts`)

## References

- Supersedes #747

### DIFF
--- a/src/commands/cli/setup.ts
+++ b/src/commands/cli/setup.ts
@@ -280,8 +280,9 @@ async function handleCompletions(
 /**
  * Handle agent skill installation for AI coding assistants.
  *
- * Detects supported agents (currently Claude Code) and installs the
- * version-pinned skill file. Silent when no agent is detected.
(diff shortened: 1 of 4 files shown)


betegon:
Moved this implementation back to #747 to keep the existing review context and comments in one place.

PR #757, opened 2026-04-15
fix(arg-parsing): normalize spaces in slugs and trim whitespace in issue IDs (CLI-14M, CLI-16M)

BYK:
## Summary

- Extend `normalizeSlug()` to convert spaces → dashes with lowercasing and consecutive dash collapsing, matching how Sentry generates slugs from display names (e.g., `"My Project"` → `"my-project"`)
- Reorder parsing flow so normalization runs before `validateResourceId()` — spaces become dashes before the forbidden-character check can reject them
- Add `.trim()` to `parseIssueArg()` and `parseSlashSeparatedArg()` to strip trailing whitespace/newlines from AI agent inputs

## Motivation

Two related `ValidationError` issues from the input validation layer:

- **CLI-14M**: `Invalid project slug: contains a space.` — User passed a display name like `"My Project"` instead of slug `"my-project"`
- **CLI-16M**: `Invalid issue identifier: contains a newline.` — AI agent (Claude) passed an issue ID with a trailing `\n`

The validation itself is correct (agent hallucination hardening), but the parsing layer should normalize recoverable inputs before validation — just like it already normalizes underscores → dashes.

## Key Design Decisions

- **Lowercase only when spaces are present** — `"my_project"` stays `"my-project"` (preserves casing for backward compat), but `"My Project"` → `"my-project"` since spaces imply a display name
- **`validateResourceId()` unchanged** — Still rejects spaces as defense-in-depth for any code path that skips `normalizeSlug`
- **No space normalization in issue IDs** — `parseIssueArg` only trims, since issue identifiers like `"CLI-G5"` should never have internal spaces silently converted
- **Reason-aware warning messages** — `"spaces"` → "use lowercase with dashes, not spaces"; `"both"` → "not spaces or underscores"

### DIFF
--- a/src/lib/arg-parsing.ts
+++ b/src/lib/arg-parsing.ts
@@ -66,11 +66,9 @@ export function normalizeSlug(slug: string): {
     result = result.toLowerCase();
   }
 
-  // Replace underscores and spaces with dashes
-  result = result.replace(/[_ ]/g, "-");
-
-  // Collapse consecutive dashes (from "My  Project" or "a__b")
-  result = result.replace(/-{2,}/g, "-");
+  // Replace runs of underscores/spaces with a single dash.
+  // Using [_ ]+ collapses "My  Project" and "a__b" in one pass.
+  result = result.replace(/[_ ]+/g, "-");
 
   // Trim leading/trailing dashes (from " My Project " → "-my-project-")
   result = result.replace(/^-+|-+$/g, "");
--- a/src/lib/arg-parsing.ts
+++ b/src/lib/arg-parsing.ts
@@ -18,34 +18,96 @@ import { isAllDigits } from "./utils.js";
 // Slug normalization
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/lib/arg-parsing.ts
```diff
@@ -18,34 +18,96 @@ import { isAllDigits } from "./utils.js";
...
+    result = result.toLowerCase();
   }
-  return { slug, normalized: false };
+
+  // Replace underscores and spaces with dashes
+  result = result.replace(/[_ ]/g, "-");
+
+  // Collapse consecutive dashes (from "My  Project" or "a__b")
+  result = result.replace(/-{2,}/g, "-");
+
```
BYK: I think we can combine these two regexes into one and save a `.replace()` call?

BYK: Good call — combined into a single `[_ ]+` regex that replaces runs of underscores/spaces with a single dash in one pass. Done in b0af8ba8.

PR #760, opened 2026-04-15
refactor(time-range): parse --period at flag level via parsePeriod

BYK:
## Summary

- Moves `parsePeriod` into the Stricli flag `parse` function on all `--period` definitions
- `flags.period` is now `TimeRange` instead of `string` — commands get a pre-parsed, validated value automatically
- Invalid periods (bad format, zero duration, etc.) are rejected at parse time, before `func()` runs
- Same pattern as `sanitizeQuery` for `--query` (#758)

## Changes

**Foundation** (`src/lib/time-range.ts`):
- `formatTimeRangeFlag(range)` — converts `TimeRange` back to CLI-friendly string for hints
- `TIME_RANGE_14D`, `TIME_RANGE_24H`, `TIME_RANGE_30D` — pre-parsed constants for runtime defaults

**Shared flag** (`src/lib/list-command.ts`):
- `LIST_PERIOD_FLAG.parse`: `String` → `parsePeriod`

**8 commands updated** (removed manual `parsePeriod()` calls):

| Command | Change |
|---|---|
| `trace list`, `span list` | Use `LIST_PERIOD_FLAG` (already updated), remove `parsePeriod` import |
| `event list`, `issue events` | Same (via `shared-events.ts`) |
| `issue list` | Inline flag: `parse: parsePeriod`, remove redundant `?? DEFAULT_PERIOD` |
| `trace logs` | Inline flag: `parse: parsePeriod` |
| `log list` | `optional: true`, runtime default via `TIME_RANGE_14D`/`TIME_RANGE_30D` |
| `dashboard view` | `optional: true`, runtime default via `parsePeriod(dashboard.period)` + `TIME_RANGE_24H` |

**Excluded**: `release list` — passes `flags.period` as raw `statsPeriod` string to the API.


src/commands/issue/list.ts
```diff
@@ -847,8 +848,8 @@ function appendIssueFlags(base: string, flags: ListFlags): string {
   if (flags.query) {
     parts.push(`-q "${flags.query}"`);
   }
-  if (flags.period && flags.period !== "90d") {
-    parts.push(`-t ${flags.period}`);
+  if (formatTimeRangeFlag(flags.period) !== "90d") {
```
BYK: Looks like:

1. We should abstract out that `"90d"` to a constant
2. Need to abstract more things out as I keep seing the two patterns:
  - `if (formatTimeRangeFlag(flags.period) !== DEFAULT_PERIOD) { ... }`
  - And `formatTimeRangeFlag(flags.period)` in general across multiple files.

BYK: Fixed in 60983673.

1. Replaced the `"90d"` literal with `DEFAULT_PERIOD` constant.
2. Extracted `appendPeriodHint(parts, period, defaultPeriod, flag?)` helper to `time-range.ts` — replaces the repeated `formatTimeRangeFlag` + comparison + push pattern in all 4 hint builders (trace list, span list, shared-events, issue list).

PR #771, opened 2026-04-18
fix(arg-parsing): accept underscores in Sentry slugs (#770)

### LINKED ISSUE #770: Slug normalization breaks projects with underscores in their slug
### Environment

- CLI version: - `0.28.0`
- Installed via: `curl https://cli.sentry.dev/install -fsS | bash`

### Steps to Reproduce

1. Create a Sentry project with underscores in the slug (e.g. test_app_for_cli)
2. Confirm it exists: `$ sentry project list` — shows the project with underscores
3. Run: `$ sentry issue list test_app_for_cli`

### Expected Result

The CLI should accept the project slug as-is and return issues for `test_app_for_cli`. Similar to `$sentry-cli`

Image

### Actual Result

The CLI normalizes the slug, replacing underscores with dashes, and fails:

```
WARN  Normalized slug to 'test-app-for-cli' (Sentry slugs use dashes, never underscores)
Error: Project 'test-app-for-cli' not found.
```

Sentry project slugs can contain underscores — the normalization in `normalizeSlug()` at [/src/lib/arg-parsing.ts#L71](https://github.com/getsentry/cli/blob/main/src/lib/arg-parsing.ts#L71) assumes they cannot.

 The regex `replace(/[_ ]+/g, "-")` replaces underscores with dashes,  so we're not allowing a reference any project whose slug contains underscores.

Introduced in #757 

More information + suggestions in this thread **[No. 179946](https://sentry.zendesk.com/agent/tickets/179946)**

BYK:
## Summary

Fixes #770. The CLI was converting underscores to dashes in `normalizeSlug()` on the assumption that "Sentry slugs never contain underscores". That assumption is wrong — the Sentry platform accepts underscores in project slugs at creation time, and rewriting them at the CLI layer made it impossible for affected customers to look up their own projects.

## Verification

Live API test against a real org before the fix:

```
$ sentry api --method POST /teams/byk-test/byk-test/projects/ \
    --field slug=test_underscore_slug \
    --field name=test_underscore_project \
    --field platform=javascript
# → 200 OK, slug preserved exactly as "test_underscore_slug"

$ sentry api /projects/byk-test/test_underscore_slug/
# → 200 OK, round-trips cleanly
```

Before (the bug):

```
$ sentry issue list byk-test/test_underscore_slug
[arg-parsing]  WARN  Normalized slug to 'byk-test/test-underscore-slug' (Sentry slugs use dashes, never underscores)
Error: Project 'test-underscore-slug' not found in organization 'byk-test'.
```

After this PR (running the dev build against the same project):

```
$ bun run dev issue list byk-test/test_underscore_slug
No issues found.
```

Space normalization still works as before:

```
$ bun run dev issue list "BYK Test/My Project"
[arg-parsing]  WARN  Normalized slug to 'byk-test/my-project' (Sentry slugs use lowercase with dashes, not spaces)
```

## History (for reviewers)

- The underscore rule was introduced in #363 as typo-correction for one specific customer (`selfbase_admin_backend`), never verified against the platform.
- #757 folded underscore and space handling into a combined `[_ ]+` regex. The reporter on #770 correctly spotted that regex — but the root cause predates that PR.
- The legacy `sentry-cli` (Rust) accepts underscored slugs, so the new CLI diverging on this is a regression from the user's perspective.

## Changes

- `normalizeSlug()`: only normalize when the input contains a space. Underscored input is returned unchanged (`normalized: false`).
- Remove the `NormalizeReason` discriminator. With underscores gone, only the "spaces" path remains, so `reason` collapses to `normalized: true` (and the warning message becomes a single string).
- `ParsedOrgProject`: drop the `reason?` field from all three variants.
- `trace-target.ts`: no source change required — already only read `.slug` and `.normalized`.
- Tests: replace underscore-normalization tests with underscore-preservation tests. Mixed inputs like `my_org/My Project` now correctly yield `{ org: "my_org", project: "my-project" }`. Delete the three command-level underscore-warning tests (the code path no longer exists; unit-level space coverage is sufficient).

## Verification

- `bun run typecheck` — clean
- `bun run lint` — clean (pre-existing warning in `markdown.ts` unrelated)
- `bun run test:unit` — 5036 / 5036 pass
- `bun run test:isolated` — 138 / 138 pass
- `bun run check:errors`, `bun run check:deps` — clean
- Manual end-to-end repro against real Sentry org (above).

Net diff: −110 lines.

### DIFF
--- a/src/lib/arg-parsing.ts
+++ b/src/lib/arg-parsing.ts
@@ -18,91 +18,55 @@ import { isAllDigits } from "./utils.js";
 // Slug normalization
 // ---------------------------------------------------------------------------
 
-/** Describes what was normalized in a slug — used for targeted warning messages. */
-export type NormalizeReason = "underscores" | "spaces" | "both";
-
 /**
- * Normalize a Sentry slug by replacing underscores and spaces with dashes.
+ * Normalize a Sentry slug by converting display-name-style input to slug form.
  *
- * Sentry enforces that slugs use lowercase letters, numbers, and dashes.
- * Users frequently type underscores by mistake or paste display names
- * (e.g., `"My Project"` instead of `"my-project"`).
+ * Sentry slugs are always lowercase and use dashes as separators. Users
+ * sometimes paste display names like `"My Project"` instead of
+ * `"my-project"`; without normalization, `validateResourceId` would reject
+ * those outright for containing spaces. Normalization gives us a chance to
+ * recover the intent before validation runs.
  *
- * Normalization rules:
- * 1. Underscores → dashes (existing)
- * 2. Spaces → dashes, with lowercase (spaces imply a display name)
- * 3. Consecutive dashes collapsed (`"My  Project"` → `"my-project"`)
- * 4. Leading/trailing dashes trimmed
+ * Normalization rules (applied only when spaces are present):
+ * 1. Lowercase the entire string
+ * 2. Collapse runs of spaces into a single dash
+ * 3. Trim leading/trailing dashes (from leading/trailing whitespace)
  *
- * Lowercasing is only applied when spaces are present. Pure underscore
- * normalization preserves casing for backward compatibility.
+ * Underscores are **not** normalized — Sentry accepts underscores in project
+ * slugs at creation time, so rewriting them here would silently break
+ * lookups for any customer who has one (see #770). Since underscores are
+ * not in `validateResourceId`'s forbidden set either, they flow through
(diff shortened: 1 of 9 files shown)


BYK:
Good catch — fixed in 51f59d67. Updated both the `warnIfNormalized()` message and the `ParsedTraceTarget` JSDoc in `trace-target.ts` to reference spaces instead of underscores.

Issue #785, opened 2026-04-20
CLI UX: cache invalidation, env var discoverability, and error message improvements


BYK:
Bundling a set of UX issues surfaced while testing the Sentry CLI against a freshly provisioned project (via Stripe Projects integration). Individually low-severity but collectively they make the CLI harder to use when the common \"works out of the box\" path doesn't apply.

### Caching

1. **\`-f/--fresh\` flag is undocumented and non-discoverable.** The flag exists on \`org list\`, \`project view\`, \`auth whoami\`, etc. but isn't mentioned in the top-level \`sentry --help\` or in the configuration docs. Users hit stale data with no indication that a refresh flag exists. Consider:
   - Adding a \"Common flags\" section to \`sentry --help\` that documents \`-f\`.
   - Emitting a subtle hint in list/view output when data came from cache (e.g. dim \"cached · 5m old · use -f to refresh\").

2. **Caches should auto-invalidate on known mutations.** When the CLI itself performs an operation that changes state (\`project delete\`, \`project create\`, \`issue resolve\`, \`issue unresolve\`, \`auth login\`, \`auth logout\`), it should invalidate the corresponding cache entries automatically. Today users have to remember \`-f\` after mutations, which defeats the purpose of caching.

3. **\`sentry auth whoami\` returns a stale identity after switching accounts.** Same family as #2 — \`auth login\` / \`auth logout\` should invalidate the \`whoami\` cache.

### Auth and environment variables

4. **\`SENTRY_FORCE_ENV_TOKEN\` is undiscoverable.** Nothing in \`sentry --help\`, \`sentry auth --help\`, or https://cli.sentry.dev/configuration/ documents this. It's critical for the common case of \"I have a valid token for a resource my logged-in user can't see.\" Suggested fix: when \`SENTRY_AUTH_TOKEN\` is set *and* there's a stored login, emit a hint:

   \`\`\`
   Detected SENTRY_AUTH_TOKEN env var but also a stored login for <user>.
   Using stored login. To force the env var, set SENTRY_FORCE_ENV_TOKEN=1.
   \`\`\`

[...]

PR #788, opened 2026-04-20
feat(cache): scope response cache per active identity (#785)

BYK:
## Summary

Cached GET responses now live in per-identity namespaces so switching accounts never serves another user's data. `buildCacheKey(method, url)` internally mixes in a fingerprint of the active bearer identity, producing distinct cache files per account for the same URL.

**Identity fingerprint** — a memoized 16-char MD5 hex of `kind|secret`, derived (in order) from:

- **OAuth** — the stored `refresh_token`. Hourly access-token rotation keeps the same refresh_token, so the cache stays hot across refreshes. If the OAuth server DOES rotate the refresh_token on refresh (per spec, allowed), the fingerprint changes and the cache naturally re-populates under the new identity.
- **Env token** — the raw `SENTRY_AUTH_TOKEN` / `SENTRY_TOKEN` value. Env tokens don't rotate, so hashing the token is equivalent to hashing a stable identity.
- **No token** — sentinel `"<anon>"` (angle brackets can't appear in a real hex fingerprint, so no collision with a computed hash is possible).

Memoization is reset on `setAuthToken` and `clearAuth`. OAuth refresh goes through `setAuthToken` so the fingerprint rotates correctly. MD5 is intentional: this is a cache namespace, not authentication — collisions are benign (two colliding identities would share cache space, like the anonymous case). CodeQL flagged this as a weak-crypto/insufficient-hash finding; dismissed as "won't fix" with context.

This fixes the stale-data bugs reported in #785:

- **#3** `auth whoami` no longer returns the previous user after `auth login` / env-token swap.
- **#2** Post-mutation cache staleness is now bounded to a single identity rather than leaking across accounts.

**Mutation hooks call `invalidateCachedResponsesMatching` for same-identity freshness:**
`project create`, `project delete`, `issue resolve`, `issue unresolve`, `issue merge` all flush the relevant detail and list endpoints immediately. The sweep gates on an `identity` field persisted in every cache entry so one identity's mutation can't evict another identity's cache on a shared cache directory.

**Cache-key / query-param correctness:** `getIssue` / `getIssueInOrg` pass `?collapse=stats&collapse=lifetime&...` so cached URLs carry query strings. Using exact-match `invalidateCachedResponse(<base url without params>)` would silently miss them. Mutation paths now use prefix-sweep (`invalidateCachedResponsesMatching`) on the trailing-slash base URL — which catches every cached variant regardless of query params.

**URL-construction helper:** Introduced `buildApiUrl(regionUrl, ...segments)` in `api/infrastructure.ts` to replace repeated `${base}/api/0/path/${encodeURIComponent(slug)}/` template-string building. Per-segment encoding means a slug containing a literal `/` (rare but possible) now produces a correctly-encoded URL instead of a silently-broken one.

Cache-invalidation helpers in `api/issues.ts` and `api/projects.ts` are contractually no-throw — a post-mutation DB/FS failure inside the helper can't turn a successful mutation into a user-visible error.

**Circular import note:** `db/auth.ts` previously imported `clearResponseCache` statically; now that `response-cache.ts` imports `getIdentityFingerprint` from `db/auth.ts`, that's resolved via a dynamic `await import()` inside the (already-async) `clearAuth()`.

## Test plan

- `bun test test/lib/db/auth.test.ts` — `getIdentityFingerprint` coverage: stability across OAuth access-token rotation, distinct fingerprints per env token / per OAuth session, `SENTRY_FORCE_ENV_TOKEN` switching the source, `kind`-prefix isolation (env/oauth with same secret produce distinct fingerprints), expired access-only token fall-through to env.
- `bun test test/lib/response-cache.test.ts` — identity isolation (two different auth states never share a cache key for the same URL; `invalidateCachedResponsesMatching` leaves other identities' cached entries alone) plus a regression test for the query-param-variant case.
- `bun test test/lib/api/infrastructure.test.ts` — `buildApiUrl` composition, trailing-slash normalization, reserved-char encoding (slash, space), and the zero-segments edge case.
- Full unit suite: 5527 passing.

Part of #785 (addresses items #2, #3, and half of #7 — the other half is fixed by a Sentry backend change).


src/lib/response-cache.ts
```diff
@@ -602,13 +616,70 @@ async function writeResponseToCache(
...
+      if (entry.url?.startsWith(prefix)) {
+        await unlink(filePath).catch(() => {
+          // Best-effort: another process may have deleted it
+        });
+      }
+    } catch {
+      // Unparseable or missing — leave to cleanup
+    }
+  });
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 6cbb21dc. Added an `identity` field to each `CacheEntry` and the prefix sweep now double-checks `entry.identity === currentIdentity` before deleting. Entries written before this field existed are treated as foreign and skipped — they still age out via TTL cleanup.

New test `does not delete entries belonging to a different identity` in `test/lib/response-cache.test.ts` covers this: identity A writes a cached entry, identity B sweeps the same prefix, A's entry survives. Also: the cache key already scoped via HMAC-SHA-256 fingerprint so filename collisions across identities were impossible — this was specifically a prefix-sweep-only hole.

src/lib/api/issues.ts
```diff
@@ -612,6 +633,17 @@ export async function mergeIssues(
...
+    // downstream reads (`issue view`, `issue list`) see the merge
+    // result immediately. Children disappear into the parent so they
+    // need invalidation too.
+    const affectedIds = [data.merge.parent, ...data.merge.children];
+    await Promise.all(
+      affectedIds.map((id) => invalidateIssueCaches(regionUrl, orgSlug, id))
+    );
+    // Also flush the org-wide issue list endpoint since merges change
+    // which groups appear in paginated results.
+    await invalidateOrgIssueList(regionUrl, orgSlug);
```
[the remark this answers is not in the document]
BYK: Fixed in 6cbb21dc. Split `invalidateIssueCaches` into `invalidateIssueDetailCaches` (detail URLs only) + `invalidateIssueCaches` (detail + list). Merge now calls detail-only in the per-ID loop and `invalidateOrgIssueList` exactly once post-loop — turning N+1 full cache scans into 1.

src/lib/api/issues.ts
```diff
@@ -671,3 +703,53 @@ export async function getSharedIssue(
 
   return (await response.json()) as { groupID: string };
 }
+
+/** Strip a trailing slash from a base URL for clean `/api/0/...` composition. */
+function stripTrailingSlashLocal(url: string): string {
+  return url.endsWith("/") ? url.slice(0, -1) : url;
+}
```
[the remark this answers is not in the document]
BYK: Fixed in 6cbb21dc. Moved `stripTrailingSlash` to `src/lib/api/infrastructure.ts` as a proper export; both `api/issues.ts` and `api/projects.ts` import from there now.

src/lib/api/projects.ts
```diff
@@ -166,6 +171,13 @@ export async function createProject(
     body,
   });
   const data = unwrapResult(result, "Failed to create project");
+
+  // Invalidate any cached project list for this org so the next
+  // `project list` / `project view` call picks up the new project
+  // instead of serving a stale list. Best-effort — failures are
+  // silently swallowed inside the helper.
+  await invalidateOrgProjectCache(orgSlug);
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in 6cbb21dc. Both `createProject` and `deleteProject` now wrap the invalidation block in try/catch with a "Non-fatal — the mutation already succeeded" comment, matching the pattern already used by the legacy path in `updateIssueStatus`. A throw from `resolveOrgRegion` after a successful create/delete was indeed a real hazard.

src/lib/response-cache.ts
```diff
@@ -602,13 +658,86 @@ async function writeResponseToCache(
...
+    // Unknown read error — best-effort: skip
+    return;
+  }
+
+  const jsonFiles = files.filter((f) => f.endsWith(".json"));
+  if (jsonFiles.length === 0) {
+    return;
+  }
+
+  const currentIdentity = getIdentityFingerprint();
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in a7f1a05f. Two layers:

1. `invalidateCachedResponsesMatching` now wraps its whole body in try/catch so `getIdentityFingerprint()` failures (or any future addition) can't propagate.
2. Every invalidation helper in `api/issues.ts` and `api/projects.ts` now has a "Never throws" contract — `invalidateIssueCaches`, `invalidateIssueDetailCaches`, `invalidateOrgIssueList`, `invalidateOrgProjectCache` all wrap their bodies in try/catch. Callers are safe without having to remember. The existing caller-side try/catch in `createProject` / `deleteProject` stays as defense-in-depth (no runtime cost).

src/lib/db/auth.ts
```diff
@@ -246,6 +250,124 @@ export function isAuthenticated(): boolean {
...
+      .get() as
+      | { token: string | null; refresh_token: string | null }
+      | undefined;
+  });
+  if (dbRow?.refresh_token) {
+    return hashIdentity("oauth", dbRow.refresh_token);
+  }
+  if (dbRow?.token) {
+    return hashIdentity("oauth-access", dbRow.token);
+  }
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in a7f1a05f. `getIdentityFingerprint` now mirrors `getAuthConfig`'s expiry semantics: an expired access-only token (with `expires_at` in the past and no refresh_token) is skipped and the lookup falls through to the env token. Added two tests: one covering the expired→env fallback, one confirming that an expired access token with a refresh token keys off the refresh token (unchanged by access-token rotation).

src/lib/api/issues.ts
```diff
@@ -562,13 +568,29 @@ export async function updateIssueStatus(
...
+  const legacyData = await apiRequest<SentryIssue>(
+    `/issues/${encodeURIComponent(issueId)}/`,
+    { method: "PUT", body }
+  );
+  // Without an org slug we don't know the region, so we can only flush
+  // the legacy issue-detail URL. Region-scoped lists still age out via
+  // their TTL. Callers that know the org should pass it to enable the
+  // broader org-scoped invalidation above.
+  try {
+    const { getApiBaseUrl } = await import("../sentry-client.js");
```
BYK: Why the lazy import here?

src/lib/api/issues.ts
```diff
@@ -612,6 +634,20 @@ export async function mergeIssues(
...
       body: { merge: 1 },
     });
+    // Flush cached detail entries for every affected group ID — children
+    // disappear into the parent, and the parent's cached detail predates
+    // the merge. Then sweep the org-wide list once. Using
+    // `invalidateIssueDetailCaches` (detail-only) instead of
+    // `invalidateIssueCaches` here avoids N+1 full cache-directory scans
+    // for large merges: a single post-loop `invalidateOrgIssueList`
+    // covers every list variant in one pass.
+    const affectedIds = [data.merge.parent, ...data.merge.children];
```
BYK: `data.merge.children.unshift()` would probably be more efficient here.

src/lib/db/auth.ts
```diff
@@ -246,6 +250,138 @@ export function isAuthenticated(): boolean {
...
 }
 
+/**
+ * Fingerprint returned when no token is present (logged out, no env var).
+ *
+ * Keeps cache-key generation total without coupling response-cache to
+ * undefined handling: anonymous callers still share a namespace across
+ * invocations, which matches the pre-identity-scoped behavior.
+ */
+export const ANON_IDENTITY = "anon";
```
BYK:
```suggestion
export const ANON_IDENTITY = "<anon>";
```

src/lib/db/auth.ts
```diff
@@ -246,6 +250,138 @@ export function isAuthenticated(): boolean {
...
+ * The result is a SHA-256 hex digest truncated to 16 characters
+ * (64 bits) — enough to make collisions astronomically unlikely for a
+ * single-user CLI while keeping cache filenames short.
+ *
+ * Performance: this function runs on every HTTP request (for cache
+ * lookups), so it must stay synchronous and cheap. SHA-256 over a
+ * short token is a few hundred nanoseconds — well below the HTTP
+ * latency budget.
+ */
+export function getIdentityFingerprint(): string {
```
BYK, line 278: The inputs for this function should never change throughout the app lifecycle so why not cache the result since it is performance-critical?

src/lib/db/auth.ts
```diff
@@ -246,6 +250,138 @@ export function isAuthenticated(): boolean {
...
+ *    token — but the checker has no way to tell the difference.
+ * 2. Adds a tiny obfuscation on cache-dir contents for anyone inspecting
+ *    the filesystem. Still not a secret, still not an authenticator,
+ *    but slightly better than a bare SHA-256.
+ *
+ * The key never changes. Rotating it would churn every user's cache
+ * namespace on upgrade — the identity fingerprints on disk must stay
+ * stable across CLI versions to avoid invalidating the whole cache.
+ */
+const IDENTITY_HMAC_KEY = "sentry-cli/identity-fingerprint/v1";
```
BYK: Let's avoid using `sentry-cli` as it can get confused with the legacy "sentry-cli" CLI project. We can just use `sentry/`

src/lib/db/auth.ts
```diff
@@ -246,6 +250,138 @@ export function isAuthenticated(): boolean {
...
+
+/**
```
BYK: Do we really need HMAC for this? Why not use a very fast and cheap hash like MD5?

src/lib/response-cache.ts
```diff
@@ -388,7 +419,7 @@ export async function getCachedResponse(
 
   let key: string;
   try {
-    key = buildCacheKey(method, url);
+    key = buildCacheKey(getIdentityFingerprint(), method, url);
```
BYK: `getIdentityFingerprint()` result should never change so why add it here and make the function call for `buildCacheKey` more complicated than it needs to be?

src/lib/response-cache.ts
```diff
@@ -602,13 +658,98 @@ async function writeResponseToCache(
...
+ *
+ * @param prefix - Full-URL prefix (e.g.
+ *   `"https://us.sentry.io/api/0/organizations/acme/projects/"`).
+ *   Exact-matching `invalidateCachedResponse` is preferred when the
+ *   caller can construct the precise URL.
+ */
+export async function invalidateCachedResponsesMatching(
+  prefix: string
+): Promise<void> {
+  // Whole-body try/catch: this helper is called from post-mutation
```
BYK: Your comments in this entire PR are a bit excessive. We can afford some brevity.

src/lib/response-cache.ts
```diff
@@ -602,13 +658,98 @@ async function writeResponseToCache(
...
+    const cacheDir = getCacheDir();
+    let files: string[];
+    try {
+      files = await readdir(cacheDir);
+    } catch (error) {
+      if (isNotFound(error)) {
+        return;
+      }
+      // Unknown read error — best-effort: skip
+      return;
```
BYK: Why do we even need the `isNotFound` check if we're gonna ignore the error anyway?

src/lib/api/issues.ts
BYK: Defensive — no actual cycle. Made it a regular top-level import in e59f7503.

src/lib/api/issues.ts
BYK: Used `[data.merge.parent, ...data.merge.children]` in e59f7503 — `unshift` would mutate the API response object that we return to the caller a line later. Spread is equivalent cost and non-mutating.

src/lib/db/auth.ts
BYK: Changed to `"<anon>"` in e59f7503.

src/lib/db/auth.ts
BYK: Good call — memoized in e59f7503. Reset on `setAuthToken`/`clearAuth` (so OAuth refresh with a rotated refresh_token still picks up correctly) plus `resetIdentityFingerprintCache()` for tests that mutate env state mid-case.

src/lib/db/auth.ts
BYK: Moot now — dropped HMAC entirely for plain MD5 per your other comment. The fixed key is gone.

src/lib/db/auth.ts
BYK: Switched to MD5 in e59f7503. You're right — this is a cache namespace, not authentication; collisions are benign (two colliding identities would share a cache slot, same as the anon case). Since the fingerprint is now memoized (computed once), even the speed difference is moot, but MD5 removes the "why HMAC" question entirely.

src/lib/response-cache.ts
BYK: Dropped — `buildCacheKey(method, url)` is 2-arg again in e59f7503. Identity is read internally via the memoized `getIdentityFingerprint()`. Cleaner call sites; matches the pre-PR ergonomics.

src/lib/response-cache.ts
BYK: Agreed. Did a broad trim pass across response-cache.ts, db/auth.ts, api/issues.ts, api/projects.ts plus the tests. Net -265 lines in e59f7503.

src/lib/response-cache.ts
BYK: Right, redundant — removed in e59f7503. The outer `try/catch` swallows the throw either way, so the `isNotFound` branch was dead weight.

src/lib/db/auth.ts
```diff
@@ -246,6 +250,85 @@
...
+}
+
+/**
+ * 16-char MD5 hex of `kind|secret`. Not used for auth — just a cheap
+ * cache namespace. Collisions are benign (identities would share a
+ * cache slot, same as the anonymous case).
+ */
+function hashIdentity(kind: string, secret: string): string {
+  return createHash("md5")
+    .update(`${kind}|${secret}`)
```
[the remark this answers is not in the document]
BYK: Switched `hashIdentity` back to SHA-256 in ec0279c1. Since the fingerprint is memoized (one call per process), the speed argument for MD5 is moot — we run the hash once, ever. Killing both CodeQL findings (this one and the derived one on `response-cache.ts:123`) is worth more than 400ns.

src/lib/response-cache.ts
```diff
@@ -108,20 +109,19 @@
...
- * @param method - HTTP method (e.g., "GET")
- * @param url - Full URL string
- * @returns Hex-encoded SHA-256 hash suitable for use as a filename
  * @internal Exported for testing
  */
 export function buildCacheKey(method: string, url: string): string {
   const normalized = normalizeUrl(method, url);
-  return createHash("sha256").update(normalized).digest("hex");
+  return createHash("sha256")
+    .update(`${getIdentityFingerprint()}|${normalized}`)
```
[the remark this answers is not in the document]
BYK: Derived from the `hashIdentity` MD5 call. Fixed at the source in ec0279c1 by switching to SHA-256; this finding goes away with it.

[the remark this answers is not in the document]
BYK: Dismissed as 'won't fix' — this is a cache namespace, not a password hash. MD5 is intentional (speed-over-correctness tradeoff that collapses given memoization). Taint chain from `hashIdentity` is handled at the source.

src/lib/db/auth.ts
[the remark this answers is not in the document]
BYK: Dismissed as 'won't fix' — false positive in this context. Input secrets are auth tokens already stored plaintext in the user's SQLite DB; hashing adds no security property here, it just gives us an opaque filename-safe identifier for cache namespacing.

src/lib/api/issues.ts
```diff
@@ -671,3 +696,56 @@ export async function getSharedIssue(
...
+ */
+async function invalidateOrgIssueList(
+  regionUrl: string,
+  orgSlug: string
+): Promise<void> {
+  const base = stripTrailingSlash(regionUrl);
+  await invalidateCachedResponsesMatching(
+    `${base}/api/0/organizations/${encodeURIComponent(orgSlug)}/issues/`
+  );
+}
```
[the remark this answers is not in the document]
BYK: Good catch — restored in ec0279c1. I trimmed these in the brevity pass, figuring "leaves are no-throw, so orchestration inherits" — but that's a weaker contract than enforcing the 'Never throws' JSDoc at the boundary itself. `issues.ts` helpers now match `projects.ts`.

PR #789, opened 2026-04-20
fix(errors): surface real API errors and granular scopes in 403s (#785)

BYK:
## Summary

Two error-message fixes:

**Item #8 — `project view` swallowed real API errors.**
`sentry project view <org>/<project>` previously caught any non-auth `ApiError` (403 forbidden, 404 not found, etc.) inside `fetchProjectDetails`, returned `null`, and then threw `ContextError("Could not auto-detect organization and project")` — despite the user having explicitly provided both.

Now fetch errors rethrow verbatim for `explicit` and `project-search` modes so the user sees the real cause. `auto-detect` mode (which may surface multiple DSN-discovered targets) keeps the pre-fix per-target tolerance, since one inaccessible DSN shouldn't block the rest.

**Item #9 — 403 hints named the wrong missing scopes.**
The 403 handler in `issue list` and `listOrganizations` said the token "may lack the required scopes (`org:read`, `project:read`)" even when it had those and was actually missing `event:read` or another scope. Sentry's API response often names the real culprit; we now parse it via a new `extractRequiredScopes(detail)` helper in `src/lib/api-scope.ts` and surface the specific scope(s) in the suggestion. Falls back to the pre-fix hardcoded list when the API doesn't tell us.

The scope extractor handles three common response shapes:
- Plain detail strings with embedded scope identifiers.
- Top-level `required` / `requiredScopes` / `scopes` arrays.
- Arrays of `{scope: "..."}` objects.

Only recognizes known Sentry scope namespaces (`org`/`project`/`team`/`member`/`event`/`release`/`alerts`) so random `foo:bar` substrings in unrelated error text don't get surfaced as "scopes".

## Test plan

- New `test/lib/api-scope.test.ts` covers all response shapes, dedup, case-insensitivity, and rejection of unknown namespaces.
- `test/commands/project/view.func.test.ts` updated: explicit/project-search targets now assert the real error bubbles up; auto-detect target keeps the pre-fix error-swallowing.
- Full unit suite: 5275 passing.

Part of #785 (addresses items #8 and #9).

### DIFF
--- a/src/commands/project/view.ts
+++ b/src/commands/project/view.ts
@@ -13,7 +13,7 @@ import {
 } from "../../lib/arg-parsing.js";
 import { openInBrowser } from "../../lib/browser.js";
 import { buildCommand } from "../../lib/command.js";
-import { ContextError, withAuthGuard } from "../../lib/errors.js";
+import { AuthError, ContextError, withAuthGuard } from "../../lib/errors.js";
 import { divider, formatProjectDetails } from "../../lib/formatters/index.js";
 import { CommandOutput } from "../../lib/formatters/output.js";
 import {
@@ -92,15 +92,21 @@ type ProjectWithDsn = {
 };
 
 /**
- * Fetch project details and keys for a single target.
- * Returns null on non-auth errors (e.g., no access to project).
- * Rethrows auth errors so they propagate to the user.
+ * Perform the parallel project + DSN fetch for a single target.
+ *
+ * Shared by both `fetchProjectDetails` (which swallows non-auth
+ * failures) and `fetchProjectDetailsOrThrow` (which rethrows them).
+ * Extracted so the two callers stay in lockstep — if a future change
+ * adds another parallel call or swaps the DSN lookup, both paths
+ * benefit automatically.
+ *
+ * `AuthError` (from `withAuthGuard`) always propagates so the
+ * auto-login middleware in `cli.ts` can fire.
  */
-async function fetchProjectDetails(
-  target: ResolvedTarget
-): Promise<ProjectWithDsn | null> {
-  const result = await withAuthGuard(async () => {
-    // Fetch project (skip if already fetched during resolution) and DSN in parallel
+function fetchProjectAndDsn(target: ResolvedTarget): Promise<ProjectWithDsn> {
+  return withAuthGuard(async () => {
+    // Fetch project (skip if already fetched during resolution) and
+    // its DSN in parallel.
     const [project, dsn] = await Promise.all([
       target.projectData
         ? Promise.resolve(target.projectData)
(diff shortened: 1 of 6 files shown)


src/commands/project/view.ts
```diff
@@ -112,6 +112,40 @@ async function fetchProjectDetails(
...
+      tryGetPrimaryDsn(target.org, target.project),
+    ]);
+    return { project, dsn };
+  });
+  if (guarded.ok) {
+    return guarded.value;
+  }
+  // Non-auth failure: rethrow the captured error unchanged.
+  throw guarded.error;
+}
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in d3f88f51 by extracting a shared `fetchProjectAndDsn` helper used by both `fetchProjectDetails` (swallows non-auth errors) and `fetchProjectDetailsOrThrow` (rethrows them). The two callers now only differ in error policy; the fetch body is a single source.

Same commit also addresses a separate review question — the scope regex needed to be verified against the real `getsentry/sentry` `SENTRY_SCOPES` list in `src/sentry/conf/server.py`. Turned out I had a few drifts:
- `release:read/write/admin` was accepted but isn't a real namespace; Sentry uses `project:releases` / `project:distribution`.
- `org:integrations`, `org:ci`, `member:invite`, `project:releases`, `project:distribution` were missing.
- `alerts:admin` was accepted but Sentry has no admin tier on alerts.

Now sourced from a single const array so adding/removing scopes is a one-line change.

src/lib/api-scope.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Build a word-bounded alternation regex from {@link SENTRY_SCOPES}.
+ *
+ * Using an explicit alternation (rather than a `<ns>:<action>` product)
+ * avoids matching nonexistent combinations like `release:write` or
+ * `alerts:admin`, which `SENTRY_SCOPES` doesn't list.
+ */
+const KNOWN_SCOPE_RE = new RegExp(
+  `\\b(?:${SENTRY_SCOPES.map((s) => s.replace(":", ":")).join("|")})\\b`,
```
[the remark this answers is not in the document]
BYK: Silly leftover from when I was mid-thought about regex escaping — `:` isn't a metacharacter so the `.map(s => s.replace(":", ":"))` was a no-op. Dropped in 9931385e, tests still green.

src/commands/project/view.ts
```diff
@@ -92,24 +92,74 @@ type ProjectWithDsn = {
...
       tryGetPrimaryDsn(target.org, target.project),
     ]);
     return { project, dsn };
+  }).then((result) => {
+    if (result.ok) {
+      return result.value;
+    }
+    // Non-auth failure: rethrow the captured error so the caller can
+    // decide whether to swallow it or surface it.
+    throw result.error;
```
BYK: Is this safe in a `.then()` handler?

src/lib/api-scope.ts
```diff
@@ -0,0 +1,201 @@
...
+function dedupe(items: string[]): string[] {
+  const seen = new Set<string>();
+  const out: string[] = [];
+  for (const item of items) {
+    if (!seen.has(item)) {
+      seen.add(item);
+      out.push(item);
+    }
+  }
+  return out;
```
BYK: Why not just `return Array.from(new Set(items))`?

src/commands/project/view.ts
BYK: Rewrote as a plain `async` function in 8a970af8 — native `try/catch` is clearer. The `.then` version *was* safe (thrown values become rejected promises that the caller's `await` re-throws) but it read awkwardly and the "AuthError bypasses the catch" comment I had on `fetchProjectDetails` was actually wrong — both paths land in the same catch.

src/lib/api-scope.ts
BYK: Right — JS preserves `Set` insertion order. Dropped the helper and inlined `[...new Set(items)]` at both call sites in 8a970af8. Also took the opportunity to trim a lot of over-explaining JSDoc across both files (-105 net lines).

PR #790, opened 2026-04-20
feat(auth): hint when env token is shadowed by stored OAuth (#785)

BYK:
## Summary

The most painful UX item from #785: a user sets `SENTRY_AUTH_TOKEN` (e.g. from a Stripe Projects or Vercel integration) but also has a stored OAuth login from `sentry auth login`. The CLI silently prefers the stored login — the user's 30 minutes of debugging invariably ends at discovering `SENTRY_FORCE_ENV_TOKEN=1`.

Surface the collision on stderr the first time an authenticated command hits the API:

```
[info] [auth] Detected SENTRY_AUTH_TOKEN env var but using stored login for alice.
              Set SENTRY_FORCE_ENV_TOKEN=1 to prefer the env var.
```

**Gating:**

- Fires only when an env token is set AND a stored OAuth login exists AND `SENTRY_FORCE_ENV_TOKEN` is not set.
- Fires at most once per process (module-local latch).
- Fires inside `authenticatedFetch`, so local-only commands (`sentry help`, `sentry cli upgrade`) stay quiet.

User label resolution prefers `username` → `email` → `name` → `"stored OAuth user"` fallback, matching what `sentry auth whoami` shows when the cache is cold.

## Test plan

- New `test/lib/auth-hint.test.ts` covers: fires on collision, silent without env token, silent without stored OAuth, silent under `SENTRY_FORCE_ENV_TOKEN`, once-per-process latch, user label fallbacks, `SENTRY_TOKEN` legacy var support.
- Full unit suite: 5269 passing.

Part of #785 (addresses item #4).


src/lib/auth-hint.ts
```diff
@@ -0,0 +1,111 @@
...
+    hasStored = hasStoredAuthCredentials();
+  } catch {
+    // DB access failed — silently skip; this is a best-effort hint.
+    return;
+  }
+  if (!hasStored) {
+    return;
+  }
+
+  hintEmitted = true;
```
BYK: Should probably report the error? Also `hasStored` would still be `false` if an exception is thrown so the explicit `return` in the `catch` block is unnecessary.

BYK: Good catches — both fixed in 83076f5d:

- `Sentry.captureException(error)` in both catch blocks (`resolveStoredUserLabel` had the same pattern — fixed there too).
- Removed the redundant `return` in the `hasStoredAuthCredentials` catch; `hasStored` stays `false` so the existing `if (!hasStored) return` guard handles it. Added a comment explaining the fall-through.

PR #797, opened 2026-04-21
refactor(init): migrate grep/glob tools to src/lib/scan/

BYK:
## Summary

Follow-up to #791. Replaces the init-wizard's `rg → git grep → fs walk` fallback chain in `src/lib/init/tools/grep.ts` + `glob.ts` with thin adapters over the pure-TS `collectGrep` / `collectGlob` helpers from `src/lib/scan/`.

The Mastra wire contract is preserved byte-identical on all existing fields. Two optional fields (`caseInsensitive`, `multiline`) are added to `GrepSearch` for future-proofing — no current server invocation sets them.

### Adapters (net −392 LOC of production code)

| File | Before | After |
|---|---:|---:|
| `src/lib/init/tools/grep.ts` | 337 LOC | 114 LOC |
| `src/lib/init/tools/glob.ts` | 147 LOC | 73 LOC |
| `src/lib/init/tools/search-utils.ts` | 146 LOC | **deleted** |
| `src/lib/init/types.ts` | — | +17 LOC (two optional fields) |

The adapters now just:
1. Sandbox the user-supplied `search.path` via `safePath`.
2. Forward each search/pattern to `collectGrep` / `collectGlob` with the wire-level constants (`maxResults`, `maxLineLength`) plumbed through.
3. Strip `absolutePath` from each `GrepMatch` — the Mastra wire has never included it.
4. Catch `ValidationError` from bad regex so a single bad pattern surfaces as an empty per-search row rather than aborting the whole payload.

### New optional `GrepSearch` fields

```ts
export type GrepSearch = {
  pattern: string;
  path?: string;
  include?: string;
  caseInsensitive?: boolean;  // NEW
  multiline?: boolean;         // NEW
};
```

No current Mastra server invocation sets these. Adding them now means the server can start sending them without a CLI update. The underlying scan engine natively supports both.

### Tests

- **Deleted 3 obsolete tests** that shadowed `rg` in `PATH` to force the fallback chain. With the pure-TS adapter there's no fallback chain to exercise; the tests had become tautological.
- **Deleted scaffolding**: `writeExecutable`, `setPath`, `helperBinDir`, `savedPath` — only used by the deleted tests.
- **Kept 3 pre-existing tests** unchanged: include filters, multi-pattern glob, sandbox rejection.
- **Added 3 new adapter-specific tests**:
  - `grep result matches MUST NOT include absolutePath` — pins the strip behavior so `absolutePath` never leaks to the Mastra agent.
  - `grep bad regex yields empty matches without crashing the payload` — documents the `ValidationError` catch contract.
  - `grep caseInsensitive flag enables case-insensitive matching` — end-to-end coverage for the new wire field.

## Behavior changes (intentional, only affects users without `rg`/`git`)

Before this PR, the init-wizard fs-walk fallback was naive: no `.gitignore` handling, narrow skip list, no binary detection. Users with `rg` or `git` installed never took this path. After this PR, **every user gets rg-like behavior** via the pure-TS scanner:

- **Nested `.gitignore`** honored (cumulative semantics, matching git + rg).
- **Wider skip-dir list** — scan excludes `.next`, `target`, `vendor`, `coverage`, `.cache`, `.turbo` in addition to the old skip set.
- **Binary files filtered** — scan runs an 8 KB NUL-byte sniff before grep-ing.

Users with `rg` installed see no behavior change relative to the rg-path of the old adapter.

## Benchmarks

Measured after rebasing onto `main` (includes #791, #804, #805, #807 — the full grep/worker-pool stack).

**Fixture:** 10k files (~80 MB), 3 monorepo packages, mix of text + binary.
**Config:** 5 runs after 2 warmup, pre-warmed page cache.
**Machine:** linux/x64, 4 CPUs, Bun 1.3.11. ripgrep 14.1.0.
**Harness:** `/tmp/bench-init-tools.ts` — invokes the adapter via the real `grep(GrepPayload)` entry point Mastra uses.

### Adapter vs rg — single search

| Pattern | rg | NEW uncapped | NEW cap=100 | uncapped vs rg | cap vs rg |
|---|---:|---:|---:|---:|---:|
| `import.*from` (215k matches) | 284 ms | 769 ms | **28.4 ms** | 2.70× slower | **10× FASTER** |
| `function\s+\w+` (216k) | 306 ms | 782 ms | **28.2 ms** | 2.56× slower | **11× FASTER** |
| `SENTRY_DSN` (677) | 81 ms | 496 ms | 79.3 ms | 6.10× slower | parity |
| no matches | 74 ms | 486 ms | 492 ms | 6.6× slower | 6.7× slower |

The init-wizard workload is always capped at 100. On dense-match patterns (the common case — Mastra greps for common code patterns like imports and function signatures) the worker pool's literal-prefilter + concurrent fan-out + early-exit hit 100 matches in ~28 ms, **10× faster than rg**. On rare/no-match patterns we're walker-bound at ~500 ms — still fine for init wizard.

### Adapter multi-search (realistic Mastra workload)

Payload: 3 patterns (`import.*from`, `SENTRY_DSN`, `function\s+\w+`) in parallel via `Promise.all`, `maxResultsPerSearch: 100`.

| Metric | Value |
|---|---:|
| p50 end-to-end | **109 ms** |

All three searches share the same worker pool (dispatched concurrently).

### Scan-harness numbers (context)

From `bun run bench --size large`, showing that the new adapter is routed through the worker-backed grep pipeline:

| Op | p50 |
|---|---:|
| `scan.grepFiles` | 166 ms |
| `scanCodeForDsns` | 232 ms |
| `detectDsn.cold` | 4.78 ms |

## Why ship this

Init wizard grep has one caller: the Mastra server, which always sends `maxResults: 100`. For that workload the NEW adapter is **10× faster than rg** on common dense patterns and sidesteps the subprocess-dependency problem:

- Users without `rg` no longer fall back to a naive walk that ignores gitignore and scans binaries.
- No spawning, no stderr draining, no exit-code decoding, no parse-out-pipe-delimited-output.
- `GrepSearch` gains `caseInsensitive` + `multiline` — passthrough to the scan engine's native support.

The correctness tax (nested `.gitignore` respected, wider skip list, binaries filtered) is paid once per scan regardless of match density. On the init wizard's capped workload it's noise; on exhaustive scans it's ~500 ms on 10k files vs rg's ~300 ms.

## Test plan

- [x] `bunx tsc --noEmit` — clean
- [x] `bun run lint` — clean (1 pre-existing warning in `src/lib/formatters/markdown.ts:281`, not touched by this PR)
- [x] `bun test test/lib/init/ test/lib/scan/` — **352 pass, 0 fail**
- [x] Rebased onto latest main (includes #791, #804, #805, #807)
- [x] `bun run /tmp/bench-init-tools.ts` — numbers in table above

## What this PR does NOT change

- `GrepPayload` / `GlobPayload` — unchanged structurally, only `GrepSearch` gains optional fields
- `src/lib/init/tools/registry.ts` — tool dispatch unchanged
- `src/lib/init/tools/shared.ts` — `safePath` + `validateToolSandbox` unchanged
- Response shape — `{ ok: true, data: { results: [{pattern, matches, truncated}] } }` identical

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


betegon (approved):
thanks for this. works great with init!

PR #802, opened 2026-04-21
fix(init): release /dev/tty handle on all exit paths

### LINKED ISSUE #798: close stdin and exit cleanly when ai setup init fails
- ai-assisted setup can fail with `Failed to analyze project structure: Forbidden`
- when setup fails, the terminal session can remain stuck instead of exiting cleanly
- likely related behavior from thread context: stdin appears to be closed explicitly on successful init completion, but not on failure paths

steps observed:
- start the setup wizard that analyzes the project and configures Sentry
- hit a failure during analysis with `Forbidden`
- after the failure, the terminal remains stuck

notes:
- the `Forbidden` error itself was resolved separately
- the remaining bug is the terminal/escape behavior after a failed init path

this issue is intentionally sanitized and excludes personal or thread-specific details.

betegon:
## Summary
- The root cause of the post-wizard shell hang isn't the install script or a generic stdin close problem. It's the fresh `/dev/tty` `ReadStream` opened by `forwardFreshTtyToStdin`, which `sentry init` owned but never explicitly tore down.
- The old `process.exit` workaround only covered the success path. This change centralizes teardown in the wizard runtime so success, cancel, and error all release the same init-owned refs.
- The fix also stops the wizard spinner on cancellation and error paths so its interval does not keep the event loop alive.

## Changes
- `src/lib/init/stdin-reopen.ts`: replace the boolean install flag with stored forwarding state and add `closeFreshTtyForwarding()` to remove listeners, destroy the fresh `/dev/tty` stream, restore patched `process.stdin` methods, and clear the temporary `isTTY` backfill.
- `src/lib/init/wizard-runner.ts`: wrap the wizard in `try/finally`, call `closeFreshTtyForwarding()` on every exit path, and stop the spinner before early returns/rethrows on cancel and error paths.
- `src/commands/init.ts`: remove the success-only `process.exit(process.exitCode ?? 0)` workaround and update the comment to reflect wizard-owned teardown.
- `src/lib/version-check.ts` and `test/commands/init.test.ts`: remove the stale fetch-keepalive rationale.
- Tests: add `test/lib/init/stdin-reopen.test.ts` and extend wizard-runner coverage for success, cancellation, and tool-error teardown behavior.

## Test plan
- [x] `bun test test/lib/init/stdin-reopen.test.ts test/lib/init/wizard-runner.test.ts test/commands/init.test.ts`
- [x] `bunx @biomejs/biome check src/commands/init.ts src/lib/init/stdin-reopen.ts src/lib/init/wizard-runner.ts src/lib/version-check.ts test/commands/init.test.ts test/lib/init/wizard-runner.test.ts test/lib/init/stdin-reopen.test.ts`
- [ ] `bun run typecheck` (currently blocked in this worktree because `generate:command-docs` fails before TS with missing `src/generated/api-schema.json`)
- [ ] Manual: `curl -fsSL https://cli.sentry.dev/install | SENTRY_INIT=1 bash -s`
- [ ] Manual: force an init failure path and confirm the shell returns immediately

Fixes #798.

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -269,7 +269,9 @@ export const initCommand = buildCommand<
       warmOrgDetection(targetDir);
     }
 
-    // 6. Run the wizard
+    // 6. Run the wizard. It owns the temporary `/dev/tty` forwarding used
+    //    for `curl | bash` flows and tears it down on every exit path, so
+    //    init can return naturally once its own refs are released.
     await runWizard({
       directory: targetDir,
       yes: flags.yes,
@@ -279,12 +281,5 @@ export const initCommand = buildCommand<
       org: explicitOrg,
       project: explicitProject,
     });
-
-    // Force exit after the wizard completes. `sentry init` is a terminal
-    // command, and Bun's global fetch dispatcher (used by MastraClient) can
-    // hold keep-alive sockets open past the wizard, leaving the libuv loop
-    // alive and the shell appearing to hang. `process.exit` flushes stdio
-    // and releases those handles unconditionally.
-    process.exit(process.exitCode ?? 0);
   },
 });
--- a/src/lib/init/stdin-reopen.ts
+++ b/src/lib/init/stdin-reopen.ts
@@ -18,7 +18,36 @@
 import { openSync } from "node:fs";
 import { isatty, ReadStream } from "node:tty";
 
-let installed = false;
+/**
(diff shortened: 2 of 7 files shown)


BYK:
Don't think this is the right approach. We should identify the dangling refs. Closing `stdin` should be the first step.

BYK (approved):
This looks pretty good!

betegon:
@BYK this may be the first time i say this to you but, brooooo

PR #806, opened 2026-04-21
fix(dsn): skip non-regular env files during detection

elucid:
## Why

We use this [.env feature from 1password](https://1password.com/blog/1password-environments-env-files-public-beta) so our .env files are named pipes. Running `sentry` cli in directories where this type of file mounted (or symlinked) hangs because it tries to read .env to see if theres a `SENTRY_DSN` in there. This PR detects this case and skips reading. I figure this is nicer than doing a blocking read on the file to trigger the auth dialog from 1pw.

If you'd prefer a different fix, let me know. Just eager to work around this because this new cli is so useful!

## Summary

Skip non-regular `.env*` files during DSN detection to avoid hangs on FIFO-backed env files (including 1Password symlink-to-FIFO env files).

This applies the guard to:
- explicit env-file scanning
- project-root walk-up checks
- cached DSN source verification

It also adds regression coverage for top-level `detectDsn()` and for cached verification when an env file becomes a symlinked FIFO.

## Testing

- `bun run test`
- `bun run lint` *(reports one unrelated pre-existing warning in `src/lib/formatters/markdown.ts`)*

### DIFF
--- a/src/lib/dsn/detector.ts
+++ b/src/lib/dsn/detector.ts
@@ -35,6 +35,7 @@ import {
   detectFromEnvFiles,
   extractDsnFromEnvContent,
 } from "./env-file.js";
+import { isRegularFile } from "./fs-utils.js";
 import { createDetectedDsn, createDsnFingerprint, parseDsn } from "./parser.js";
 import { findProjectRoot } from "./project-root.js";
 import type {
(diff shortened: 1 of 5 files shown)


BYK:
Thanks so much! I'm refactoring file walkers so holding off for just a bit before I land that and rebase/refactor your patch on top of it

BYK (approved):
Thanks a lot!

PR #815, opened 2026-04-22
perf(init): tighten list-dir hot loop + emit POSIX paths

BYK:
## Summary

Focused code-audit cleanup of `src/lib/init/tools/list-dir.ts`.
No behavior changes on POSIX hosts; Windows hosts now get a
consistent wire contract.

### Correctness: emit POSIX-separator paths

`path.relative(cwd, ...)` returns native separators, so Windows
hosts were sending paths like \`src\\app.ts\` on the Mastra wire.
Two downstream callers had \`replaceAll(\"\\\\\", \"/\")\` patches
to cope — one (\`workflow-inputs.ts::preReadCommonFiles\`, the
direct list-dir consumer) is cleaned up here; the other
(\`formatters.ts\`, normalizing data from a different source) stays
as-is.

The existing \`filesystem-tools.test.ts\` already asserted
\`src/app.ts\` (POSIX) and would have failed on Windows pre-fix.

### Perf: prune the hot loop

Per-entry changes in \`walkDirectory\`:

1. \`dir + NATIVE_SEP + entry.name\` instead of \`path.join(...)\` — manual concat is ~10× faster in V8 since inputs are already clean.
2. \`abs.slice(state.cwdPrefixLen)\` instead of \`path.relative(...)\` — O(1) slice instead of a parse-and-allocate call, since \`dir\` always starts with \`cwd + sep\`.
3. \`cwdPrefixLen\` cached once at walk entry rather than recomputed per dirent.
4. Dropped the redundant mid-loop \`reachedWalkLimit\` call (the function-entry guard handles \`depth\`; an inline \`state.entries.length >= state.maxEntries\` check handles the cap).
5. Same manual-concat fix applied to the recursion call site.

Microbench on a 500-entry recursive listing (30 runs, 3 warmup):

| Version | p50 |
|---|---:|
| Before | 1.705 ms |
| After | **1.130 ms** |

−34% at p50. Absolute numbers are small in both cases because
list-dir isn't on a hot path (called once per wizard step), but
cleanup stacked on top of the correctness fix.

## New test coverage

Added \`test/lib/init/tools/list-dir.test.ts\` with 11 dedicated tests. Pre-PR, list-dir was covered by a single smoke test in \`filesystem-tools.test.ts\`.

## Test plan

- [x] \`bunx tsc --noEmit\` — clean
- [x] \`bun run lint\` — clean (1 pre-existing markdown.ts warning, unrelated)
- [x] \`bun test test/lib/init/\` — **135 pass, 0 fail** (+11 new)
- [x] \`bun test --timeout 15000 test/lib test/commands test/types\` — **5664 pass, 0 fail**
- [x] \`bun test test/isolated\` — 138 pass
- [x] Microbench before/after — −34% p50 on 500-entry fixture

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


src/lib/init/tools/list-dir.ts
```diff
@@ -46,64 +58,68 @@ async function readDirEntries(dir: string): Promise<fs.Dirent[]> {
   }
 }
 
+function shouldRecurseInto(entry: fs.Dirent, state: WalkState): boolean {
+  return (
+    state.recursive &&
+    entry.isDirectory() &&
+    !entry.isSymbolicLink() &&
+    !entry.name.startsWith(".") &&
+    entry.name !== "node_modules"
```
BYK: We should lift `node_modules` into a constant. I'd say it should be an app-level constant as I think we use it in another place too

BYK: Good call. Lifted to `NODE_MODULES_DIRNAME` in `src/lib/constants.ts` in [ce213c4a](https://github.com/getsentry/cli/commit/ce213c4a) and replaced the three discrete-concept sites: `list-dir.ts` (skip target), `sourcemap/inject.ts` (skip target), and `upgrade.ts` (path-segment sentinel for package-manager detection).

Left `src/lib/scan/constants.ts` alone — its `DEFAULT_SKIP_DIRS` list has `"node_modules"` as one of a dozen bare literals alongside `"dist"`, `"build"`, `"target"`, etc. Pulling out just one entry would be inconsistent with the rest of the array.

src/lib/init/tools/list-dir.ts
```diff
@@ -1,43 +1,56 @@
...
 import type { InitToolDefinition } from "./types.js";
 
+/** Native → POSIX separator. Identity on POSIX hosts. */
+const NATIVE_IS_POSIX = path.sep === "/";
+const NATIVE_SEP = path.sep;
+
+/** Normalize a native-separator path to POSIX. No-op on POSIX hosts. */
+function toPosix(nativePath: string): string {
+  return NATIVE_IS_POSIX ? nativePath : nativePath.replaceAll(NATIVE_SEP, "/");
+}
```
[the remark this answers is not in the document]
BYK: Good catch. Swapped the inline helper for the shared `normalizePath` from `src/lib/scan/index.js` in [54e9723d](https://github.com/getsentry/cli/commit/54e9723d). No new cross-module dep — `init/tools/` already pulls from `scan/` elsewhere (e.g., `glob.ts` → `collectGlob`).

PR #857, opened 2026-04-28
feat: add `sentry explore` command for aggregate event queries

### LINKED ISSUE #853: add aggregate queries for spike analysis
the cli can't currently query aggregate data to verify spikes directly, which pushes it toward slow issue-by-issue/event-by-event workarounds.

- use case: inspect whether a reported spike is real from the cli, ideally by querying aggregate results instead of fetching many individual issues or events in a loop
- likely desired capability: query discover-style aggregates or event stats from the cli
- concrete example shared: using discover for errors with fields like `title`, grouped over time with metrics such as `count()` and `count_unique(user)` to inspect a spike window
- current limitation observed: there doesn't seem to be a cli path for this today, so the fallback was pulling down up to a thousand issues individually, which is slow and awkward for this workflow
- related context from the thread: a relative `--period 1h` query came back empty, so explicit UTC windows were used instead to tie comparisons to the actual spike window

Action taken on behalf of Josh Ferge.

BYK:
## Summary

Add a new top-level `sentry explore` command that wraps the Sentry Explore/Events API for aggregate queries — the missing capability for spike analysis from the CLI (#853).

## Usage

```bash
# Top errors in a project (last 24h)
sentry explore sentry/cli

# Custom fields and aggregates
sentry explore sentry/cli -F title -F "count()" -F "count_unique(user)" --period 1h

# Span performance — p50/p95 by route
sentry explore my-org/cli -F span.op -F "p50(span.duration)" -F "p95(span.duration)" \
  --dataset spans --sort "-count()"

# Custom metrics
sentry explore my-org/cli -F transaction -F "avg(measurements.fcp)" --dataset metrics

# Logs by severity
sentry explore my-org/cli -F severity -F "count()" --dataset logs --period 1h

# JSON for scripting
sentry explore my-org/cli -F title -F "count()" --json | jq '.data[:5]'
```

## Key design decisions

- **Top-level command** (`sentry explore`), not a subcommand — maps to Sentry's Explore UI, crosses dataset boundaries
- **`--field` / `-F` repeatable flag** for arbitrary fields and aggregates — each `-F` adds a column; column order in output preserves user input (the API returns `meta.fields` in non-deterministic order)
- **`--dataset`** — `errors` (default), `spans`, `metrics`, `logs`. Deprecated `transactions` and `discover` still work but are hidden from help
- **Target resolution**: `<org>/<project>` auto-injects `project:<slug>` into the query (via `buildProjectQuery`), `<org>/` queries all projects, bare `<slug>` fuzzy-searches across orgs
- **Sort only on `spans` dataset** — other datasets reject `sort` with 400; we silently omit it (matching `queryWidgetTable` behavior) and warn when the user explicitly passed `--sort`
- **Default: top errors** — `dataset=errors`, `field=[title, count()]`, last 24h — the most common spike-analysis starting point
- **`--limit` capped at 100** for now (Events API silently caps `per_page` server-side). Auto-pagination tracked in #861

## Shared helpers added

Driven by review feedback to avoid duplicating logic across commands:

- `src/lib/formatters/meta-table.ts` (NEW) — `buildMetaColumns()` + `formatCellValue()` + `NUMERIC_FIELD_TYPES` for rendering tables driven by Sentry Events API `meta.fields`/`meta.units`. Reusable for any future command rendering Discover/Explore results.
- `src/lib/list-command.ts` — `appendQueryHint()` and `appendSortHint()` helpers for consistent pagination-hint construction (sort values are quoted for shell safety).

## Files

| File | Description |
|------|-------------|
| `src/lib/api/discover.ts` | NEW — `queryEvents()` wraps `GET /organizations/{org}/events/` with `EventsTableResponseSchema` validation |
| `src/lib/api-client.ts` | Re-export `queryEvents` + `API_MAX_PER_PAGE` |
| `src/commands/explore.ts` | NEW — command definition, target resolution, dynamic formatting, pagination |
| `src/app.ts` | Register `explore` route |
| `src/lib/formatters/meta-table.ts` | NEW — extracted shared table-rendering helpers |
| `src/lib/list-command.ts` | Add `appendQueryHint` + `appendSortHint` |
| `test/commands/explore.test.ts` | 21 tests — target resolution, API params, sort, output, pagination, column order |
| `test/lib/api/discover.test.ts` | 14 tests — URL construction, repeated `field` params, schema validation, Link-header pagination |
| `docs/src/fragments/commands/explore.md` | Usage examples and target pattern docs |

## Follow-ups

- #860 — Generalize `resolveOrgProjectTarget` to support `org-all` (replaces local `resolveExploreTarget`)
- #861 — Auto-paginate `queryEvents` to support `--limit > 100`

Closes #853

### DIFF
--- a/plugins/sentry-cli/skills/sentry-cli/references/explore.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/explore.md
@@ -17,11 +17,11 @@ Query aggregate event data (Explore)
 
 **Flags:**
 - `-F, --field <value>... - API field or aggregate (repeatable). E.g., title, "count()", "p50(transaction.duration)"`
-- `-d, --dataset <value> - Dataset to query (errors, transactions, spans, discover) - (default: "errors")`
+- `-d, --dataset <value> - Dataset to query (errors, transactions, spans, metrics, discover) - (default: "errors")`
 - `-q, --query <value> - Search query (Sentry search syntax)`
 - `-s, --sort <value> - Sort field (prefix with - for desc, e.g., "-count()")`
 - `-n, --limit <value> - Number of rows (1-1000) - (default: "25")`
-- `-t, --period <value> - Time range: "7d", "2026-03-01..2026-04-01", ">=2026-03-01" - (default: "7d")`
+- `-t, --period <value> - Time range: "7d", "2026-03-01..2026-04-01", ">=2026-03-01" - (default: "24h")`
 - `-f, --fresh - Bypass cache, re-detect projects, and fetch fresh data`
 - `-c, --cursor <value> - Navigate pages: "next", "prev", "first" (or raw cursor string)`
 
--- a/plugins/sentry-cli/skills/sentry-cli/references/explore.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/explore.md
@@ -20,7 +20,7 @@ Query aggregate event data (Explore)
 - `-d, --dataset <value> - Dataset to query (errors, spans, metrics, logs) - (default: "errors")`
 - `-q, --query <value> - Search query (Sentry search syntax)`
 - `-s, --sort <value> - Sort field (prefix with - for desc, e.g., "-count()")`
-- `-n, --limit <value> - Number of rows (1-1000) - (default: "25")`
+- `-n, --limit <value> - Number of rows (1-100) - (default: "25")`
 - `-t, --period <value> - Time range: "7d", "2026-03-01..2026-04-01", ">=2026-03-01" - (default: "24h")`
 - `-f, --fresh - Bypass cache, re-detect projects, and fetch fresh data`
 - `-c, --cursor <value> - Navigate pages: "next", "prev", "first" (or raw cursor string)`
@@ -48,22 +48,21 @@ sentry explore my-org/cli -F title -F "count()" -F "count_unique(user)" \
 sentry explore my-org/cli -F title -F "count()" \
   -q "error.type:TypeError" --period 1h
 
-# Transaction p50/p95 by endpoint
-sentry explore my-org/cli -F transaction \
-  -F "p50(transaction.duration)" -F "p95(transaction.duration)" \
-  --dataset transactions --period 1h
+# Span operation latency by route
+sentry explore my-org/cli -F span.op -F "p50(span.duration)" \
+  -F "p95(span.duration)" --dataset spans --period 1h
 
-# Slowest transactions
-sentry explore my-org/cli -F transaction -F "avg(transaction.duration)" \
-  --dataset transactions
-
-# Span operations by count
-sentry explore my-org/cli -F span.op -F "count()" \
-  --dataset spans --period 1h
-
-# Sort by count (sort is supported on spans dataset)
+# Top spans by count
 sentry explore my-org/cli -F span.op -F "count()" \
   --dataset spans --sort "-count()"
+
+# Custom metric aggregations
+sentry explore my-org/cli -F transaction -F "avg(measurements.fcp)" \
+  --dataset metrics --period 24h
+
+# Log severity counts in the last hour
+sentry explore my-org/cli -F severity -F "count()" \
+  --dataset logs --period 1h
 
 # Pipe to jq for filtering
 sentry explore my-org/cli -F title -F "count()" --json | jq '.data[:5]'
--- a/plugins/sentry-cli/skills/sentry-cli/references/explore.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/explore.md
@@ -17,7 +17,7 @@ Query aggregate event data (Explore)
 
 **Flags:**
 - `-F, --field <value>... - API field or aggregate (repeatable). E.g., title, "count()", "p50(transaction.duration)"`
-- `-d, --dataset <value> - Dataset to query (errors, transactions, spans, metrics, discover) - (default: "errors")`
+- `-d, --dataset <value> - Dataset to query (errors, spans, metrics, logs) - (default: "errors")`
 - `-q, --query <value> - Search query (Sentry search syntax)`
 - `-s, --sort <value> - Sort field (prefix with - for desc, e.g., "-count()")`
 - `-n, --limit <value> - Number of rows (1-1000) - (default: "25")`
--- a/plugins/sentry-cli/skills/sentry-cli/references/explore.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/explore.md
@@ -1,0 +1,76 @@
+---
+name: sentry-cli-explore
+version: 0.30.0-dev.0
+description: Query aggregate event data (Explore)
+requires:
+  bins: ["sentry"]
+  auth: true
+---
+
+# Explore Commands
+
+Query aggregate event data (Explore)
+
+### `sentry explore <target>`
+
+Query aggregate event data (Explore)
+
+**Flags:**
+- `-F, --field <value>... - API field or aggregate (repeatable). E.g., title, "count()", "p50(transaction.duration)"`
+- `-d, --dataset <value> - Dataset to query (errors, transactions, spans, discover) - (default: "errors")`
+- `-q, --query <value> - Search query (Sentry search syntax)`
+- `-s, --sort <value> - Sort field (prefix with - for desc, e.g., "-count()")`
+- `-n, --limit <value> - Number of rows (1-1000) - (default: "25")`
+- `-t, --period <value> - Time range: "7d", "2026-03-01..2026-04-01", ">=2026-03-01" - (default: "7d")`
+- `-f, --fresh - Bypass cache, re-detect projects, and fetch fresh data`
+- `-c, --cursor <value> - Navigate pages: "next", "prev", "first" (or raw cursor string)`
+
+**Examples:**
+
+```bash
+# Top errors in the last 24 hours, scoped to a project
+sentry explore my-org/cli
+
+# All projects in an org
+sentry explore my-org/
+
+# Bare project slug (searches across orgs)
+sentry explore cli
+
+# Auto-detect from DSN/config
+sentry explore
+
+# Errors with user impact for a specific UTC window
+sentry explore my-org/cli -F title -F "count()" -F "count_unique(user)" \
+  --period "2024-01-15T00:00:00Z/2024-01-16T00:00:00Z"
+
+# Filter by specific error type (combines with auto-injected project filter)
+sentry explore my-org/cli -F title -F "count()" \
+  -q "error.type:TypeError" --period 1h
+
+# Transaction p50/p95 by endpoint
+sentry explore my-org/cli -F transaction \
+  -F "p50(transaction.duration)" -F "p95(transaction.duration)" \
+  --dataset transactions --period 1h
+
+# Slowest transactions
+sentry explore my-org/cli -F transaction -F "avg(transaction.duration)" \
+  --dataset transactions
+
+# Span operations by count
+sentry explore my-org/cli -F span.op -F "count()" \
+  --dataset spans --period 1h
+
+# Sort by count (sort is supported on spans dataset)
+sentry explore my-org/cli -F span.op -F "count()" \
+  --dataset spans --sort "-count()"
+
+# Pipe to jq for filtering
+sentry explore my-org/cli -F title -F "count()" --json | jq '.data[:5]'
+
+# Get raw data for analysis
+sentry explore my-org/cli -F title -F "count()" -F "count_unique(user)" \
+  --json --limit 100
+```
+
+All commands also support `--json`, `--fields`, `--help`, `--log-level`, and `--verbose` flags.
--- a/src/commands/explore.ts
+++ b/src/commands/explore.ts
@@ -154,7 +154,6 @@ function parseDataset(value: string): string {
   return resolved;
 }
 
-/** Parse --limit flag with range validation */
 /**
  * Parse --limit flag. Capped at `API_MAX_PER_PAGE` (100) since the Events
  * API silently caps `per_page` server-side and `queryEvents` does not yet
--- a/src/commands/explore.ts
+++ b/src/commands/explore.ts
@@ -28,7 +28,6 @@ import { type Column, formatTable } from "../lib/formatters/table.js";
 import {
   buildListCommand,
   LIST_MAX_LIMIT,
-  LIST_PERIOD_FLAG,
   PERIOD_ALIASES,
   paginationHint,
 } from "../lib/list-command.js";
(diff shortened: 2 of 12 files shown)


src/commands/explore.ts
```diff
@@ -0,0 +1,579 @@
...
+  error: "errors",
+  transactions: "transactions",
+  transaction: "transactions",
+  spans: "spans",
+  span: "spans",
+  discover: "discover",
+};
+
+/** Canonical dataset names for display */
+const VALID_DATASETS = ["errors", "transactions", "spans", "discover"];
```
JoshFerge: we also have metrics :D

src/commands/explore.ts
```diff
@@ -0,0 +1,579 @@
...
+        brief: 'Sort field (prefix with - for desc, e.g., "-count()")',
+        optional: true,
+      },
+      limit: {
+        kind: "parsed",
+        parse: parseLimit,
+        brief: `Number of rows (1-${LIST_MAX_LIMIT})`,
+        default: "25",
+      },
+      period: LIST_PERIOD_FLAG,
```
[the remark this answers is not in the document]
BYK: Good catch. Fixed: replaced `LIST_PERIOD_FLAG` (which defaults to 7d) with a custom inline flag that uses `DEFAULT_PERIOD = "24h"`, matching the documented intent.

src/commands/explore.ts
BYK: Added! `--dataset metrics` now maps to `metricsEnhanced` at the API level, matching the dashboard widget dataset mapping.

src/commands/explore.ts
```diff
@@ -0,0 +1,579 @@
...
+  }
+  // Include --field flags when non-default
+  const fieldList = flags.field ?? [];
+  const currentFieldStr = fieldList.join(",");
+  if (currentFieldStr !== DEFAULT_FIELDS.join(",") && fieldList.length > 0) {
+    for (const f of fieldList) {
+      parts.push(`-F "${f}"`);
+    }
+  }
+  appendPeriodHint(parts, flags.period, DEFAULT_PERIOD);
```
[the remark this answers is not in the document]
BYK: Fixed together with the Cursor Bugbot finding — defined a custom period flag with `default: DEFAULT_PERIOD` ("24h") instead of reusing `LIST_PERIOD_FLAG` ("7d").

plugins/sentry-cli/skills/sentry-cli/references/explore.md
```diff
@@ -0,0 +1,76 @@
...
+
+Query aggregate event data (Explore)
+
+### `sentry explore <target>`
+
+Query aggregate event data (Explore)
+
+**Flags:**
+- `-F, --field <value>... - API field or aggregate (repeatable). E.g., title, "count()", "p50(transaction.duration)"`
+- `-d, --dataset <value> - Dataset to query (errors, transactions, spans, metrics, discover) - (default: "errors")`
```
JoshFerge: i would just hide discover / transactions. so it'd be (errors, spans, metrics, logs)

JoshFerge: (transactions/discover are being deprecated... or at least we're trying to encourage people to use the new apis based on spans)

src/commands/explore.ts
```diff
@@ -0,0 +1,593 @@
...
+// ---------------------------------------------------------------------------
+
+/** Append active non-default flags to a base command string */
+function appendFlagHints(
+  base: string,
+  flags: Pick<ExploreFlags, "dataset" | "sort" | "query" | "period" | "field">
+): string {
+  const parts: string[] = [];
+  if (flags.dataset !== DEFAULT_DATASET) {
+    parts.push(`--dataset ${flags.dataset}`);
```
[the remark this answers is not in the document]
BYK: Good catch. Added `API_TO_USER_DATASET` reverse map — pagination hints and the human header now emit the user-facing name (`metrics`) instead of the API-level name (`metricsEnhanced`).

plugins/sentry-cli/skills/sentry-cli/references/explore.md
BYK: Done — updated to show (errors, spans, metrics, logs) in help/validation. Also added `logs` dataset which was missing. `transactions`/`discover` still work if passed explicitly but are hidden from the display list.

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+  transaction: "transactions",
+  discover: "discover",
+};
+
+/**
+ * User-facing dataset names shown in `--help` and validation errors.
+ * Deprecated datasets (transactions, discover) are omitted from the display
+ * list but still work as `--dataset` values via `DATASET_ALIASES`.
+ */
+const VALID_DATASETS = ["errors", "spans", "metrics", "logs"];
```
BYK: Isn't this better as a `Set()`?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+
+/**
+ * Reverse map from API-level dataset name → canonical user-facing name.
+ * Used by pagination hints so they emit `--dataset metrics` not `--dataset metricsEnhanced`.
+ */
+const API_TO_USER_DATASET: Record<string, string> = Object.fromEntries(
```
BYK: Why not a `Map()` here?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+// Formatting helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Format a cell value based on its type metadata.
+ *
+ * Uses type hints from the API response `meta.fields` to apply appropriate
+ * formatting: numbers with grouping, durations with units, etc.
+ */
+function formatCellValue(
```
BYK: Don't we have this function somewhere else already?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+  }
+  return escapeMarkdownCell(String(value));
+}
+
+/**
+ * Build dynamic table columns from the API response metadata.
+ *
+ * Each field in the response becomes a column. Numeric fields are right-aligned.
+ */
+function buildColumns(
```
BYK: I wonder if this already exists somewhere in the codebase. If not, mighte be better to place it in a more generic place rather than confining to `explore` as it looks quite generic and useful.

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+    // Emit user-facing name, not API-level name (e.g. "metrics" not "metricsEnhanced")
+    const displayDataset = API_TO_USER_DATASET[flags.dataset] ?? flags.dataset;
+    parts.push(`--dataset ${displayDataset}`);
+  }
+  if (flags.sort) {
+    parts.push(`--sort ${flags.sort}`);
+  }
+  if (flags.query) {
+    parts.push(`-q "${flags.query}"`);
+  }
```
BYK: We should probably have this functionality as a generic helper as it is in many places?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+// ---------------------------------------------------------------------------
+
+/** Resolved explore target: org slug + optional project filter */
+type ExploreTarget = {
+  org: string;
+  project?: string;
+};
+
+/** Usage hint shown when target resolution fails */
+const USAGE_HINT = "sentry explore [<org>/[<project>]]";
```
BYK: We say `[<org>/[<project]]` here but `<target>` in the help text? These should be consistent, right?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+/**
+ * Resolve the explore target to an org slug and optional project filter.
+ *
+ * Semantics:
+ * - `<org>/<project>` → explicit org with project filter (adds `project:slug` to query)
+ * - `<org>/` → org-all mode (no project filter)
+ * - `<project>` (bare) → fuzzy-search across all orgs for the project
+ * - undefined → auto-detect from DSN/config
+ */
+async function resolveExploreTarget(
```
BYK, line 317: Can we not use a common generic helper for this as it is a well-esablished pattern across the entire CLI at this point?

src/commands/explore.ts
```diff
@@ -0,0 +1,609 @@
...
+    // Pagination hints preserve the original target shape (org/ vs org/project)
+    const baseTarget = project ? `${org}/${project}` : `${org}/`;
+    const nav = paginationHint({
+      hasPrev,
+      hasMore,
+      prevHint: appendFlagHints(`sentry explore ${baseTarget} -c prev`, flags),
+      nextHint: appendFlagHints(`sentry explore ${baseTarget} -c next`, flags),
+    });
+
+    const hint = buildResultHint(response.data.length, nav);
```
BYK, line 545: I don't see us mentioning the newly added `-f for fresh results` hint anywhere? (you may need to rebase onto latest main to see this)

src/commands/explore.ts
BYK: Done — converted to Set.

src/commands/explore.ts
BYK: Done — converted to Map.

src/commands/explore.ts
BYK: No identical helper exists. The low-level blocks (formatNumber, appendUnitSuffix) exist in numbers.ts but nothing composes them with a fieldType switch. Extracted to src/lib/formatters/meta-table.ts as the shared formatCellValue.

src/commands/explore.ts
BYK: Extracted to src/lib/formatters/meta-table.ts as buildMetaColumns() alongside formatCellValue and NUMERIC_FIELD_TYPES. Reusable for any command rendering dynamic Events API results.

src/commands/explore.ts
BYK: Added appendQueryHint and appendSortHint to list-command.ts as shared building blocks (same shape as appendPeriodHint). Each command still composes its own appendFlagHints since flag sets differ, but the common parts are now shared.

src/commands/explore.ts
BYK: Fixed — USAGE_HINT now uses "sentry explore <target>" to match the positional placeholder.

src/commands/explore.ts
BYK: The closest helper is resolveOrgProjectTarget but it rejects org-all mode (throws ContextError) and returns a required project. Explore needs org-all as its primary mode. This is thin glue over existing shared helpers (parseOrgProjectArg + resolveProjectBySlug + resolveOrg). Generalizing with an allowOrgAll flag would be a good follow-up.

src/commands/explore.ts
BYK: This is being added in #859, not live on main yet. Will pick it up when rebasing.

src/commands/explore.ts
```diff
@@ -0,0 +1,550 @@
...
+  flags: Pick<ExploreFlags, "dataset" | "sort" | "query" | "period" | "field">
+): string {
+  const parts: string[] = [];
+  if (flags.dataset !== DEFAULT_DATASET) {
+    // Emit user-facing name, not API-level name (e.g. "metrics" not "metricsEnhanced")
+    const displayDataset =
+      API_TO_USER_DATASET.get(flags.dataset) ?? flags.dataset;
+    parts.push(`--dataset ${displayDataset}`);
+  }
+  appendSortHint(parts, flags.sort);
```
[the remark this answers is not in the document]
BYK: The auto-computed sort is intentionally omitted from hints. For spans, the user passes --sort explicitly if they want a specific order. The auto-default (-count()) is recomputed from the same --field list each time, so pages are consistent. Including it would clutter every hint with a sort the user never typed.

src/commands/explore.ts
```diff
@@ -0,0 +1,550 @@
...
+function resolveSort(
+  fieldList: string[],
+  dataset: string,
+  explicitSort?: string
+): string | undefined {
+  const sort =
+    explicitSort ??
+    (findFirstAggregate(fieldList)
+      ? `-${findFirstAggregate(fieldList)}`
+      : undefined);
```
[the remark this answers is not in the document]
BYK: Fixed — stored in a local variable.

src/lib/api-client.ts
```diff
@@ -26,6 +26,10 @@ export {
   queryAllWidgets,
   updateDashboard,
 } from "./api/dashboards.js";
+export {
+  type ExploreQueryOptions,
+  queryEvents,
+} from "./api/discover.js";
```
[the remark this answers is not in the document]
BYK: Removed the unused type re-export.

src/lib/list-command.ts
```diff
@@ -263,6 +263,39 @@ export function paginationHint(opts: {
...
+ * and command-specific enum values (e.g. `"date" | "duration"`).
+ */
+export function appendSortHint(
+  parts: string[],
+  sort: string | undefined,
+  defaultSort?: string
+): void {
+  if (sort && sort !== defaultSort) {
+    parts.push(`--sort ${sort}`);
+  }
```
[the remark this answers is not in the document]
BYK: Good catch — quoted the sort value in appendSortHint so values like -count() are shell-safe.

src/commands/explore.ts
```diff
@@ -0,0 +1,547 @@
...
+  }
+  return ordered;
+}
+
+/** Format explore results for human-readable terminal output */
+function formatExploreHuman(data: ExploreData): string {
+  if (data.data.length === 0) {
+    return data.hasMore
+      ? "No results on this page."
+      : "No results matched the query.";
```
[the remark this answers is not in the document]
BYK: Fixed — empty-page hint now only shows navigation without repeating "No results on this page." (the human formatter already shows it).

src/commands/explore.ts
```diff
@@ -0,0 +1,552 @@
...
+  }
+  return resolved;
+}
+
+/** Parse --limit flag with range validation */
+/**
+ * Parse --limit flag. Capped at `API_MAX_PER_PAGE` (100) since the Events
+ * API silently caps `per_page` server-side and `queryEvents` does not yet
+ * auto-paginate. Tracked for follow-up in #861.
+ */
```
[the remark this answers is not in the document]
BYK: Removed the orphaned JSDoc.

src/commands/explore.ts
```diff
@@ -0,0 +1,551 @@
...
+  const fieldList = flags.field ?? [];
+  const currentFieldStr = fieldList.join(",");
+  if (currentFieldStr !== DEFAULT_FIELDS.join(",") && fieldList.length > 0) {
+    for (const f of fieldList) {
+      parts.push(`-F "${f}"`);
+    }
+  }
+  appendPeriodHint(parts, flags.period, DEFAULT_PERIOD);
+  return parts.length > 0 ? `${base} ${parts.join(" ")}` : base;
+}
```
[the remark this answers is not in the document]
BYK: Good catch — added --limit to pagination hints when non-default.

PR #870, opened 2026-04-28
fix(api): guard against excessively long query strings in issue list (CLI-FA)

BYK:
## Summary
- Adds a 4KB query length guard in `listIssuesPaginated` that throws a helpful `ValidationError` before the request is sent
- Prevents the cryptic "400 Bad Request" error that 95 users (1589 events) hit when passing complex queries with many OR terms
- The error message suggests narrowing the search, reducing OR terms, or using `--period` to limit the time range

## Changes
- **`src/lib/api/issues.ts`**: Add `MAX_QUERY_LENGTH` constant (4096) and pre-flight check in `listIssuesPaginated`
- **`test/lib/api/issues.test.ts`**: Add tests for the query length guard (too long → ValidationError, within limit → passes through)

## Why 4KB?
Most HTTP servers/load balancers limit total URL length to ~8KB. With URL encoding overhead (~3x for non-ASCII/special chars) and other query params (project, cursor, sort, collapse, statsPeriod), 4KB for the query value alone is a safe threshold.

Fixes https://sentry.sentry.io/issues/7341195391/

### DIFF
--- a/src/lib/api/issues.ts
+++ b/src/lib/api/issues.ts
@@ -28,6 +28,17 @@ import {
 const TRAILING_SLASH_RE = /\/$/;
 
 /**
+ * Maximum length for the search query parameter (before URL encoding).
+ *
+ * Most HTTP servers/load balancers limit total URL length to ~8KB.
+ * With URL encoding overhead (~3x for non-ASCII) and other query params
+ * (project, cursor, sort, collapse, statsPeriod), 4KB for the query
+ * value alone is a safe threshold. Exceeding this causes the Sentry API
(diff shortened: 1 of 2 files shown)


BYK:
Closing: investigation of actual CLI-FA events reveals the 400 errors are caused by invalid Sentry search syntax from users/agents (e.g., `error.http.status_code:[401,403,429,500,)` with trailing comma/paren), not excessively long URLs. The existing error handling in `build400Detail` already propagates the API's parse error detail with suggestions. A query length guard is not the right fix.

Issue #600, opened 2026-03-30
Catch up with old sentry-cli


BYK:
# Gap Analysis: Old sentry-cli (Rust) vs New Sentry CLI (Bun/TS)

## Executive Summary

The two CLIs serve fundamentally different audiences:
- **Old CLI**: CI/CD build tool — releases, deploys, debug files, sourcemaps, cron monitors
- **New CLI**: Developer observability tool — issues, traces, spans, logs, dashboards, AI analysis

~60% of old CLI commands are absent from the new CLI, almost all CI/CD-oriented. However, several are legacy/niche and not worth porting. The critical gaps are **releases**, **deploys**, and **sourcemap upload enhancements**.

The biggest unbridgeable gap is **debug-files** — it depends on the `symbolic` Rust crate for native binary parsing (Mach-O, ELF, PDB, DWARF) which has no JS equivalent.

---

## 1. Commands Fully Present in New CLI ✅

These are at parity or exceed the old CLI:

| Old CLI | New CLI | Notes |
|---------|---------|-------|
| `login` | `auth login` | **Better** — OAuth device flow + token auth |
| `info` | `auth status` | **Better** — shows defaults, token metadata, org verification |
| `update` | `cli upgrade` | **Better** — delta patching, stable/nightly channels |
| `completions` | `cli setup` | **Better** — auto-detects shell, installs agent skills |
| `organizations list` | `org list` | **Better** — multi-region support, cached |
| `projects list` | `project list` | **Better** — 4-mode dispatch, auto-detect |
| `repos list` | `repo list` | At parity |
| `events list` | `issue list` / `trace list` | **Better** — old CLI's event list was very limited |

### New CLI features with no old CLI equivalent:
- `org view`, `project view`, `project create`, `project delete`
- `issue view`, `issue explain` (Seer AI), `issue plan` (Seer AI)
- `event view` (with full stack trace, breadcrumbs, span tree)
- `trace list/view/logs`, `span list/view`, `log list/view` (with `--follow`)
- `dashboard list/view/create`, `dashboard widget add/edit/delete`
- `api` (generic authenticated API requests, like `gh api`)
- `schema` (API schema browser)
- `init` (AI-powered project setup wizard)
- `auth whoami`, `auth refresh`, `auth token`
- `cli feedback`, `cli fix`
- `team list`, `trial list/start`

---

### 2a. `sourcemaps inject` — Missing sourceMappingURL handling

**Old CLI capabilities not in new CLI:**
[...]

PR #890, opened 2026-04-30
feat(sourcemap): add parity flags and sourceMappingURL following

BYK:
## Summary

Adds missing sourcemap flags for old sentry-cli parity and implements `sourceMappingURL` directive following. Ref #600.

### Sourcemap upload: 6 new flags

- **`--dist`** — Distribution identifier to disambiguate builds within a release
- **`--ext`** — Custom file extensions on upload (was only on inject)
- **`--no-rewrite`** — Upload files without injecting debug IDs
- **`--ignore` / `--ignore-file`** — Gitignore-style file exclusion (uses `ignore` npm package, already in devDeps)
- **`--strip-prefix`** — Remove explicit prefix from uploaded file URLs
- **`--strip-common-prefix`** — Auto-strip longest common directory prefix
- `--strip-prefix` and `--strip-common-prefix` are mutually exclusive (throws `ValidationError`)

### Sourcemap inject: `sourceMappingURL` following

Companion map discovery was previously strictly filename-based (`foo.js` → `foo.js.map`). Now also parses `//# sourceMappingURL=<url>`:

- Reads only the last 512 bytes of each file for efficiency
- Takes the **last** directive (spec-correct for concatenated bundles)
- Resolution order: convention naming first, then `sourceMappingURL` fallback
- Gracefully skips `data:` (inline) and `http:`/`https:` (remote) URLs

### Tests

- 13 new tests covering all flags, `sourceMappingURL` following, ignore patterns, strip-prefix, and mutual exclusion

### DIFF
--- a/plugins/sentry-cli/skills/sentry-cli/references/sourcemap.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/sourcemap.md
@@ -17,7 +17,7 @@ Inject debug IDs into JavaScript files and sourcemaps
 
 **Flags:**
 - `--ext <value> - Comma-separated file extensions to process (default: .js,.cjs,.mjs)`
-- `--ignore <value> - Glob pattern to exclude (gitignore-style, repeatable)`
+- `--ignore <value> - Comma-separated glob patterns to exclude (gitignore-style)`
 - `--ignore-file <value> - Path to a file with gitignore-style patterns to exclude`
 - `--dry-run - Show what would be modified without writing`
 - `--allow-empty - Exit successfully when no JS + sourcemap pairs are found (default: error out to catch silent build misconfigurations)`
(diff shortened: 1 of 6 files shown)


BYK:
Both Seer and Bugbot correctly identified that the `--strip-prefix` / `--strip-common-prefix` mutual exclusion check ran after `injectDirectory` had already modified files. Fixed in ae975ff2 — validation now runs before any file mutation.

BYK:
Added path containment guard in ff84b448 — sourceMappingURL-resolved paths that escape the upload directory are now skipped with a debug log.

BYK:
Fixed both findings in b81bb5b2:
1. Directory boundary check now uses `absDir + '/'` to prevent prefix collisions (e.g. `/dist` matching `/dist-backup`).
2. `sourcemapFilename` now uses `relative(dirname(jsPath), mapPath)` instead of `basename(mapPath)`, so cross-directory maps produce correct `Sourcemap` headers.

PR #911, opened 2026-05-04
feat(auth): Add bundled dev OAuth client ID

dcramer:
Bundle the public development OAuth client ID currently used in local env as the final fallback for OAuth flows. Local development and non-release binary builds no longer need a `.env.local` file just to authenticate against sentry.io, while runtime `SENTRY_CLIENT_ID` and build-time `SENTRY_CLIENT_ID_BUILD` still take precedence for self-hosted and production use.

**Packaging Guardrails**

Release binary builds and the npm bundle still require an explicit `SENTRY_CLIENT_ID`, so production distributions cannot accidentally ship the development OAuth app.

Validated locally with `bun test --timeout 15000 --isolate test/lib/security/refresh-token-poison.test.ts`, `bun run lint`, and `bun run typecheck`.

### DIFF
--- a/.cursor/rules/bun-cli.mdc
+++ b/.cursor/rules/bun-cli.mdc
@@ -138,8 +138,11 @@ await Bun.build({
   },
   define: {
     CLI_VERSION: JSON.stringify(version),
-    // Inject secrets at build time for npm distribution
-    SENTRY_CLIENT_ID_BUILD: JSON.stringify(process.env.SENTRY_CLIENT_ID ?? ""),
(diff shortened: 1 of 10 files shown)


BYK:
Superseded by #916, which takes a simpler approach: commit the production OAuth client ID directly as the fallback rather than introducing a separate dev ID and a `RELEASE_BUILD` guard. The client ID is public by design (Device Authorization Grant / RFC 8628, no client secret) and is already a plain repo variable in CI — not a secret.

Issue #907, opened 2026-05-03
Make session replay data agent-actionable


dcramer:
## Summary

Make Session Replay data in `sentry-cli` highly actionable for coding agents by adding first-class replay segment fetching, local caching, normalized event extraction, and inspection commands.

The CLI should **not** try to be the agent or own an `ask` command as the primary interface. Instead, it should become the replay data plane: a reliable way for agents to fetch replay segments, cache them, inspect DOM/rrweb/custom events, search timelines, and pull evidence windows around user actions. Agents can then compose those tools, plus a bundled skill that explains the replay data model, to answer questions such as:

- "When the user clicked X, what happened next?"
- "Where did the user spend the most time?"
- "What hangups caused the user to struggle?"
- "Were there failed requests, console errors, dead clicks, rage clicks, or DOM changes around this action?"

## Current State

The CLI currently has basic replay support:

- `sentry replay list` queries replay metadata.
- `sentry replay view` fetches replay detail, related issues/traces, and a very small activity preview from recording segments.
- `sentry explore --dataset replays` exposes replay index fields.

Relevant CLI files:

- `src/commands/replay/list.ts`
- `src/commands/replay/view.ts`
- `src/lib/api/replays.ts`
- `src/lib/formatters/replay.ts`
- `src/lib/replay-search.ts`
- `src/types/replay.ts`

The main limitation is that replay segments are treated as display garnish. `replay view` currently extracts only a handful of activity events from raw segments, capped to a tiny preview. There is no replay-specific local cache, no normalized event stream, no segment index, and no way for agents to inspect all DOM/rrweb/custom events.

There is also a likely correctness gap: the Sentry recording-segments endpoint is paginated, while the CLI currently downloads it as a single request. The frontend fetches segment pages with `per_page=100` until `count_segments` is exhausted. The CLI should mirror that behavior so long replays are fully available.

## Relevant Sentry/rrweb Model

Sentry replay data has several useful layers:

1. Replay metadata from org/project replay endpoints: duration, urls, counts, errors, traces, clicks, user/browser/os/sdk/device, tags, etc.
2. Recording segments from the project-scoped `recording-segments` endpoint. These are compressed/packed storage blobs returned as rrweb/custom JSON when downloaded.
3. rrweb events: full snapshots, incremental DOM mutations, mouse interactions, inputs, scrolls, viewport changes, media interactions, console logs, and more.
[...]

PR #913, opened 2026-05-04
feat(replay): Add agent-readable replay timelines

dcramer:
Add path-aware replay discovery and deterministic replay summaries so agents can investigate user behavior from Session Replay without reading raw rrweb frames.

**Replay Discovery**

`replay list` now supports `--path`, `--entry-path`, and `--exit-path` for parsed URL pathname matching. `--url` remains a broader replay-search text filter, while `--friction` exposes indexed error, warning, rage-click, and dead-click cohorts.

**Replay Evidence**

`replay events` lists normalized events with offsets, URL path/query fields, filters, and JSONL output. `replay summarize` returns route flow, counts, timings, friction signals, and notable events for evidence-backed analysis.

**Generated Skill**

Replay command docs and committed skill references are regenerated from command metadata and fragments; generated skill content was not hand-authored.

Refs GH-907


docs/src/fragments/commands/replay.md
```diff
@@ -36,3 +42,32 @@ sentry replay view my-org/frontend/346789a703f6454384f1de473b8b9fcc
...
+# Focus on user actions and failures on a page
+sentry replay events my-org/346789a703f6454384f1de473b8b9fcc \
+  --path /signup --kind click,network,console,error --json
+
+# Pull an evidence window around a timestamp
+sentry replay events my-org/346789a703f6454384f1de473b8b9fcc \
+  --around 01:23 --json
+
+# Emit newline-delimited JSON for large timelines
+sentry replay events my-org/346789a703f6454384f1de473b8b9fcc --json --jsonl
```
BYK: Our `--json` mode should default to JSONL in this case?

plugins/sentry-cli/skills/sentry-cli/references/replay.md
```diff
@@ -11,15 +11,64 @@ requires:
...
 Search and inspect Session Replays
 
+### `sentry replay event list <replay-id-or-url...>`
+
+List normalized events from a Session Replay
+
+**Flags:**
+- `-k, --kind <value>... - Event kind filter (navigation, click, tap, input, focus, blur, scroll, viewport, mutation, dom-snapshot, breadcrumb, network, console, error, span, web-vital, memory, video, mobile, unknown)`
+- `-u, --url <value> - Filter events by current or target URL substring`
+- `--path <value> - Filter events by parsed URL pathname`
```
BYK, line 20: I wonder if we can make this an optional positional argument instead as it looks quite common and looks like a natural sub-section?

plugins/sentry-cli/skills/sentry-cli/references/replay.md
```diff
@@ -11,15 +11,64 @@ requires:
...
+- `-q, --contains <value> - Filter events by text in labels, messages, URLs, selectors, or data`
```
BYK: `-q` is reserved for the short version of `--search` so we should either change it or change `--contains` to `--search`?

plugins/sentry-cli/skills/sentry-cli/references/replay.md
```diff
@@ -11,15 +11,64 @@ requires:
...
+- `--selector <value> - Filter events by selector substring`
```
BYK: Not clear how `--selector` and `--contains` work together or how they defer?

plugins/sentry-cli/skills/sentry-cli/references/replay.md
```diff
@@ -11,15 +11,64 @@ requires:
...
+- `--from <value> - Start offset (seconds, 90s, 01:23, or 1:02:03)`
+- `--to <value> - End offset (seconds, 90s, 01:23, or 1:02:03)`
+- `--around <value> - Center an evidence window around this offset`
+- `--before <value> - Window before --around (default: 10s)`
+- `--after <value> - Window after --around (default: 30s)`
```
BYK: Should we use the existing `-t/--period` notation for these: 

```
-t, --period <period> | Time range: "7d", "2026-04-01..2026-05-01", ">=2026-04-01" (default: "7d")
```

script/generate-docs-sections.ts
```diff
@@ -127,13 +127,20 @@ function isStandaloneCommand(route: RouteInfo): boolean {
...
+  const prefix = `sentry ${route.name} `;
+  return [
+    ...new Set(
+      route.commands.map((cmd) =>
+        cmd.path.startsWith(prefix)
+          ? cmd.path.slice(prefix.length)
+          : (cmd.path.split(" ").at(-1) ?? route.name)
+      )
+    ),
+  ];
```
BYK: Use `Array.from(new Set(....))` as that's way more efficient

src/commands/replay/index.ts
```diff
@@ -5,21 +5,28 @@
...
     view: viewCommand,
   },
+  aliases: { events: "event" },
   defaultCommand: "view",
   docs: {
     brief: "Search and inspect Session Replays",
     fullDescription:
       "Search and inspect Session Replays from your Sentry organization.\n\n" +
       "Commands:\n" +
+      "  event    Inspect normalized events from a replay (alias: events)\n" +
```
BYK: Trying to undertand the difference between `replay list` and `replay event list`. Care to elaborate?

src/commands/replay/list.ts
```diff
@@ -78,20 +87,77 @@ type ReplayListResult = {
...
+  warnings: "-count_warnings",
 };
 
 const DEFAULT_PERIOD = LIST_PERIOD_FLAG.default;
 const DEFAULT_SORT: ReplaySortValue = SORT_MAP.date;
 const PAGINATION_KEY = "replay-list";
 const COMMAND_NAME = "replay list";
+const SIMPLE_SEARCH_VALUE_RE = /^[^\s:"]+$/;
+
+function encodeReplayCursor(
```
BYK: Don't we already have generic cursor management commands? Do replays have different cursors?

src/commands/replay/list.ts
```diff
@@ -125,6 +191,219 @@ function replayUserLabel(replay: ReplayListItem): string {
   return getReplayUserLabel(replay) ?? "—";
 }
 
+function quoteSearchValue(value: string): string {
+  return SIMPLE_SEARCH_VALUE_RE.test(value) ? value : JSON.stringify(value);
```
BYK: Is this the same search syntax we use in explore? If yes we should be using existing parsers and validators for them.

BYK (changes requested):
Not a fan of the API. Looks like we spread the search syntax arguments to custom flags instead of reusing the existing helpers and UX with `-q/--search`. I'm also skeptical about the `--jsonl` flag. That shouldn't be necessary our existing `--json` output is capable of emitting JSONL by default when the output system is fed json data through the yield iterations.

PR #915, opened 2026-05-04
feat: update @sentry/api to 0.133.0 and adopt pagination improvements

MathurAditya724:
Update `@sentry/api` from 0.113.0 to 0.133.0 and adopt the new pagination type improvements. This eliminates 5 unsafe `as { cursor?: string }` casts, replaces 2 manual pagination loops with `autoPaginate`, and adds `prevCursor` support to `PaginatedResponse`.

## Changes

- **`@sentry/api` ^0.133.0** — includes pagination wrappers (`fetchPage_*`, `paginateAll_*`, `paginateUpTo_*`), widened query types (`cursor` + `per_page` now typed), and `prevCursor` in link header parsing
- **infrastructure.ts** — `PaginatedResponse<T>` gains `prevCursor?: string`; `unwrapPaginatedResult` passes it through
- **projects.ts** — manual 30-line pagination loop → `autoPaginate` one-liner; casts cleaned up
- **repositories.ts** — manual 15-line pagination loop → `autoPaginate`; casts cleaned up
- **teams.ts, releases.ts, issues.ts, events.ts** — `as { cursor?: string }` casts narrowed/removed; multi-line result casts collapsed

## What's NOT migrated (follow-up)

`traces.ts`, `discover.ts`, and `dashboards.ts` widget queries stay on raw `apiRequestToRegion` — they use Zod validation schemas and the SDK wrappers' error handling (plain `Error`) is incompatible with the CLI's `ApiError`/`AuthError` pipeline.

## Testing

- `bun test test/lib/api-client.coverage.test.ts` — 91/91 pass
- `npx tsc --noEmit` — clean (no new errors)

### DIFF
--- a/src/lib/api/projects.ts
+++ b/src/lib/api/projects.ts
@@ -27,7 +27,6 @@ import {
 } from "../db/project-cache.js";
 import { getCachedOrganizations } from "../db/regions.js";
 import { type AuthGuardSuccess, withAuthGuard } from "../errors.js";
-import { logger } from "../logger.js";
 import { getApiBaseUrl } from "../sentry-client.js";
 import { buildProjectUrl } from "../sentry-urls.js";
 import { isAllDigits } from "../utils.js";
@@ -35,8 +34,8 @@ import { isAllDigits } from "../utils.js";
 import {
   API_MAX_PER_PAGE,
   apiRequestToRegion,
+  autoPaginate,
   getOrgSdkConfig,
-  MAX_PAGINATION_PAGES,
   ORG_FANOUT_CONCURRENCY,
   type PaginatedResponse,
   unwrapPaginatedResult,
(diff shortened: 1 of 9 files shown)


src/lib/api/projects.ts
```diff
@@ -54,38 +53,23 @@ import { getUserRegions, listOrganizations } from "./organizations.js";
...
-  for (let page = 0; page < MAX_PAGINATION_PAGES; page++) {
+  const { data: allResults } = await autoPaginate(async (cursor) => {
     const result = await listAnOrganization_sProjects({
       ...config,
       path: { organization_id_or_slug: orgSlug },
-      // per_page is supported by Sentry's pagination framework at runtime
-      // but not yet in the OpenAPI spec
-      query: { cursor, per_page: API_MAX_PER_PAGE } as { cursor?: string },
+      query: { cursor, per_page: API_MAX_PER_PAGE } as {
+        cursor?: string;
```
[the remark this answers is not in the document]
MathurAditya724: Good catch — this was a real regression. The old per-callsite warning logs were dropped when switching to `autoPaginate`, and `autoPaginate` itself had no warning on the `MAX_PAGINATION_PAGES` exit path.

Fixed in a528058: added `logger.warn()` directly inside `autoPaginate` when the page-limit safety net is hit. This is better than the old approach since all callers now get the warning automatically instead of each callsite needing to remember to add one.

src/lib/api/projects.ts
```diff
@@ -54,38 +53,23 @@ import { getUserRegions, listOrganizations } from "./organizations.js";
...
-    cursor = nextCursor;
-
-    if (page === MAX_PAGINATION_PAGES - 1) {
-      logger.warn(
-        `Pagination limit reached (${MAX_PAGINATION_PAGES} pages, ${allResults.length} items). ` +
-          "Results may be incomplete for this organization."
-      );
-    }
-  }
+  }, Number.MAX_SAFE_INTEGER);
```
BYK: Shouldn't we use a sensible limit here rather than effectively going till infinity?

MathurAditya724: Good call. Replaced `Number.MAX_SAFE_INTEGER` with `MAX_PAGINATION_PAGES * API_MAX_PER_PAGE` (50 × 100 = 5,000) in both `listProjects` and `listAllRepositories`. This aligns the row-count cap with the page-count safety net already inside `autoPaginate`, so intent is explicit.

Fixed in 1a376c0.

BYK (approved):

PR #919, opened 2026-05-05
fix(install): pass through no-agent-skills flag

elucid:
Bit of a strange request--allow avoid installation of skills via the curl installer. Previously the install script didn't pass this flag though. My use case: I already have a much briefer skill for sentry cli and if these get autoinstalled my agents get different guidance.

## Summary
- Add `--no-agent-skills` to the curl installer arguments
- Forward the flag to `sentry cli setup` so users can skip agent skill installation during install
- Add an installer test that stubs downloads and asserts setup delegation args

## Test plan
- `bun test --timeout 15000 --isolate test/lib/install-script.test.ts`

### DIFF
--- a/install
+++ b/install
@@ -85,6 +85,7 @@ Options:
     -v, --version <version> Install a specific version (e.g., 0.2.0) or "nightly"
         --no-modify-path    Don't modify shell config files (.zshrc, .bashrc, etc.)
         --no-completions    Don't install shell completions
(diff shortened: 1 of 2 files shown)


BYK (approved):

BYK:
Thanks a lot @elucid. Just curious, why use your own skill? How can we improve ours?

PR #924, opened 2026-05-06
chore: update @sentry/api to ^0.141.0

MathurAditya724:
Bumps `@sentry/api` from `^0.133.0` to `^0.141.0`, picking up v0.141.0 which adds auto-generated Zod v3 schemas via a new `@sentry/api/zod` subpath export ([getsentry/sentry-api-schema#70](https://github.com/getsentry/sentry-api-schema/pull/70)).

This makes auto-generated Zod schemas available for the CLI to use instead of the hand-written ones in `src/types/`. The schemas stay in sync with the OpenAPI spec automatically on every `@sentry/api` release.

### Usage

```ts
import { zOrganization, zAlertRule } from "@sentry/api/zod";
```

### Testing

`bun run generate:schema && bun run generate:sdk && npx tsc --noEmit` — typecheck passes. `bun test test/lib/api-schema.test.ts` — 19 tests pass.

### DIFF
--- a/plugins/sentry-cli/skills/sentry-cli/references/event.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/event.md
@@ -47,19 +47,19 @@ List events for an issue
 |-------|------|-------------|
 | `id` | string | Internal event ID |
 | `event.type` | string | Event type (error, default, transaction) |
-| `groupID` | unknown |  |
+| `groupID` | string \| null | Group (issue) ID |
 | `eventID` | string | UUID-format event ID |
-| `projectID` | string |  |
-| `message` | string |  |
-| `title` | string |  |
-| `location` | unknown |  |
-| `culprit` | unknown |  |
-| `user` | unknown |  |
-| `tags` | array |  |
-| `platform` | unknown |  |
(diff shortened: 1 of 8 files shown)


plugins/sentry-cli/skills/sentry-cli/references/event.md
```diff
@@ -47,19 +47,19 @@ List events for an issue
 |-------|------|-------------|
 | `id` | string | Internal event ID |
 | `event.type` | string | Event type (error, default, transaction) |
-| `groupID` | string \| null | Group (issue) ID |
+| `groupID` | unknown |  |
```
BYK: We need to fix all these `unknown` changes in references

MathurAditya724: fixed in 73bbd09 — `zodTypeToString` didn't handle `ZodUnion` (which `@sentry/api/zod` uses for nullable fields instead of `ZodNullable`). Added union resolution + `ZodRecord` support, and re-added `.describe()` calls on all fields that lost descriptions when switching from hand-written to SDK schemas.

BYK (approved):

PR #954, opened 2026-05-11
fix(plan): only show progress from current step

### LINKED ISSUE #950: plan: stale RCA progress shown during solution polling
`issue plan` shows stale RCA progress during solution polling, making it appear as if root cause analysis is being re-run.

## What happens

When `sentry issue plan` polls for a solution after calling `triggerSolutionPlanning`, `getProgressMessage()` in `src/lib/formatters/seer.ts:69` walks backwards through **all** `state.steps[].progress[]` and returns the last message from any step — including stale messages from the earlier root cause analysis step.

This surfaces "Figuring out root cause"-style messages in the spinner even though `ensureRootCauseAnalysis()` already confirmed the RCA was `COMPLETED` and skipped re-triggering.

## Observed behavior

```
ℹ Creating plan for cause #0...
ℹ "Next.js automatic prefetching of multiple React Server Components..."
⠸ Looking at `src/mdx.ts` in `getsentry/sentry-docs`...
⠧ Looking at `app/layout.tsx` in `getsentry/sentry-docs`...
No solution found.
```

The "Looking at" messages are from the RCA step's progress, shown during solution polling.

## Expected behavior

During solution polling, progress messages should come from the solution step, not replay stale RCA step progress.

## Suggested fix

In `getProgressMessage()`, filter to the most recent **in-progress** step (or at minimum the step matching the current phase) rather than unconditionally walking backwards through all steps.

## References

- `src/lib/formatters/seer.ts:69` — `getProgressMessage()`
- `src/commands/issue/plan.ts` — plan command flow
- `src/commands/issue/utils.ts:885` — `ensureRootCauseAnalysis()`
- Observed on issue [DOCS-84Y](https://sentry.sentry.io/issues/5873366209/)

Action taken on behalf of Burak Yigit Kaya.

MathurAditya724:
`getProgressMessage()` walked backwards through all autofix steps and returned the first progress message found from any step. During solution polling after RCA completes, this surfaced stale RCA messages (e.g. "Looking at `src/mdx.ts`...") because the completed RCA step still had progress entries while the new solution step had none yet.

Now only reads progress from the last (current) step and falls back to a status-based message when it has no progress.

## Testing

`bun test test/lib/formatters/seer.test.ts` — 40 tests pass, including two new tests for the exact scenario from the issue.

Fixes #950

### DIFF
--- a/src/lib/formatters/seer.ts
+++ b/src/lib/formatters/seer.ts
@@ -63,19 +63,18 @@ export function formatProgressLine(message: string, tick: number): string {
 /**
  * Extract the latest progress message from autofix state.
  *
+ * Only returns progress from the most recent (last) step to avoid showing
+ * stale messages from earlier phases (e.g. RCA progress during solution polling).
+ *
  * @param state - Current autofix state
  * @returns Latest progress message or default
  */
 export function getProgressMessage(state: AutofixState): string {
-  if (!state.steps) {
-    return "Processing...";
-  }
-
-  // Find the most recent progress message
-  for (let i = state.steps.length - 1; i >= 0; i--) {
-    const step = state.steps[i];
-    if (step?.progress && step.progress.length > 0) {
-      const lastProgress = step.progress.at(-1);
+  if (state.steps && state.steps.length > 0) {
+    // Only look at the last step — earlier steps belong to previous phases
+    const currentStep = state.steps.at(-1);
+    if (currentStep?.progress && currentStep.progress.length > 0) {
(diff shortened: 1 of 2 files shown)


MathurAditya724:
All 18 CI checks passed (unit tests, e2e, lint, typecheck, build, CodeQL, semgrep). Self-review confirmed the fix is correctly scoped — only `getProgressMessage()` changed, no regression for the explain flow since single-step RCA still works via `steps.at(-1)`. Two new tests cover the exact bug scenario from #950. Marking ready for review.

PR #962, opened 2026-05-13
feat: migrate autofix to agent-based (explorer) endpoint

### LINKED ISSUE #961: Migrate Autofix to new agent-based endpoint
The CLI currently uses the old (non-agent) Autofix endpoint. The new agent/explorer-based version is the path forward and the team wants to deprecate the old one to consolidate engineering effort.

- The request shape is the same but response types are incompatible
- Update the CLI to consume the new Autofix endpoint's response format
- Determine a deprecation/support timeline for the old endpoint so existing CLI installations degrade gracefully

**Open questions**
- How long should the old endpoint remain supported for existing installations?
- Should older CLI versions get a clear error/upgrade prompt once the old endpoint is removed?

Raised by Hector Dearman.

Action taken on behalf of Burak Yigit Kaya.

MathurAditya724:
Switches all Seer autofix API calls to the agent-based endpoint (`?mode=explorer`), which returns blocks instead of steps. This is the path forward — the old endpoint will be deprecated.

- API layer uses `apiRequestToRegion` with `mode=explorer` instead of `@sentry/api` SDK functions
- Status values normalized from agent lowercase (`processing`, `completed`, `error`, `awaiting_user_input`) to CLI uppercase
- Root cause extraction handles both agent artifact format (`key: "root_cause"`) and legacy step format
- `triggerSolutionPlanning` simplified: posts `step: "solution"` with `run_id` instead of legacy `select_root_cause` update
- `--cause` flag removed from `issue plan` (agent determines the root cause)

## Testing

`bun test test/lib/api-client.seer.test.ts test/types/seer.test.ts test/lib/formatters/seer.test.ts test/commands/issue/utils.test.ts` — 140 tests pass.

Closes #961

### DIFF
--- a/plugins/sentry-cli/skills/sentry-cli/references/issue.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/issue.md
@@ -143,7 +143,6 @@ sentry issue plan 123456789 --cause 0
 Generate a solution plan using Seer AI
 
 **Flags:**
-- `--cause <value> - Root cause ID to plan (required if multiple causes exist)`
 - `--force - Force new plan even if one exists`
 - `-f, --fresh - Bypass cache, re-detect projects, and fetch fresh data`
 
--- a/src/commands/issue/plan.ts
+++ b/src/commands/issue/plan.ts
@@ -7,8 +7,8 @@
 
 import type { SentryContext } from "../../context.js";
 import { triggerSolutionPlanning } from "../../lib/api-client.js";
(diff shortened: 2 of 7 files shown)


MathurAditya724:
all 20 CI checks green, self-review clean — marked ready for review. the migration covers API calls, status normalization, root cause extraction from agent artifacts, and the `--cause` flag removal. tests updated to match the new response format.

PR #978, opened 2026-05-19
fix(init): fix InkUI on native Windows terminals (CLI-1NT)

betegon:
## Summary

`sentry init` was failing on Windows in CMD.exe, PowerShell, and Windows Terminal with "The interactive UI failed to load" since 0.33.0 (when InkUI replaced OpenTUI). Git Bash / MSYS2 users were fine because their POSIX layer maps `/dev/tty` — native Windows terminals don't have it.

The failure chain: `openFreshTtyForInk()` returns `null` on Windows → Ink falls back to `process.stdin` (Bun's broken fd 0, oven-sh/bun#6862) → `mountApp()` throws → `getUIAsync()` swallows it silently → `LoggingUI` is used → `LoggingUIPromptError` in `preamble()` → `WizardError`.

## Changes

**`ink-ui.ts`** — try `\\.\CON` as the TTY device path on Windows. It's the console device equivalent of `/dev/tty`: a freshly-opened handle is not fd 0 and should deliver `readable` events correctly, bypassing the Bun bug. A `setRawMode(true/false)` probe runs immediately after opening to verify the handle actually supports raw mode — if it throws, the function returns `null` and falls back gracefully rather than blowing up inside `mountApp`.

**`factory.ts`** — add a breadcrumb (with stack) in the `getUIAsync` catch block. Previously the actual InkUI throw was swallowed with no trace; only the downstream `LoggingUIPromptError` appeared in Sentry. The breadcrumb will confirm whether `\\.\CON` fixes the issue or if there's a secondary failure to address.

**`wizard-runner.ts`** — when `LoggingUIPromptError` fires but `stdin.isTTY` is true, suggest `--no-tui --yes` instead of just `--yes`. The old message was misleading for any user who expected interactive mode.

## Test Plan

- Run `sentry init` from CMD.exe or PowerShell on Windows 10/11 with Bun — should enter the wizard instead of immediately erroring
- Git Bash on Windows should be unaffected (already worked)
- After shipping, new CLI-1NT events should carry a breadcrumb showing the real InkUI error if `\\.\CON` doesn't fully resolve it on some setups

Fixes CLI-1NT

### DIFF
--- a/src/lib/init/ui/factory.ts
+++ b/src/lib/init/ui/factory.ts
@@ -29,6 +29,7 @@
  *     companions add a fraction of that and use no native code.
  */
 
+import { addBreadcrumb } from "@sentry/node-core/light";
 import { LoggingUI } from "./logging-ui.js";
 import type { WelcomeOptions, WizardUI } from "./types.js";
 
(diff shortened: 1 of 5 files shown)


BYK:
I could not repro this on Win 11 using Powershell + Windows Terminal or Xonsh + Wezterm. Both cases I used `npx sentry init`-- both ran v0.34.0

PR #988, opened 2026-05-20
fix: use Node 22 in E2E CI and simplify zstd imports

BYK:
## Summary

Node 20 is EOL. This PR:

1. **Adds `setup-node@v6` with `node-version: 22` to the E2E test job** — the library E2E tests run `node -e "..."` with the system Node, which was v20 on `ubuntu-latest`. The npm bundle now requires Node 22+ (`node:zlib` zstd APIs).

2. **Simplifies zstd imports** — removes the feature-detection dance that was added in #986 to support Node 20. Now uses direct `import { zstdCompress } from "node:zlib"` since Node 22 is the minimum.

3. **Removes stale Bun references** in comments.

### Files changed

- `.github/workflows/ci.yml` — Add `setup-node` to E2E job
- `src/lib/api/sourcemaps.ts` — Direct zstd import, clean promisify
- `src/lib/telemetry/zstd-transport.ts` — Direct zstd import, clean promisify

All 7014 tests pass.


BYK:
Superseded by follow-up PR that addresses this plus all other review comments from PRs #967-#988.

PR #1009, opened 2026-05-22
fix: handle with { type: "file" } import attributes in tsx dev mode

betegon:
Depends on #1008.

## Problem

`ink-ui.ts` imports the Ink app sidecar using a `with { type: "file" }` import attribute:

```typescript
import inkAppPath from "./ink-app.tsx" with { type: "file" };
```

This is a static module-level import, so Node.js evaluates it when `ink-ui.ts` is first loaded. Node.js only supports `with { type: "json" }` — everything else throws `ERR_IMPORT_ATTRIBUTE_UNSUPPORTED`. The factory catches the error and falls back to `LoggingUI`, but `LoggingUI` can't do interactive prompts, so the wizard fails with:

```
Error: The interactive UI failed to load. Run with --yes for non-interactive mode.
```

This affected anyone running `pnpm cli init` or `pnpm tsx /path/to/bin.ts init` in a real terminal — the interactive wizard UI was never reachable in dev mode.

**Why it worked before:** Bun supports `with { type: "file" }` natively. In esbuild builds, `text-import-plugin` transforms the attribute away before Node.js ever sees it.

## Fix

Register a synchronous Node.js module hook in `require-shim.mjs` using `registerHooks()` (available from Node 22.15, our minimum). When a module is imported with `with { type: "file" }`, the hook returns a synthetic ESM module that exports the file's absolute path as a string — the same behaviour Bun provides natively.

```js
registerHooks({
  load(url, context, nextLoad) {
    if (context.importAttributes?.type === "file") {
      return {
        format: "module",
        shortCircuit: true,
        source: `export default ${JSON.stringify(new URL(url).pathname)};`,
      };
    }
    return nextLoad(url, context);
  },
});
```

## Test plan

```
# Before
pnpm tsx /path/to/bin.ts init
→ Error: The interactive UI failed to load. Run with --yes for non-interactive mode.

# After
pnpm tsx /path/to/bin.ts init
→ Launches the Ink wizard UI
```

Note: tsx v4.20.6 itself emits a `[DEP0205]` deprecation warning for `module.register()` — that's a tsx bug unrelated to this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,7 +259,7 @@ jobs:
         run: pnpm install --frozen-lockfile
       - name: Setup codesign dependencies
         env:
-          APPLE_CERT_DATA: ${{ secrets.CSC_LINK }}
+          APPLE_CERT_DATA: ${{ secrets.APPLE_CERT_DATA }}
           APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
         run: |
           curl -L 'https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign%2F0.29.0/apple-codesign-0.29.0-x86_64-unknown-linux-musl.tar.gz' -o 'rcodesign.tar.gz'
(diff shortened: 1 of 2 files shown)


BYK:
Changes incorporated into #1008 (file import hook + createRequire fixes)

PR #1026, opened 2026-05-27
fix: preserve TTY in dev mode by inlining tsx invocation

BYK:
## Problem

`pnpm run cli init` and `pnpm run dev init` cannot run `sentry init` interactively because TTY is lost during execution.

**Root cause:** The `cli` script was `"pnpm tsx src/bin.ts"`. `pnpm tsx` is shorthand for `pnpm run tsx`, which creates a nested pnpm lifecycle spawn (outer pnpm → shell → inner pnpm → shell → tsx). The inner spawn loses TTY inheritance, so `process.stdin.isTTY` and `process.stdout.isTTY` are `undefined` in the final Node process.

This triggers three gates that block interactive mode:
1. `isNonInteractiveContext()` in `init.ts` → requires `--yes` + `--features`
2. `isInteractiveTerminal()` in `factory.ts` → forces LoggingUI instead of InkUI
3. Wizard preamble in `wizard-runner.ts` → throws WizardError

## Fix

Inline `tsx --import ./script/require-shim.mjs` directly into the `cli` and `dev` scripts, eliminating the nested `pnpm run tsx` lifecycle spawn.

With a single pnpm lifecycle layer, pnpm uses `stdio: 'inherit'`, preserving the terminal's TTY fds.

The `tsx` script alias remains for the 19+ non-interactive scripts (build, generate, check, bench) where TTY doesn't matter.

## Verification

In a real terminal:
```
pnpm run cli init --dry-run
pnpm run dev init --dry-run
```
Both should now enter interactive mode (InkUI) instead of erroring about non-interactive context.

> **Note:** Do not use `--` before subcommands (e.g. `pnpm run cli -- init`) — pnpm passes `--` through as a literal argument, which breaks subcommand parsing.


betegon:
The verification steps show `pnpm run cli -- init --dry-run` but the `--` actually gets passed through to `process.argv`, causing commander to fail with `Command not found: -- init ...`.

The working form is without `--`:
```
pnpm run cli init --dry-run
pnpm run cli init /path/to/project
```

`--` is only needed when passing flags that pnpm might intercept — but since pnpm passes it through verbatim here, it breaks subcommand parsing.

BYK:
Good catch! Updated the PR description to use `pnpm run cli init` (no `--`). Thanks!

PR #1027, opened 2026-05-27
fix(sdk): apply Stricli flag defaults in SDK invoke path

BYK:
## Problem

**CLI-1W1** (3 events) and **CLI-1W0** (1 event) crash at `serializeTimeRange(flags.period)` because `flags.period` is `undefined`.

Both issues affect users calling the CLI programmatically via `createSentrySDK()` (the typed SDK).

## Root Cause

The typed SDK's invoke path (`src/lib/sdk-invoke.ts`) calls command handlers **directly**, bypassing Stricli's `buildArgumentScanner`. This means parsed flags with defaults (e.g., `period: { kind: 'parsed', default: '7d', parse: parsePeriod }`) never have their `parse` function called on the default string when the SDK caller omits them.

The generated SDK methods (`sdk.generated.ts`) pass `period: params?.period`, which evaluates to `undefined` when omitted. All 14+ commands with a `period` flag are affected.

## Fix

Adds `applyFlagDefaults()` to the SDK invoke layer, which replicates Stricli's default application logic:

- **`kind: "parsed"` flags** with a string `default` → calls `flag.parse(flag.default)` (same as Stricli's `parseInput`)
- **Boolean/enum/counter flags** → uses the raw default value
- **Strips `undefined` values** from caller-provided flags so they don't shadow defaults
- **Skips flags** already set by the caller

`resolveCommand()` now returns both the handler function and the command's flag definitions, so `buildInvoker` can apply defaults before calling the command func.

This is a systemic fix — protects **all** flags across **all** SDK-invoked commands, not just `period`.

## Testing

- 11 new unit tests for `applyFlagDefaults()` covering: parsed flags with defaults, boolean defaults, enum defaults, optional flags without defaults, undefined stripping, caller value preservation, parse failure handling, and combined scenarios.

Fixes CLI-1W1, CLI-1W0

### DIFF
--- a/src/lib/sdk-invoke.ts
+++ b/src/lib/sdk-invoke.ts
@@ -59,17 +59,43 @@ function buildIsolatedEnv(
 /** Type alias for command handler functions loaded from Stricli's route tree. */
 type CommandHandler = (...args: unknown[]) => unknown;
 
-/** Cached command loaders, keyed by joined path (e.g., "org.list") */
-const commandCache = new Map<string, () => Promise<CommandHandler>>();
+/**
+ * Stricli flag definition shape — subset needed for default application.
+ *
+ * The SDK invoke path bypasses Stricli's `buildArgumentScanner`, so parsed
+ * flags with defaults never have their `parse` function called on the default
+ * string. We capture the flag definitions here so {@link applyFlagDefaults}
+ * can replicate that behavior at invocation time.
+ */
+export type FlagDef = {
+  kind: string;
+  default?: unknown;
+  optional?: boolean;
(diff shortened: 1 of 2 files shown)


BYK (commented):
Addressed warden's finding: `resolveFlagDefault` now **re-throws** if `flag.parse(flag.default)` fails instead of silently returning `undefined`. A parse function that rejects its own default string is a command definition bug that should surface immediately, not be silently absorbed.

Updated the test to verify the error propagates (`expect(() => applyFlagDefaults({}, flagDefs)).toThrow("parse error")`).

PR #1046, opened 2026-06-01
fix(upgrade): Resolve symlinks before self-copy guard in installBinary

sergical:
`installBinary` deletes the source file it is about to copy when the install directory is reached through a symlink, crashing the upgrade with `ENOENT`. The self-copy guard compared paths with `resolve()`, which makes paths absolute but does not follow symlinks. On macOS `/tmp` is a symlink to `/private/tmp`, so a download at `/private/tmp/.../sentry.download` (the canonicalized `process.execPath`) and the temp target `/tmp/.../sentry.download` looked different despite being the same file — the guard misfired, `unlink()` removed the source, and `copyFile()` then failed.

This surfaces during `sentry cli upgrade` when `SENTRY_INSTALL_DIR` points through a symlink; a normal `~/.local/bin` install is unaffected, which is why it was intermittent.

**Symlink-aware comparison**

The guard now compares `realpathSync`-canonicalized paths, falling back to `resolve()` when the path does not exist yet (the normal case where the temp file has not been created).

```text
Before: resolve(/private/tmp/.../sentry.download) !== resolve(/tmp/.../sentry.download)  // true -> deletes source
After:  realpath(/private/tmp/.../sentry.download) === realpath(/tmp/.../sentry.download) // true -> skips copy
```

Adds a regression test that installs through a symlinked dir and fails (ENOENT) without the fix.


MathurAditya724 (commented):
Reviewed the diff — the fix is correct and well-scoped.

**What I verified:**
- `realpathSync` resolves through symlinks (unlike `resolve()` which only makes paths absolute), so the `/tmp` → `/private/tmp` macOS case is correctly handled.
- The `try/catch` fallback to `resolve()` is necessary because `realpathSync` throws `ENOENT` on non-existent paths, which is the normal case when `tempPath` hasn't been created yet.
- The new test correctly reproduces the bug scenario: canonical `sourcePath` vs symlinked `installDir` pointing to the same real directory.
- All 49 tests in `binary.test.ts` pass. CI is green.

**One minor note (non-blocking):** The `catch` block in the `canonical` helper at `src/lib/binary.ts:458` is silent (no `log.debug`). The repo's AGENTS.md prohibits silent catch blocks in `src/`, though the existing file already has several (lines 70, 280, 285, 416, 467). Since this is a pre-existing pattern in this file and the fallback behavior is clearly documented in the comment block above, this isn't worth blocking on — but a `log.debug("realpathSync failed, falling back to resolve()", error)` would be consistent with the stated policy if you want to tighten it up.

No findings.

sergical:
Thanks for the careful review! Addressed the silent-catch note in 73277532.

One nuance worth calling out: I applied it slightly differently than the literal suggestion. Dropping `log.debug("realpathSync failed, falling back to resolve()", error)` straight into the catch would actually fire on *every* normal install — `realpathSync(tempPath)` throwing `ENOENT` is the expected path when `tempPath` hasn't been created yet, so it isn't really a failure. To avoid crying wolf on normal operation, I short-circuit that expected case with `existsSync` (no log) and reserve the `log.debug` for the catch, which now only triggers on genuinely unexpected failures (e.g. permissions):

```ts
const canonical = (p: string): string => {
  if (!existsSync(p)) {
    return resolve(p);
  }
  try {
    return realpathSync(p);
  } catch (error) {
    logger.debug("realpathSync failed, falling back to resolve()", error);
    return resolve(p);
  }
};
```

Left the pre-existing silent catches (lines 70, 280, 285, 416, 467) out of scope to keep this PR focused — happy to do a separate cleanup pass on them if that's useful.

BYK (commented):
Good catch — fixed in af2191b. ENOENT is now silently ignored (expected case: no leftover temp file), while unexpected errors still log via `logger.debug`.

PR #1048, opened 2026-06-02
fix(issue): handle GitHub-style #SHORTID in issue identifiers

BYK:
## Summary

AI agents (claude-code, codex) frequently pass GitHub-style issue references such as `org/project#SHORTID`. These previously failed with:

```
ValidationError: Invalid issue identifier: contains "#" (URL fragment)
```

because `parseIssueArg` stripped `/` but not `#` before calling `validateResourceId` (which correctly rejects `#` as a forbidden URL-fragment character).

This supersedes #1004 and fixes the underlying issue, while fixing a semantic bug in that draft and aligning with repo conventions, validation, tests, and docs.

## Approach

A dedicated `parseWithHash` helper (mirroring the existing `parseWithColon`) treats `#` as the separator before the short ID:

| Input | Result |
|---|---|
| `org/project#CLI-G` | `explicit` (org + project + suffix) |
| `org/project#123` | `explicit-org-numeric` |
| `project#CLI-G` | `project-search` |
| `#CLI-G` | bare identifier → `project-search` |
| `#123` | `numeric` |
| `#G` | `suffix-only` |

- `org/project#ID` delegates to the existing slash path, reusing the full-short-ID prefix-match logic in `parseMultiSlashIssueArg`.
- `project#ID` is treated as a project-scoped lookup (org auto-detected) — fixing the draft PR's bug where bare `project#ID` was misparsed as `org/suffix`.
- Steps 2–5 of `parseIssueArg` were extracted into `parseBareIssueIdentifier`, and `parseProjectIdentifier` is now shared between the colon and hash parsers.

### Validation

Because the `#` path short-circuits **before** the main `validateResourceId` guard, `parseWithHash` validates **both** the project prefix and the fragment itself. A `:` mixed with `#` is rejected explicitly to avoid silently swallowing the colon into a suffix (cf. the `sentry/CLI:W9` precedent).

## Behavioral change

`parseIssueArg("CLI-G#anchor")` previously threw (treated as fragment injection); it now parses as `project-search{ projectSlug: "cli-g", suffix: "ANCHOR" }`. The injection-hardening test was updated to use a fragment that still contains a forbidden character.

## Tests

- New example-based describe block covering all `#` forms and error cases (multiple `#`, empty fragment, forbidden chars, `:`-mix, invalid project prefix).
- New property tests: `org/project#FULL-SHORTID` ≡ `org/project/FULL-SHORTID`; `project#FULL-SHORTID` → `project-search`; forbidden char in fragment always throws.

## Docs

- `features.md` — new "GitHub-Style (`#` separator)" subsection under Issue ID Formats.
- `issue.md` fragment + regenerated skill reference — `#` examples.
- `issue view` `fullDescription` and the shared `<issue>` positional `brief`.
- `issue-id.ts` module JSDoc.

## Verification

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm test -- test/lib` ✅ (5464 passing)
- `pnpm run generate:docs` + `pnpm run check:fragments` ✅

Fixes CLI-1G1.


BYK:
Both bot findings (Sentry Seer and Cursor Bugbot) are valid — the `org/project#SHORTID` path delegated to `parseWithSlash` without validating the prefix for forbidden characters. Fixed in the follow-up PR #1049.

Issue #1055, opened 2026-06-03
Tracking: remaining niche legacy sentry-cli commands (bash-hook, uninstall, code-mappings, dart-symbol-map, bundle-jvm)


BYK:
Part of the parity effort tracked in #600. Tracking issue for the remaining low-priority / niche legacy `sentry-cli` commands not yet ported. Each can be split into its own issue when picked up.

## Candidates (feasible, pure TS)
- [x] `bash-hook` — output a bash script snippet for error capture. Trivial, static output. (#1083)
- [x] `uninstall` — reverse of `cli setup`: remove binary, completions, PATH entries, config dir. (#1084)
- [x] `code-mappings upload` — API call + file reading. Usually managed via UI/GitHub integration. (#1086)
- [x] `dart-symbol-map upload` — plain-text files, chunk-upload like proguard. (#1085)
- [x] `debug-files bundle-jvm` — bundle Java/Kotlin sources into a ZIP + manifest (no native parsing; the rest of `debug-files` is NOT feasible — see below). (#1089)

## Explicitly NOT porting (per #600)
- `debug-files` (upload/check/find/bundle-sources/print-sources) — requires the native `symbolic` Rust crate (Mach-O/ELF/PDB/DWARF). No JS equivalent.
- `react-native xcode/gradle` — handled by platform SDK plugins.
- `build upload/download/snapshots` — SaaS-only internal feature.
- `send-envelope` — deliberately deprecated in favor of `event send --raw`.

## Note
`release archive`/`restore`, `sourcemap resolve`, and `proguard uuid` are being implemented as quick wins (separate PR) and are NOT part of this tracking issue.

BYK:
All five candidate commands are now implemented and merged:

- `bash-hook` — #1083
- `cli uninstall` — #1084
- `dart-symbol-map upload` — #1085
- `code-mappings upload` — #1086
- `debug-files bundle-jvm` — #1089

The final command (`debug-files bundle-jvm`) landed in #1089 (merge commit a1674824a25e7e6a066f932c2f3746bb0ff70c3b), which also made the `debug-files` route visible and added symlink following to the source walker. Closing as complete.

Issue #1054, opened 2026-06-03
sourcemap: handle inline base64 sourcemaps in inject/upload


BYK:
Part of the parity effort tracked in #600.

## Goal
Handle inline base64-encoded sourcemaps in `sentry sourcemap inject` / `upload`, matching legacy `sentry-cli sourcemaps`.

## Background
The new CLI's `sourcemap inject` already follows external `//# sourceMappingURL=<path>` directives (slow-path in `src/lib/sourcemap/inject.ts`, `SOURCE_MAPPING_URL_RE`). The remaining gap is **inline** sourcemaps embedded as data URLs:

```
//# sourceMappingURL=data:application/json;base64,<...>
```

## Scope
- Detect `data:application/json;...;base64,` sourceMappingURL values.
- Base64-decode, inject the debug ID into the decoded JSON map, re-encode, and rewrite the directive in place.
- For upload, extract/handle the inline map accordingly.

## Critical gotcha (from issue #600 / legacy PR #3243)
Base64 **decode failures MUST be non-fatal** — warn and skip, never crash. Bundled terser/babel code contains template literals that produce false-positive `sourceMappingURL` lines.

## Effort
Medium (2-3 days).

Issue #1053, opened 2026-06-03
proguard: add 'proguard upload' (chunk-upload of R8/ProGuard mappings)


BYK:
Part of the parity effort tracked in #600.

## Goal
Implement `sentry proguard upload` to upload Android ProGuard/R8 mapping files, matching legacy `sentry-cli proguard upload` / `upload-proguard`.

Companion to `sentry proguard uuid` (filed/implemented separately) which computes the deterministic mapping UUID.

## Scope
- `sentry proguard upload <path>...` — upload one or more mapping files via the chunk-upload + assemble protocol.
- Each mapping is bundled with name `/proguard/<uuid>.txt` where `<uuid>` is the content-checksum UUID.
- Flags from legacy CLI to consider: `--uuid` (force UUID), `--no-upload`, `--require-one`, `--no-reprocessing`, project/org targeting.

## Technical notes
- The UUID is `uuidv5(uuidv5(NAMESPACE_DNS, "guardsquare.com"), <raw file bytes>)` — verified byte-for-byte against the legacy CLI and the `rust-proguard` crate. Reuse the helper from `src/lib/proguard.ts` (added with `proguard uuid`).
- Reuse the artifact-bundle + chunk-upload infrastructure in `src/lib/api/sourcemaps.ts` (`getChunkUploadOptions`, `buildArtifactBundle`, `hashChunks`, assemble flow). The assemble endpoint differs for proguard — confirm the correct endpoint.
- Pure TS, no native deps (mapping files are plain text).

## Effort
Medium (2-3 days).

Issue #1052, opened 2026-06-03
monitor: add cron monitor check-ins (monitor run / list)


BYK:
Part of the parity effort tracked in #600.

## Goal
Implement legacy `sentry-cli monitors` cron check-in support. This is the biggest remaining net-new gap from the old Rust CLI. Most other CI/CD parity items (releases, deploys, issue mutations, sourcemap inject/upload) are already done.

## Scope
- `sentry monitor run <slug> -- <command>` — wrap an arbitrary command, send a check-in envelope on start (`in_progress`), then `ok`/`error` on completion based on exit code. Forward stdin/stdout/stderr and signals (SIGINT/SIGTERM); preserve the child exit code.
- `sentry monitor list` — list configured monitors (API-token path).
- (Optional) `--schedule` to auto-create the monitor if it doesn't exist.

## Technical notes
- Check-ins are sent via **DSN** (not auth token) as an envelope with a `check_in` item type. Reuse the existing envelope infra in `src/lib/envelope/` (`buildEnvelopeUrl`, `sendEnvelopeRequest`, `@sentry/core` serialization).
- Monitor CRUD (`list`) uses `apiRequestToRegion` from `src/lib/api/infrastructure.ts`.
- New module `src/lib/api/monitors.ts` + check-in/monitor types in `src/types/sentry.ts`.
- New `src/commands/monitor/` route group wired into `src/app.ts`.
- Process management via `node:child_process` (project migrated off Bun APIs).

## Effort
Large (~1 week). Signal handling and exit-code preservation need careful cross-platform testing.

andrianvaleanu:
This parity gap is a realpain. Getting cron signal handling right is tedious work, but https://pulsetic.com/ helps too.

PR #1058, opened 2026-06-03
feat: add release archive/restore, sourcemap resolve, proguard uuid

BYK:
## Summary

Parity quick wins toward replacing the legacy Rust `sentry-cli` (ref #600). Adds three commands whose infrastructure already existed:

- **`release archive` / `release restore`** — toggle a release's lifecycle status between `archived` and `open` via `updateRelease({ status })`. Archived releases are hidden from the default `release list` but retained; restore brings them back. Both support `--dry-run`.
- **`sourcemap resolve`** — read-only diagnostic that reports, per JS file, how its sourcemap resolves (convention `<name>.map` / external `sourceMappingURL` / inline `data:` URL / remote / none) and whether a Sentry debug ID has been injected. Reuses the `inject` discovery logic and **never mutates files**. Helps debug why `sourcemap upload` may not find expected maps.
- **`proguard uuid`** — computes the deterministic ProGuard/R8 mapping UUID. Pure TS via `node:crypto`, no native deps.

## ProGuard UUID — verified against the legacy CLI

The UUID is a content checksum matching `rust-proguard`'s `ProguardMapping::uuid()`:

```
NAMESPACE = uuidv5(NAMESPACE_DNS, "guardsquare.com")  // 4f44f30f-24be-53d0-bab6-f47c7120ad6c
uuid      = uuidv5(NAMESPACE, <raw file bytes>)
```

Reproduced **byte-for-byte** against the legacy `sentry-cli` integration fixtures (used as golden tests):
- `void\n` (5 bytes) → `5db7294d-87fc-5726-a5c0-4a90679657a5`
- sample `mapping.txt` (155 bytes) → `c038584d-c366-570c-ad1e-034fa0d194d7`

The variant-nibble encoding uses a string-based approach (no bitwise ops, to satisfy `noBitwiseOperators`); I verified it is mathematically equivalent to the canonical `byte[8] & 0x3f | 0x80` bitmask for all 16 hex digits.

## Changes
- `src/lib/api/releases.ts`: add `status?` to `updateRelease` body type.
- `src/commands/release/{archive,restore}.ts` + route wiring.
- `src/commands/sourcemap/resolve.ts` + `src/lib/sourcemap/inject.ts` (`resolveDirectorySourcemaps`) + route wiring.
- `src/commands/proguard/{uuid,index}.ts` + `src/lib/proguard.ts` + `src/app.ts` wiring.
- `src/lib/complete.ts`: register the new org-positional `release archive`/`restore` verbs.
- Docs fragments (`release`, `sourcemap`, new `proguard`) + regenerated skill references.

## Tests
40 new tests (all green); full suite: **7440 passed / 0 failed / 13 skipped**.
- `proguard.property.test.ts`: golden vectors + properties (well-formed UUIDv5, determinism, content-sensitivity).
- `sourcemap/resolve.test.ts`: companion-map detection, inline/remote/missing classification, debug-ID extraction, sort order.
- `release/archive.test.ts`: archive/restore status calls, `--dry-run`, error paths.

## Out of scope (tracked separately)
- #1052 cron `monitor` check-ins · #1053 `proguard upload` · #1054 inline base64 sourcemaps · #1055 niche commands.

`send-envelope` is intentionally **not** revived (deprecated in favor of `event send --raw`).

### DIFF
--- a/src/commands/release/restore.ts
+++ b/src/commands/release/restore.ts
@@ -1,0 +1,120 @@
+/**
+ * sentry release restore
+ *
+ * Restore an archived release by setting its status back to "open", making it
+ * visible in the default release list again.
+ */
+
+import type { SentryContext } from "../../context.js";
+import { updateRelease } from "../../lib/api-client.js";
+import { buildCommand } from "../../lib/command.js";
+import { ContextError } from "../../lib/errors.js";
+import {
+  colorTag,
+  escapeMarkdownInline,
+  mdKvTable,
+  renderMarkdown,
+  safeCodeSpan,
+} from "../../lib/formatters/markdown.js";
+import { CommandOutput } from "../../lib/formatters/output.js";
+import { DRY_RUN_ALIASES, DRY_RUN_FLAG } from "../../lib/mutate-command.js";
+import { resolveOrg } from "../../lib/resolve-target.js";
+import type { SentryRelease } from "../../types/index.js";
+import { parseReleaseArg } from "./parse.js";
+
+/**
+ * Render the human-readable result of restoring a release.
+ *
+ * @param data - Either the dry-run preview object or the updated release.
+ */
+function formatReleaseRestored(data: Record<string, unknown>): string {
+  if (data.dryRun) {
+    return renderMarkdown(
+      `Would restore release ${safeCodeSpan(String(data.version))} (dry run)`
+    );
+  }
+  const release = data as unknown as SentryRelease;
+  const lines: string[] = [];
+  lines.push(`## Release Restored: ${escapeMarkdownInline(release.version)}`);
+  lines.push("");
+  const kvRows: [string, string][] = [];
+  kvRows.push(["Version", safeCodeSpan(release.version)]);
+  kvRows.push(["Status", colorTag("green", "Open")]);
+  lines.push(mdKvTable(kvRows));
+  return renderMarkdown(lines.join("\n"));
+}
+
+export const restoreCommand = buildCommand({
+  docs: {
+    brief: "Restore an archived release",
+    fullDescription:
+      "Restore an archived release by setting its status back to open, " +
+      "making it visible in the default `sentry release list` again.\n\n" +
+      "Examples:\n" +
+      "  sentry release restore 1.0.0\n" +
+      "  sentry release restore my-org/1.0.0\n" +
+      "  sentry release restore 1.0.0 --dry-run",
+  },
+  output: {
+    human: formatReleaseRestored,
+  },
+  parameters: {
+    positional: {
+      kind: "array",
+      parameter: {
+        placeholder: "org/version",
+        brief: "[<org>/]<version> - Release version to restore",
+        parse: String,
+      },
+    },
+    flags: {
+      "dry-run": DRY_RUN_FLAG,
+    },
+    aliases: { ...DRY_RUN_ALIASES },
+  },
+  async *func(
+    this: SentryContext,
+    flags: {
+      readonly "dry-run": boolean;
+      readonly json: boolean;
+      readonly fields?: string[];
+    },
+    ...args: string[]
+  ) {
+    const { cwd } = this;
+
+    const joined = args.join(" ").trim();
+    if (!joined) {
+      throw new ContextError(
+        "Release version",
+        "sentry release restore [<org>/]<version>",
+        []
+      );
+    }
+
+    const { version, orgSlug } = parseReleaseArg(
+      joined,
+      "sentry release restore [<org>/]<version>"
+    );
+    const resolved = await resolveOrg({ org: orgSlug, cwd });
+    if (!resolved) {
+      throw new ContextError(
+        "Organization",
+        "sentry release restore [<org>/]<version>"
+      );
+    }
+    if (flags["dry-run"]) {
+      yield new CommandOutput({ dryRun: true, version, org: resolved.org });
+      return { hint: "Dry run — release was not restored." };
+    }
+
+    const release = await updateRelease(resolved.org, version, {
+      status: "open",
+    });
+    yield new CommandOutput(release);
+    const hint = resolved.detectedFrom
+      ? `Detected from ${resolved.detectedFrom}`
+      : undefined;
+    return { hint };
+  },
+});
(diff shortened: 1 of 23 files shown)


src/commands/sourcemap/resolve.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    const absDir = resolvePath(dir);
+    const files = resolutions.map((r) => ({
+      ...r,
+      relPath: relative(absDir, r.jsPath) || r.jsPath,
+      relMapPath: r.mapPath ? relative(absDir, r.mapPath) : undefined,
+    }));
+
+    const resolved = files.filter((f) => f.mapPath).length;
+    const withDebugId = files.filter((f) => f.debugId).length;
```
BYK, line 182: Instead of doing two `files.filter()` calls, you can do a single for-of loop and increment the counters there to save a full loop.

src/lib/sourcemap/inject.ts
```diff
@@ -403,6 +403,108 @@ export async function diagnoseEmptyDiscovery(
...
+      if (ignoreMatcher.ignores(rel)) {
+        continue;
+      }
+    }
+
+    const sourceMappingUrl = await extractSourceMappingUrl(jsPath);
+    const inline = sourceMappingUrl?.startsWith("data:") ?? false;
+    const remote =
+      (sourceMappingUrl?.startsWith("http://") ?? false) ||
+      (sourceMappingUrl?.startsWith("https://") ?? false);
```
BYK, line 483: Wouldn't it be nicer to use `sourceMappingUrl && /https?:\/\//i.test(sourceMappingUrl)` instead?

src/lib/sourcemap/inject.ts
```diff
@@ -403,6 +403,108 @@ export async function diagnoseEmptyDiscovery(
...
+      jsPath,
+      mapPath,
+      sourceMappingUrl,
+      inline,
+      remote,
+      debugId,
+    });
+  }
+
+  results.sort((a, b) => a.jsPath.localeCompare(b.jsPath));
```
BYK, line 504: Do we really need `localeCompare` for paths? (serious question, not an implicit suggestion)

src/lib/proguard.ts
```diff
@@ -0,0 +1,85 @@
...
+  );
+}
+
+/**
+ * The ProGuard namespace UUID, derived as `uuidv5(NAMESPACE_DNS,
+ * "guardsquare.com")`. Equals `4f44f30f-24be-53d0-bab6-f47c7120ad6c`.
+ */
+export const PROGUARD_NAMESPACE = uuidV5(
+  Buffer.from("guardsquare.com", "utf-8"),
+  NAMESPACE_DNS
```
BYK, line 69: Since this value never changes, why not pre-compute the `Buffer.from(namespace.replaceAll(....)` stuff in `uuidV5` anyway?

src/lib/proguard.ts
```diff
@@ -0,0 +1,85 @@
...
+ * The UUID is a deterministic content checksum: `uuidv5(PROGUARD_NAMESPACE,
+ * <raw file bytes>)`. Identical file contents always yield the same UUID,
+ * matching the value assigned by `sentry proguard upload` and the legacy
+ * `sentry-cli proguard uuid`.
+ *
+ * @param content - Raw bytes of the mapping file
+ * @returns Lowercase hyphenated UUID string
+ */
+export function computeProguardUuid(content: Buffer): string {
+  return uuidV5(content, PROGUARD_NAMESPACE);
```
BYK, line 84: Same as above

src/commands/sourcemap/resolve.ts
```diff
@@ -0,0 +1,201 @@
...
+    );
+
+    const absDir = resolvePath(dir);
+    const files = resolutions.map((r) => ({
+      ...r,
+      relPath: relative(absDir, r.jsPath) || r.jsPath,
+      relMapPath: r.mapPath ? relative(absDir, r.mapPath) : undefined,
+    }));
+
+    const resolved = files.filter((f) => f.mapPath).length;
```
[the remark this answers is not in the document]
BYK: Fixed in #1067. The summary count was already corrected in #1063 via `hasSourcemap()` (includes inline + remote). The inject-hint sub-point is intentional: `sentry sourcemap inject` writes the debug ID into the JS file itself regardless of where its sourcemap lives, so the hint is appropriate for any file missing a debug ID.

src/commands/release/restore.ts
```diff
@@ -0,0 +1,120 @@
...
+  },
+  parameters: {
+    positional: {
+      kind: "array",
+      parameter: {
+        placeholder: "org/version",
+        brief: "[<org>/]<version> - Release version to restore",
+        parse: String,
+      },
+    },
```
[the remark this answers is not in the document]
BYK: Fixed in #1067. Rather than a one-off guard, all release commands now use a single `kind: "tuple"` positional so Stricli enforces arity (extra positionals are rejected). A new GritQL lint rule (`no-args-join-in-release.grit`) prevents the `args.join()` pattern from being reintroduced.

src/commands/sourcemap/resolve.ts
BYK: Done in #1067 — replaced the two `files.filter()` passes with a single `for...of` loop tallying both counters.

src/lib/sourcemap/inject.ts
BYK: Done in #1067 — now uses an anchored `/^https?:\/\//i` regex (`REMOTE_SOURCE_MAPPING_URL_RE`) instead of the two `startsWith` checks.

src/lib/sourcemap/inject.ts
BYK: Good call — dropped `localeCompare` in #1067 for a byte-wise comparison, matching the plain `.sort()` already used for paths in `upload.ts`. Locale-aware comparison isn't needed for filesystem paths.

src/lib/proguard.ts
BYK: Done in #1067 — `uuidV5()` now takes precomputed namespace bytes; `NAMESPACE_DNS_BYTES` and `PROGUARD_NAMESPACE_BYTES` are computed once at module load instead of per call. Golden vectors unchanged.

src/lib/proguard.ts
BYK: Same fix in #1067 — `PROGUARD_NAMESPACE_BYTES` is precomputed once and passed to `uuidV5()`.

PR #1098, opened 2026-06-12
fix(docs): Deploy markdown pages to production

dcramer:
Generated Starlight markdown routes are now published to the GitHub Pages production root on main pushes, while PR previews continue to live under _preview. The docs workflow preserves the preview build separately, rebuilds production with root URLs, applies the sourcemap inject/upload/remove flow to both artifacts, and replaces only the production root so preview directories remain intact.

This also makes the Astro prerender workaround explicit by declaring nanoid in the docs package and documenting why the nanoid/non-secure CJS alias exists. I verified the generated docs, production and preview builds, artifact URL bases, sourcemap stripping, workflow YAML, and patch checks.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -30,6 +30,9 @@ concurrency:
 jobs:
   preview:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


dcramer:
ignore this. some fixes in here but idk that we need the larger change

PR #1119, opened 2026-06-22
feat(init): launch browser login when unauthenticated in a TTY

betegon:
## Summary

Running `sentry init` while logged out currently fails fast with "No authentication token found. Run `sentry auth login`". Every other command already auto-launches the browser device-flow login in an interactive terminal (via `autoAuthMiddleware`); this brings `sentry init` to parity by reusing the same `runInteractiveLogin` flow — no new login code.

## Changes

In `src/commands/init.ts`, before any wizard UI/timer setup: if the run is interactive (not `--yes` / `--dry-run` / non-TTY) and there's no auth token, call `runInteractiveLogin()`. On success the wizard proceeds; on cancel it exits 1 (matching `auth login`).

Why here rather than `autoAuthMiddleware`: the wizard's `finally` arms a macOS force-exit timer and the Ink UI takes over the terminal once mounted — either would interfere with an `AuthError`-triggered login mid-poll. Running it early in `init.ts` (before `resolveTarget`'s API calls, before the timer, before any UI) avoids both.

For genuinely non-interactive runs (`--yes`, piped, CI), `checkReadiness` now prints a clearer message — how to export `SENTRY_AUTH_TOKEN` and where to create a token.

## Test Plan

- `test/commands/init.test.ts` — new "interactive auth gate" cases: unauthenticated interactive run launches login then proceeds; cancelled login bails with exit 1 and never runs the wizard; authenticated run skips login.
- `biome check` clean; `vitest run test/commands/init.test.ts` → 48 passed.

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -393,6 +393,32 @@ export const initCommand = buildCommand<
       throw err;
     }
 
+    // 3b. Launch the browser login flow when unauthenticated in an
+    //     interactive terminal, matching the auto-login every other command
+    //     gets from `autoAuthMiddleware`. The wizard can't lean on that
+    //     middleware: `runWizard`'s `finally` arms a macOS force-exit timer
+    //     (step 7) that would kill an `AuthError`-triggered login mid-poll,
+    //     and the Ink UI owns the terminal once it mounts. Running login here
+    //     — before `resolveTarget`'s API calls, before the timer is armed, and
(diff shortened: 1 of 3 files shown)


betegon:
Wrong target — this belongs in `autoAuthMiddleware` (every command), not `init`. The existing auto-login already runs for all commands but is gated to TTY; the actual change is to drop that gate so it also fires in non-TTY. Redoing as a minimal middleware change.

PR #1134, opened 2026-06-23
fix sql injection

betegon:
## Summary

Removes a SQL injection vulnerability in the local SQLite layer by replacing string-concatenated SQL with parameterized queries.

The previous pattern interpolated user input straight into the statement text:

```ts
db.prepare(`SELECT * FROM users WHERE name LIKE '%${term}%'`).all();
```

An input like `'; DROP TABLE users; --` could alter the query structure. All dynamic values are now bound as `?` parameters and identifiers are validated against a strict allow-list.

## Changes (5 files)

- `src/lib/db/secure-query/query-builder.ts` — fluent builder emitting bound `?` placeholders; validates table/column identifiers.
- `src/lib/db/secure-query/user-repository.ts` — data-access layer using bound params for lookup, search, and insert.
- `src/lib/db/secure-query/sanitize.ts` — `escapeLike`, `stripControlChars`, `safeIdentifier` for values that cannot be bound.
- `src/lib/db/secure-query/README.md` — documents the vulnerability and the parameterization rules.
- `test/lib/db/secure-query.test.ts` — regression tests asserting injection payloads are treated as data, not SQL.

> Note: opened as a test fixture for the PR risk analyzer.

### DIFF
--- a/src/lib/db/secure-query/README.md
+++ b/src/lib/db/secure-query/README.md
@@ -1,0 +1,39 @@
+# secure-query
+
+Parameterized query helpers for the local SQLite layer.
+
+## Why this exists
+
+Several historical call sites built SQL by concatenating user input directly
+into the statement text, e.g.:
+
+```ts
+db.prepare(`SELECT * FROM users WHERE name LIKE '%${term}%'`).all();
+```
+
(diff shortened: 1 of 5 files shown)


betegon:
this was a test. closing it.

PR #1143, opened 2026-06-26
feat(init): run the setup agent locally via the Claude Agent SDK

betegon:
## Summary

Replaces the remote Mastra workflow (suspend/resume over the network) with a
local coding agent powered by `@anthropic-ai/claude-agent-sdk`. The agent
inspects the project, fetches Sentry docs on demand, installs the SDK, and
applies changes locally. We keep all the pre-agent work (preflight, org/project
resolution, project creation, feature selection, UI) and drop the suspend/resume
protocol and `@mastra/client-js`.

## Changes

- New local agent runner (`src/lib/init/agent/`): drives `query()`, gates tools via `canUseTool` (.env block + bash allowlist), and isolates the spawned agent from the user's own Claude Code setup (scratch `CLAUDE_CONFIG_DIR`, `CLAUDE_CODE_DISABLE_AUTO_MEMORY=1`, and `settingSources: []`) so the user's `CLAUDE.md`/settings can't leak in as a prompt-injection vector.
- Model traffic routes through the Sentry init gateway -> Vercel AI Gateway (`ANTHROPIC_BASE_URL`). `SENTRY_INIT_ANTHROPIC_API_KEY` is a BYO-key/self-host/dev escape hatch straight to Anthropic.
- Docs are local and iterative: `get_docs_by_keywords` walks `docs.sentry.io/doctree.json` and fetches `.md` pages — the agent calls it as often as it needs (`src/lib/init/docs/`). No remote docs service.
- Deterministic Xcode/pbxproj transforms (sentry-cocoa SPM, React Native build phases) ship as in-process tools the agent invokes when it detects the platform (`src/lib/init/agent/framework/`).
- Containment is layered: the always-on `canUseTool` gate + `safePath` are the primary boundary; the SDK OS sandbox (filesystem/network allowlists) is defense-in-depth and degrades gracefully (`failIfUnavailable: false`) on hosts without sandbox support.
- `wizard-runner` rewritten to the local flow; removed `init-service-auth` and the old suspend/resume test.

## Distribution / size

The SDK's JS is bundled at build time, but its per-platform native runtime is
not — the CLI stays fully bundled with zero runtime deps (`check:no-deps`). On
first `init` the native runtime is downloaded and cached under
`~/.sentry/agent/<version>/<platform>/` (integrity-checked) and reused; running
from source uses the SDK's own binary and skips the download. So the shipped
artifacts barely grow; the heavy part is a one-time, per-machine, cached fetch.

Measured fresh on darwin-arm64 (other platforms similar):

| Artifact | Before (main) | After | Δ |
| --- | --- | --- | --- |
| Single binary (SEA) | 101.3 MB | 101.6 MB | +~0.3 MB |
| npm package, packed | 2.39 MB | 2.50 MB | +~0.1 MB |
| npm package, unpacked | 8.9 MB | 9.2 MB | +~0.3 MB |
| Bundle `dist/index.cjs` | 4060 KB | 4384 KB | +~324 KB |
| Agent runtime (native `claude`) | — | ~62 MB download / ~210 MB on disk | not shipped; fetched once on first `init`, cached in `~/.sentry` |

(For comparison, embedding the native runtime into the binary instead would take it to ~312 MB per platform, ~3x — which is why we download-and-cache.)

## Test plan

- `pnpm typecheck`, `pnpm lint`, `pnpm check:deps`, and the unit suite (`vitest run test/lib/init`, including the new agent/docs tests) pass.
- Ran `sentry init` on 19 framework test projects (JS, Python, Cloudflare, native iOS, monorepos, large apps): 19/19 applied a working integration.
- Runtime-verified data lands in Sentry: node-express (errors + traces) and flask (errors).
- Verified from the **compiled binary** run outside any node_modules: first run downloads + caches the runtime (`~/.sentry/agent/.../claude`), subsequent runs reuse it.
- Compared against production (0.37.0, Mastra) across the same 19 projects: produced a working integration on every project where production did, and also completed several that the current production flow did not (e.g. monorepos it declines, a run that timed out). Runs felt faster in practice, though this wasn't formally benchmarked.

## Known gaps (follow-ups)

- Monorepo app-selection isn't gated: the agent auto-picks an app instead of requiring `--app` like prod. Usually it picks well, but for strapi it chose the framework's own package. Needs the deferred app-listing + `--app` gating.
- Package-manager detection is non-deterministic (one nextjs run used npm in a bun project). Worth pinning.

Depends on getsentry/cli-init-api#182 (the gateway) being deployed. Merge/deploy the gateway first.


betegon:
Triaged the latest review findings:

- **`.envrc` not blocked** and **Grep glob/include `.env` bypass** (permissions.ts) — fixed in dad734c4f: the env-file guard now also blocks `.envrc` (direnv) and is glob-aware, so `**/.env*` / `*.env` patterns are denied too. Added tests.
- **Missing integrity skips runtime verification** (runtime.ts) — fixed in dad734c4f: we now refuse to execute the downloaded agent runtime unless it verifies against the registry's sha512 `integrity` (falling back to the sha1 `shasum`), instead of silently skipping when integrity was absent.
- **Sentry token in the agent subprocess env** (reachable by install lifecycle scripts) — accepted with mitigation. The token must live in the subprocess env for the SDK to authenticate model calls to the gateway; there is no out-of-band channel (this is the same pattern Claude Code and PostHog's wizard use). Exfiltration is constrained by the OS sandbox's `network.allowedDomains` egress allowlist, and the subprocess is isolated from the user's `~/.claude` via `CLAUDE_CONFIG_DIR` + `CLAUDE_CODE_DISABLE_AUTO_MEMORY`. Residual risk acknowledged.

Resolving these threads.

PR #1144, opened 2026-06-26
fix(agent-skills): write skill files atomically to prevent partial reads

BYK:
## Problem

`installAgentSkills` writes each skill file with an in-place `writeFile`,
which **truncates the destination before rewriting it**. AI coding agents
(Claude Code, OpenCode) discover skills by globbing the skills tree for
`SKILL.md` and parsing the YAML frontmatter. If an agent's skill scan lands
inside that truncate→rewrite window, it reads empty/partial frontmatter and
**silently drops the skill**.

Because the generated `SKILL.md` carries a build-timestamp `version`, its
content changes on every dev build, so the file is rewritten frequently —
making the race realistically reachable during normal install churn.

> Note: this hardens the *write* side. The most common reason a freshly
> installed skill doesn't appear is that OpenCode memoizes its skill scan at
> session startup and only re-scans on restart — that's a separate,
> agent-side concern and is not addressed here.

## Fix

Add `atomicWriteFile(destPath, content)` and use it in `writeSkillFiles`:

- Write to a uniquely-named temp file `.<name>.<pid>.<rand>.tmp` in the
  **same directory** as the destination, then `rename()` into place.
- Same directory ⇒ same filesystem ⇒ the rename is atomic on POSIX, so a
  concurrent reader always sees the complete old or complete new file.
- The temp name is dot-prefixed and `.tmp`-suffixed so it never matches the
  `SKILL.md` glob agents look for, even while it briefly exists.
- On failure the temp file is removed best-effort (cleanup errors reported at
  `debug` level, no silent catch) and the original error is re-thrown.

## Tests

- New test asserts no `.tmp` leftovers remain in the skill dir or
  `references/` after install (guards the rename/cleanup path).
- `agent-skills.test.ts`: 13/13 pass · `cli/setup.test.ts`: 30/30 pass ·
  typecheck + biome clean.

### DIFF
--- a/src/lib/agent-skills.ts
+++ b/src/lib/agent-skills.ts
@@ -61,13 +61,17 @@ export function getSkillInstallPath(
  * skills by globbing the skills tree for `SKILL.md` files and parsing their
  * frontmatter. A plain in-place `writeFile` truncates the file before
  * rewriting it; an agent scanning during that window reads empty frontmatter
- * and silently drops the skill. The rename eliminates that race.
+ * and silently drops the skill. On POSIX the rename eliminates that race.
+ * (On Windows the replace still works, but can fail with a sharing violation
+ * if a reader holds the destination open — a transient, best-effort outcome.)
  *
- * The temp file is named `.SKILL.md.<pid>.<rand>.tmp` so it never matches the
- * `SKILL.md` glob agents look for, even while it exists.
+ * The temp file is named `.<basename>.<pid>.<rand>.tmp` (e.g.
+ * `.SKILL.md.1234.ab12.tmp`) so it never matches the `SKILL.md`/`*.md` globs
+ * agents look for, even while it exists.
  *
- * On failure the temp file is removed (best-effort) and the original error is
- * re-thrown for the caller to handle.
+ * Precondition: `dirname(destPath)` must already exist; callers create the
+ * directory before invoking this helper. On failure the temp file is removed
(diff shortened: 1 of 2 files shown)


BYK:
### CI note: E2E `skill-eval` failure is an Anthropic API outage, not a regression

The red **E2E Tests** job is the two `test/e2e/skill-eval.test.ts` LLM-eval cases (`claude-sonnet-4-6`/`claude-opus-4-6 meets threshold`) failing with:

```
[planner] API error: Invalid response body while trying to fetch https://api.anthropic.com/v1/messages: Premature close
AssertionError: expected 0 to be greater than or equal to 0.75
```

These call `api.anthropic.com` directly; the `Premature close` → score `0` is an upstream API outage. All **126** non-LLM E2E tests pass, and this PR only touches the skill-file *write* mechanism (`agent-skills.ts`), which cannot affect eval scores. Re-running once the API recovers should clear it.

### Pushed: test hardening (commit `3393368`)
An adversarial self-review found the original atomic-write test was tautological — it passed even when `atomicWriteFile` was reverted to a plain `writeFile` (verified by mutation). Replaced it with real guards:
- asserts every `writeFile` targets a `.tmp` staging path and the dest is published via `rename` (fails under a non-atomic mutation);
- a rollback test covering the previously-untested catch/cleanup branch (rename fails → returns `null`, temp cleaned up, no partial `SKILL.md`).

PR #1160, opened 2026-07-01
feat(init): speak the Vercel Workflows protocol

MathurAditya724:
Updates the init CLI to talk to the migrated init server, which moved from the Mastra workflow engine to **Vercel Workflows**. That migration (getsentry/cli-init-api#183) replaced the Mastra client contract (`/api/workflows/sentry-wizard/{create-run,start-async,resume-async}`) with a new one: `POST /api/wizard`, `GET /api/run`, `POST /api/resume`. Pointing the CLI at the new server currently 404s because those endpoints don't match.

## Changes
- Add `src/lib/init/workflow-client.ts` — a shim presenting the same `createRun/startAsync/resumeAsync/runById` surface the wizard runner already uses, while speaking the new protocol underneath. It maps the server's `{ status, seq, request }` into the existing `WorkflowRunResult` shape and polls `GET /api/run` between suspends.
- Resume targets the deterministic hook token `wizard:<runId>:<seq>`, reconstructed from the runId (returned by `/api/wizard`) and the `seq` reported by `/api/run` — the token is never received from the server, so no capability leaks.
- Replace `MastraClient` in `wizard-runner.ts` with the shim; drop the now-unused `@mastra/client-js` dependency.
- Update `wizard-runner.test.ts` to mock `createWorkflowClient` instead of `MastraClient.prototype.getWorkflow`.

## Testing
`vitest run test/lib/init` — 418 passing. Typecheck + biome clean on changed files.

## Notes / follow-ups
- `DEFAULT_MASTRA_API_URL` still points at the old Cloudflare worker; it should be updated to the new Vercel URL once the server is deployed. Local testing works today via `MASTRA_API_URL=http://localhost:4111`.
- Server-side counterpart: getsentry/cli-init-api#183.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "@biomejs/biome": "2.3.8",
     "@clack/prompts": "0.11.0",
     "@hono/node-server": "^2.0.6",
-    "@mastra/client-js": "^1.26.0",
     "@sentry/api": "^0.180.0",
     "@sentry/core": "10.50.0",
     "@sentry/node-core": "10.50.0",
--- a/plugins/sentry-cli/skills/sentry-cli/references/dashboard.md
+++ b/plugins/sentry-cli/skills/sentry-cli/references/dashboard.md
@@ -42,7 +42,7 @@ View a dashboard
 - `-w, --web - Open in browser`
 - `-f, --fresh - Bypass cache, re-detect projects, and fetch fresh data`
 - `-r, --refresh <value> - Auto-refresh interval in seconds (default: 60, min: 10)`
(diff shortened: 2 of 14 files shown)


MathurAditya724:
good catch — `runById` was indeed writing the recovered seq into its own local `seqRef` that got discarded, so the next `resumeAsync` would build a stale token. fixed in b306e85: `toWorkflowRunResult` now carries `_seq` on the result, `WorkflowRun` exposes `updateSeq()`, and both recovery paths in `resumeWithRecovery` call `syncSeqFromRecovery()` before returning.

PR #1164, opened 2026-07-01
feat(debug-files): add --il2cpp-mapping for Unity IL2CPP line mappings

BYK:
> **Stacked on #1163** (embedded Portable PDB extraction). Review that first; GitHub will retarget this PR's base to `main` automatically once #1163 merges. The diff shown here is only this PR's changes.

## What

Adds `sentry debug-files upload --il2cpp-mapping`: for each prepared debug file it computes a **Unity IL2CPP C++→C# line mapping** (from the file's primary debug-info object) and uploads it as a separate `il2cpp` DIF carrying the object's debug id — faithful to legacy `sentry-cli`'s `create_il2cpp_mappings`.

Consumes the new WASM API from **`@sentry/symbolic@13.7.0`** (getsentry/symbolic#1005):
- top-level `il2cppLineMapping(object, provider)`
- `SourceBundleWriter.collectIl2cppSources` (present since 13.5.0)

## Behavior (matches legacy)

- The generated C++ source files an object references are read from disk via a provider; IL2CPP `//<source_info:File.cs:line>` markers are parsed from them. Objects that yield an empty mapping are skipped.
- The mapping DIF is named `<name>.il2cpp` and carries the source file's debug id.
- When combined with `--include-sources`, the per-file source bundle **also** collects the referenced C# source files (`collectIl2cppSources`) — otherwise the source bundle is unchanged.
- Missing/unreadable sources and mapping failures are swallowed (debug-logged), never aborting the upload.
- `--symbol-maps` (BCSymbolMap resolution) remains the only deferred legacy upload option.

## Implementation

- `src/lib/dif/index.ts`: `createIl2cppLineMapping(data, readSource)` + a `collectIl2cppSources` option on `createSourceBundle()`.
- `src/commands/debug-files/upload.ts`: `--il2cpp-mapping` flag; extracted a shared `readSourceFile` disk-reader used by both the source bundle and the IL2CPP provider.

## Tests

- `test/lib/dif/il2cpp.test.ts` — `createIl2cppLineMapping` (mapping JSON, provider-null → null, no-markers → null) and `createSourceBundle` `collectIl2cppSources` (C# included only when enabled, verified by unzipping the bundle).
- `test/commands/debug-files/upload.test.ts` — `--il2cpp-mapping` queues the `.il2cpp` DIF, absent without the flag, and is threaded through to `uploadDebugFiles`.

All text fixtures (Breakpad `FILE` records + synthetic C++/C#), no binaries.

Verified locally: `typecheck`, `lint`, `test` (358 debug-files/dif/scan tests), `check:deps`, `check:fragments` all green.

### DIFF
--- a/docs/src/fragments/commands/debug-files.md
+++ b/docs/src/fragments/commands/debug-files.md
@@ -38,6 +38,10 @@ sentry debug-files upload ./build --no-zips
 sentry debug-files upload ./dsyms --type dsym --wait
 sentry debug-files upload ./build --id <debug-id> --require-all
 
+# Unity: upload IL2CPP line mappings (optionally with referenced C# sources)
+sentry debug-files upload ./build --il2cpp-mapping
+sentry debug-files upload ./build --il2cpp-mapping --include-sources
+
 # Preview what would be uploaded without uploading (no credentials needed)
 sentry debug-files upload ./build --no-upload
 ```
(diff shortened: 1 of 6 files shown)


BYK:
Superseded by #1165. GitHub auto-closed this PR when its base branch (`byk/feat/debug-files-embedded-ppdb`, #1163) was deleted on merge; the same commit, rebased onto `main`, continues in #1165.

PR #1172, opened 2026-07-02
feat(build): add `build upload` for Android APK/AAB

BYK:
Second PR of the **build/Emerge** workstream (after `build download` #1170). Adds the flagship `build upload` for Android.

## `build upload <paths...>`
Uploads mobile builds to Sentry for preprod size analysis.

- **Format detection** (`src/lib/build/`): APK (root `AndroidManifest.xml`) and AAB (`BundleConfig.pb` + `base/manifest/AndroidManifest.xml`) via ZIP entry names — no decompression. IPA is detected only to emit a clear "iOS not yet supported" message.
- **Normalization**: wraps each build into a deterministic ZIP (build under its basename + `.sentry-cli-metadata.txt`) using **STORE + fixed mtime** via `fflate`. Deterministic bytes let the server dedup chunks across re-uploads.
- **Upload** (`uploadBuild` in `preprod-artifacts.ts`): chunk-upload (reusing `chunk-upload.ts`) + the single-object `projects/{org}/{project}/files/preprodartifacts/assemble/` endpoint, polling until the server returns `artifactUrl` (5-min cap, 1s poll).
- Multiple paths with per-file success/error aggregation; **exits non-zero if any build fails**. `--build-configuration`, `--release-notes`, repeatable `--install-group` fold into the assemble body.
- Org+project resolved from context/flags/env. **Sentry SaaS only.**

### Why not zstd-in-zip (matching the legacy CLI)?
Node's built-in zstd is raw-stream only; Zstd *inside* a ZIP needs a writer emitting method-93 entries, which the JS ZIP libs don't do. It also buys nothing here: the wrapper's method is transparent to the server (it just unzips), the inner APK/AAB is already compressed (so STORE avoids pointless double-compression), and dedup determinism comes from the fixed mtime, not the method. Transport already compresses chunks over the wire.

## Tests (33 across 3 files)
- `build/index.test.ts` — APK/AAB/IPA detection, plugin parsing, normalization determinism + contents.
- `preprod-artifacts.test.ts` — `uploadBuild` assemble-body shape, missing-chunk→upload→artifactUrl (fake timers), error → `ApiError`.
- `build/upload.test.ts` — APK upload, metadata pass-through, unsupported-file + directory rejection, partial-failure exit code.

## Follow-ups (not in this PR)
git/VCS metadata + CI auto-detect; iOS XCArchive/IPA upload (`.car` asset parsing is native-macOS-only and skipped); `snapshots` (`download`/`diff` now, `upload` in the objectstore session).

`typecheck`, `lint`, `check:deps`, `check:fragments` all pass.

### DIFF
--- a/docs/src/content/docs/contributing.md
+++ b/docs/src/content/docs/contributing.md
@@ -54,7 +54,7 @@ cli/
 │   ├── commands/       # CLI commands
 │   │   ├── alert/       # create, delete, edit, list, view
 │   │   ├── auth/        # login, logout, refresh, status, token, whoami
-│   │   ├── build/       # download
+│   │   ├── build/       # download, upload
 │   │   ├── cli/         # defaults, feedback, fix, import, setup, uninstall, upgrade
 │   │   ├── code-mappings/# upload
 │   │   ├── dart-symbol-map/# upload
--- a/docs/src/fragments/commands/build.md
+++ b/docs/src/fragments/commands/build.md
@@ -3,6 +3,15 @@
 ## Examples
 
 ```bash
+# Upload an Android build (APK or AAB) for size analysis
+sentry build upload ./app-release.apk
+
+# Upload with a build configuration and release notes
+sentry build upload ./app.aab --build-configuration Release --release-notes "Nightly"
+
+# Tag a build with install groups (repeatable)
+sentry build upload ./app.aab --install-group qa --install-group beta
+
 # Download a build artifact by ID
 sentry build download 1234567890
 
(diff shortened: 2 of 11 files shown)


BYK:
## Self-review (subagent-assisted) + fixes

Ran a critical review against the legacy Rust source. Key outcome:

**Critical (fixed):** `--install-group` was accidentally *required* (`variadic` without `optional: true`), so the primary `build upload <path>` invocation failed at parse time. Fixed + added a **parser-level regression test** — the previous command tests drove the command via `loader()`, which bypasses Stricli arg parsing, so they couldn't see it. New test runs through `run(app, …)`.

**Also fixed:** fail-fast on a terminal `ok` assemble state without `artifactUrl` (L3, matches legacy); tests for empty-metadata body omission and the ok-without-url throw.

**Two deliberate deviations from the legacy Rust CLI (flagging for visibility):**
1. **Partial-failure exit code** — legacy exits 0 when *some* builds upload and others fail; this port exits **non-zero on any failure** so CI/agents don't mistake a partial upload for success. (Comment updated to state this.)
2. **Wrapper compression** — legacy uses Zstd-in-ZIP; this uses **STORE + fixed mtime** (APK/AAB are already compressed; avoids a method-93 ZIP writer; deterministic for dedup within this CLI). Documented in `src/lib/build/index.ts`.

**Known follow-up:** builds are read fully into memory (`readFile`), so files above Node's ~2 GiB Buffer cap will throw where the legacy mmap path succeeds. Noted in-code; a streaming-normalization follow-up can use the file-based chunk path.

PR #1201, opened 2026-07-06
docs: add v3→v4 migration guide (sentry-cli → sentry)

BYK:
Adds a migration guide for the upcoming **major release** (v3 → v4), where the legacy Rust `sentry-cli` is replaced by the new TypeScript `sentry` CLI.

New page: `docs/src/content/docs/migrating-from-v3.md` (wired into the **Getting Started** sidebar, between Installation and Self-Hosted).

## Covers
- **The rename** — `sentry-cli` → `sentry` binary, `@sentry/cli` → `sentry` npm package, with install/uninstall steps for npm/brew/curl/npx.
- **Command changes**, grouped for quick scanning:
  - *Still work unchanged* (`info`, `send-event`, `send-envelope`, `bash-hook`, `sourcemaps`, `debug-files`, `react-native …`).
  - *Renamed groups* (plural → singular: `releases`→`release`, `projects`→`project`, …) with a note that bare plurals still work as list shortcuts.
  - *Moved* (`login`→`auth login`, `update`→`cli upgrade`, `uninstall`→`cli uninstall`, `deploys`→`release deploys`, `upload-dif`/`upload-dsym`→`debug-files upload`, `difutil`→`debug-files`, `upload-proguard`→`proguard upload`).
  - *Removed* (`send-metric`).
- **A copy-paste `sentry-cli()` shim** that transparently translates every moved/renamed command to its v4 form, so existing scripts and muscle memory keep working during migration (the user-requested "practical alternatives").
- **Output & scripting** (`--json` instead of screen-scraping, `NO_COLOR`/`SENTRY_PLAIN_OUTPUT`), **OAuth device-flow auth** (tokens still honored), **semantic exit codes**, **debug-ID-first sourcemaps**, and **config/`.sentryclirc` backward-compat**.

All command mappings were verified against the actual v4 route map (`src/app.ts`), `src/commands/`, and the existing exit-codes/configuration docs. `cd docs && pnpm run build` passes (page renders, no broken links).

> Note: I inferred the release will be branded **v4**; if it lands under a different version the title/headings are a trivial find-replace.


docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+| `sentry-cli upload-dsym …` | `sentry debug-files upload …` |
+| `sentry-cli difutil check …` | `sentry debug-files check …` |
+| `sentry-cli upload-proguard …` | `sentry proguard upload …` |
+
+### Removed
+
+| v3 | v4 |
+|----|----|
+| `sentry-cli send-metric …` | **Removed** — metrics were deprecated in Sentry. No replacement. |
+
```
szokeasaurusrex: [nit] `sentry-cli send-metric` was already removed in v3.0.

```suggestion
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+### Renamed groups (plural → singular)
+
+Command **groups** are singular now. The plural name still works as a shortcut
+for the bare **list** (`sentry releases` → lists releases), but any
+**subcommand** must use the singular form:
+
+| v3 | v4 |
+|----|----|
+| `sentry-cli organizations …` / `orgs` | `sentry org …` |
```
szokeasaurusrex: l: There is no `sentry-cli orgs` command in v3

```suggestion
| `sentry-cli organizations …` | `sentry org …` |
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+| `sentry-cli projects …` | `sentry project …` |
+| `sentry-cli releases new/finalize/set-commits …` | `sentry release new/finalize/set-commits …` |
```
szokeasaurusrex: [nit] I would align this with the other commands here

```suggestion
| `sentry-cli releases …` | `sentry release …` |
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+    login|logout)            local c=$1; shift; command sentry auth "$c" "$@" ;;
+    update)                  shift; command sentry cli upgrade "$@" ;;
+    uninstall)               shift; command sentry cli uninstall "$@" ;;
+    deploys)                 shift; command sentry release deploys "$@" ;;
+    upload-dif|upload-dsym)  shift; command sentry debug-files upload "$@" ;;
+    upload-proguard)         shift; command sentry proguard upload "$@" ;;
+    difutil)                 shift; command sentry debug-files "$@" ;;
+
+    # Renamed groups (plural → singular) so subcommands keep working
+    organizations|orgs)      shift; command sentry org "$@" ;;
```
szokeasaurusrex: l: As I mentioned above, `sentry-cli orgs` is not a command in `v3`.

```suggestion
    organizations)      shift; command sentry org "$@" ;;
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+    issues)                  shift; command sentry issue "$@" ;;
+    monitors)                shift; command sentry monitor "$@" ;;
+    repos)                   shift; command sentry repo "$@" ;;
+    events)                  shift; command sentry event "$@" ;;
+
+    # Removed
+    send-metric)
+      echo "sentry-cli send-metric was removed in v4 (metrics are deprecated)." >&2
+      return 1 ;;
+
```
szokeasaurusrex: l: Metrics were already removed

```suggestion
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+# Check who you are / token validity
+sentry auth status
+sentry auth whoami        # or the top-level: sentry whoami
+
+# CI / non-interactive — unchanged from v3
+export SENTRY_AUTH_TOKEN=sntrys_…
+sentry auth status
+```
+
+`SENTRY_AUTH_TOKEN` (and the legacy `SENTRY_TOKEN`) continue to take precedence
```
szokeasaurusrex: m: `sentry-cli` v3 does not have a `SENTRY_TOKEN` environment variable, so I am not sure where this comes from.

```suggestion
`SENTRY_AUTH_TOKEN` continue to take precedence
```

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+Sourcemap upload is now **debug-ID-first** and decoupled from releases — you no
+longer have to create a release just to upload maps:
+
+```bash
+# v3
+sentry-cli releases files 1.0.0 upload-sourcemaps ./dist
+
+# v4 — debug-ID based (release optional)
+sentry sourcemap upload ./dist
+```
```
szokeasaurusrex, line 461: m: Sourcemap uploading already was debug-ID first in v3 and even v2 of the Sentry CLI; the command `sentry-cli releases files`, as written here, does not even exist in v3.

The command for uploading sourcemaps in v3 is `sentry-cli sourcemaps upload`; the only difference to v4 is that `sourcmaps` became singular `sourcemap`.

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Getting help
+
+- `sentry --help` — top-level command list
+- `sentry <command> --help` — details and flags for any command
+- [Command reference](/commands/) · [Exit codes](/exit-codes/) ·
+  [Configuration](/configuration/)
+
+If a command you relied on isn't covered here, please
+[open an issue](https://github.com/getsentry/sentry-cli/issues) — we want the
```
szokeasaurusrex: m: Is this the correct link? Shouldn't we link to the `getsentry/cli` repo instead?

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+These are identical, or the old plural form still works as a shortcut:
+
+| v3 | v4 |
+|----|----|
+| `sentry-cli info` | `sentry info` |
+| `sentry-cli send-event …` | `sentry send-event …` |
+| `sentry-cli send-envelope …` | `sentry send-envelope …` |
+| `sentry-cli bash-hook` | `sentry bash-hook` |
+| `sentry-cli sourcemaps …` | `sentry sourcemaps …` (or `sentry sourcemap …`) |
+| `sentry-cli debug-files …` | `sentry debug-files …` |
```
szokeasaurusrex, line 101: Just wondering: why did we choose to keep `sentry debug-files` plural, but we have `sentry sourcemap` as a singular?

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+| v3 | v4 |
+|----|----|
+| `sentry-cli login` | `sentry auth login` |
+| `sentry-cli logout` | `sentry auth logout` |
+| `sentry-cli update` | `sentry cli upgrade` |
+| `sentry-cli uninstall` | `sentry cli uninstall` |
+| `sentry-cli deploys new …` | `sentry release deploys new …` |
+| `sentry-cli upload-dif …` | `sentry debug-files upload …` |
+| `sentry-cli upload-dsym …` | `sentry debug-files upload …` |
+| `sentry-cli difutil check …` | `sentry debug-files check …` |
```
szokeasaurusrex, line 142: l: FWIW, these commands were already soft-deprecated in v3 in favor of the new commands listed here (they are hidden from the `--help` texts). We may wish to indicate that to avoid confusing users who may not have been aware of these legacy names and/or we can just avoid mentioning them

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,285 @@
...
+For CI, either update your install step to `npm install -g sentry` and call
+`sentry`, or add a one-line shim in your job:
+
+```bash
+# GitHub Actions / any CI shell
+npm install -g sentry
+sentry-cli() { sentry "$@"; }   # if you didn't use any moved commands
+```
+
+## Command changes
```
szokeasaurusrex, line 88: h: One overarching question here: do all the arguments continue to work the same as they did with the old CLI? Including the global arguments, like `--url` for specifying the self-hosted URL?

If not, this definitely needs to be called out, and we also should address it in the shim.

szokeasaurusrex (commented):
Added some comments.

Most of them are small; my biggest concern is about the lack of mention about whether any of the command line flags need to be changed

szokeasaurusrex:
Had the clanker take a quick look at the differences between the two CLI's APIs for some of the more important commands; seems like the new CLI lacks support for several flags that were supported in the old CLI.

The full investigation [is here](https://gist.github.com/szokeasaurusrex/696534571b95131d89726ee07fac516a).

docs/src/content/docs/migrating-from-v3.md
```diff
@@ -0,0 +1,347 @@
...
+| `sentry-cli upload-dif …` | `sentry debug-files upload …` |
+| `sentry-cli upload-dsym …` | `sentry debug-files upload …` |
+| `sentry-cli difutil check …` | `sentry debug-files check …` |
+| `sentry-cli upload-proguard …` | `sentry proguard upload …` |
+
+### Removed
+
+| v3 | v4 |
+|----|----|
+| `sentry-cli send-metric …` | **Removed** — metrics were deprecated in Sentry. No replacement. |
```
lcian: Technically we have application metrics now, only the old flavor of metrics was deprecated.
Maybe this is worth supporting (both sending and querying) in this new CLI?

docs/src/content/docs/migrating-from-v3.md
BYK: Great catch — this was the biggest gap. Added a **Global flags and options** section: v4 keeps `--org`/`--project`/`--log-level`/`-v` as globals but drops per-command `--auth-token`/`--url`/`--header` (→ `SENTRY_AUTH_TOKEN` / `SENTRY_URL`|`SENTRY_HOST` / `SENTRY_CUSTOM_HEADERS`) and `--quiet`. The shim now also translates `--auth-token`/`--url`/`--header` into those env vars per-call, and there's a caution to confirm command-specific flags with `sentry <cmd> --help`. (Your gist was really useful — one note: `sentry debug-files upload` *does* exist in the current CLI, so that row in the 0.38.0 delta is stale.)

docs/src/content/docs/migrating-from-v3.md
BYK: Honestly an inconsistency: `debug-files` keeps the legacy `sentry-cli debug-files` name verbatim for compat/muscle-memory, while the newer resource groups use the singular convention (`sourcemap`, with `sourcemaps` aliased). Not really a migration-guide matter — happy to file a follow-up if we want to reconcile it (e.g. add a `debug-file`/`sourcemaps` alias pair).

BYK:
Thanks @szokeasaurusrex — all addressed in the latest push:

**Factual fixes**
- Dropped the `orgs` alias everywhere (v3 only had `organizations`).
- Removed the `send-metric` "Removed" row + shim case — it was removed back in v3.0, not a v3→v4 change.
- Corrected the sourcemaps section: v3 command is `sentry-cli sourcemaps upload` (not `releases files … upload-sourcemaps`), and dropped the false "now debug-ID-first" framing (it's been debug-ID-first since v2). Now documents the real deltas: single `<directory>` positional + the dropped flags.
- Clarified `SENTRY_TOKEN` is a v4 alias, not a v3 carryover.
- Noted `upload-dif`/`upload-dsym`/`upload-proguard` were already soft-deprecated hidden aliases in v3.
- Fixed the "open an issue" link → `getsentry/cli`; aligned the renamed-group rows.

**The big one (flags)** — new **Global flags and options** section documenting which globals survive (`--org`/`--project`/`--log-level`/`-v`) vs. which moved to env vars (`--auth-token`→`SENTRY_AUTH_TOKEN`, `--url`→`SENTRY_URL`/`SENTRY_HOST`, `--header`→`SENTRY_CUSTOM_HEADERS`) vs. dropped (`--quiet`). The shim now translates the env-var-backed flags automatically, and there's a caution to verify per-command flags via `sentry <cmd> --help`.

**Codemod** (Bugbot) — the `.execute()`→`.run()` and `.releases.*` rewrites are now gated on a tracked `SentryCli` instance, so unrelated `.execute()` APIs (DB clients, etc.) are left alone. Added regression tests.

One correction on the gist: `sentry debug-files upload` **does** exist now (built recently), so that "no equivalent" note reflects an older `0.38.0`.

codemods/sentry-v3-to-v4.cjs
```diff
@@ -0,0 +1,250 @@
...
+    }
+    const args = path.node.arguments;
+    // The v3 constructor is (configFile, options); v4 takes just options.
+    let options = null;
+    if (args.length >= 2) {
+      options = args[1];
+    } else if (args.length === 1 && args[0].type === "ObjectExpression") {
+      options = args[0];
+    }
+    if (options && options.type === "ObjectExpression") {
```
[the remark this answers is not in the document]
BYK: Good catch — fixed. A single non-object argument is now preserved as `createSentrySDK(myVar)` (instead of being dropped when it wasn't an object literal), with a TODO flagging the v3 ambiguity: v3's first constructor param was a `configFile` path (removed in v4), so the arg is either that (drop it) or an options variable (map `authToken`→`token`). Nothing is silently lost now. Added a regression test.

codemods/sentry-v3-to-v4.cjs
```diff
@@ -0,0 +1,250 @@
...
+  newDeploy: ["release", "deploy"],
+};
+
+/** Methods whose first positional (a release/version) moves into the options object. */
+const RESHAPE = {
+  new: { key: "orgVersion" },
+  finalize: { key: "orgVersion" },
+  setCommits: { key: "orgVersion", spreadSecond: true, todo: "verify set-commits options (repo/commit/auto → commit/auto/local)" },
+  uploadSourceMaps: { key: "release", spreadSecond: true, todo: "sourcemaps are debug-ID-first: map `include` → the `directory` positional and review options" },
+  newDeploy: { key: "orgVersionEnvironmentName", spreadSecond: true, todo: "verify deploy options (env/name/url/started/finished/time)" },
```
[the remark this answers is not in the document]
BYK: Right — `release.deploy` takes env/name as part of the positional `orgVersionEnvironmentName` (org/version/env/name), not option keys. The codemod no longer spreads the v3 options for deploy; it emits `release.deploy({ orgVersionEnvironmentName: v })` plus a TODO explaining that env/name go into the positional target and url/started/finished/time are options. Fixed the guide's mapping row to match, and added a test asserting env isn't spread.

docs/src/content/docs/migrating-from-v3.md
BYK: Good point. I've removed that "Removed" row entirely — the old statsd-style `send-metric` was deprecated back in v3.0, so it doesn't belong in a v3→v4 guide, and calling metrics "deprecated" was misleading given application metrics exist now. Adding first-class metrics support (send + query) to the new CLI is a great idea but its own feature — worth a separate issue rather than this migration doc.

codemods/sentry-v3-to-v4.cjs
```diff
@@ -0,0 +1,275 @@
...
+    if (options && options.type === "ObjectExpression") {
+      for (const prop of options.properties) {
+        if (prop.key && (prop.key.name === "authToken" || prop.key.value === "authToken")) {
+          prop.key = j.identifier("token");
+        }
+      }
+    }
+    path.replace(
+      j.callExpression(j.identifier("createSentrySDK"), options ? [options] : [])
+    );
```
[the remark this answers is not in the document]
BYK: Fixed. When the options are an inline object literal the codemod renames `authToken`→`token` in place; when they're a variable/expression it can't reach in, so it now emits a TODO to update it manually. Added a test for the variable case.

codemods/sentry-v3-to-v4.cjs
```diff
@@ -0,0 +1,275 @@
...
+      }
+      callArgs = [j.objectExpression(props)];
+    }
+
+    path.replace(
+      j.callExpression(member(j.memberExpression(recv, j.identifier(route)), target), callArgs)
+    );
+    if (reshape && reshape.todo) {
+      addTodo(path, reshape.todo);
+    }
```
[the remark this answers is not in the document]
BYK: Good defensive catch — reordered so every `addTodo` runs before `path.replace()`. (It happened to work after replace since the statement ancestor is unaffected, but pre-replace is clearly correct.)

BYK:
Two more codemod findings from the latest review, both addressed in `bd6263352`:

- **"Codemod breaks custom import names"** — the codemod no longer forces the `createSentrySDK` binding. It keeps your local name and only changes the module specifier (`@sentry/cli` → `sentry`), rewriting `new <name>(…)` → `<name>(…)` (v4's default export is the `createSentrySDK` factory, not a class). Re-exports, passing the module around, and multiple `require`s all stay valid — no dangling refs or duplicate declarations. Added a test with `export { Sentry }`.

- **"authToken not renamed in variables"** — intended behavior: `authToken`→`token` is renamed in place when the options are an inline object literal; when they're a variable/expression defined elsewhere the codemod can't reach in, so it emits a `// TODO(sentry-v4): … authToken → token` breadcrumb instead of silently doing nothing. Covered by the "flags authToken rename when options is a variable" test.

BYK:
Latest Bugbot review — all three addressed in `e42fc0073`:

- **Execute rewrite keeps broken argv** — the `execute()`→`run()` rewrite now emits a `// TODO(sentry-v4)` noting the argv tokens are raw v3 CLI args that may need remapping to v4 command names (`releases`→`release`, `new`→`create`, `login`→`auth login`, …). The codemod can't reliably remap arbitrary argv arrays, so it flags rather than guesses.
- **authToken shorthand key rename** — `{ authToken }` shorthand now expands to `{ token: authToken }` (keeps referencing the original binding) instead of printing a bare `{ token }`.
- **Shim drops repeated headers** — the shim now merges multiple `--header` flags into one semicolon-separated `SENTRY_CUSTOM_HEADERS` (`X-A: 1; X-B: 2`), which is v4's expected format, instead of overwriting.

Codemod tests: 16, all green.

BYK:
Both addressed:

- **Shim breaks list and create** — the shim now maps a **bare** `sentry-cli releases`/`projects`/… to `sentry <singular> list` (matching v4's native `sentry releases` → `release list`); a plural **with** a subcommand routes to the singular group. Note `new` *is* a valid alias for `create` in v4 (buildRouteMap auto-injects `create`→`new`), so `sentry release new …` works — the guide example is correct.
- **Unused jscodeshift dependencies** — removed `jscodeshift`, `@types/jscodeshift`, and `tslib` (the codemod moved to Codemod/jssg, run via `npx codemod`); lockfile regenerated.

Also rebased on `main` and resolved the package.json/lock conflict.

BYK:
Addressed the latest round:

- **Shim mishandles deploys list** — `deploys` now branches: `new` → `sentry release deploy`, `list` → `sentry release deploys` (drops the v3 `list` subcommand, since `release deploys` *is* the list leaf), bare/other → `sentry release deploys`.
- **Codemod spreads invalid deploy options** — `newDeploy` no longer spreads the v3 options object. v4's `ReleaseDeployParams` has no `env`/`name` keys (they fold into the `orgVersionEnvironmentName` positional), so the spread produced non-compiling code. It now emits `release.deploy({ orgVersionEnvironmentName: <version> })` + a TODO covering env/name → positional and url/started/finished/time → options.
- **Unpinned `codemod@latest` in OIDC publish job** — pinned to `codemod@1.12.7` (matches the CI `codemod-test` job).

(The two earlier findings that reappeared at line `:0` are the already-fixed shim/deps items — they're outdated against the current diff.)

BYK:
- **Shim drops deploy env flags** (High) — fixed. Since v4's `release deploy` takes environment/name as *positionals* (not `-e`/`-n` flags), a bash shim can't transparently reorder them. The shim now prints the new positional syntax and returns a usage error for `deploys new` instead of silently forwarding broken flags. `deploys`/`deploys list` still map to `release deploys`.

- **Codemod spreads invalid deploy options** (codemod.ts:49) — this is a false positive. `newDeploy` does **not** use `spread: true` (that flag is only on `setCommits`/`uploadSourceMaps`, which spread *valid* options). The committed fixture proves it: `await cli.release.deploy({ orgVersionEnvironmentName: "1.0.0" })` — no `env`/`name` keys, plus a TODO. No change needed.

BYK:
Latest round:

- **Shim steals command `--url` flags** (High) — fixed. The shim now translates the v3 global `--auth-token`/`--url`/`--header` only in the *leading* flag run and stops at the first command word. Command-level `--url` (e.g. `release deploy … --url <deploy-url>`) is passed through untouched. Verified: `--url X releases list` → `SENTRY_URL=X` + `releases list`; `release deploy 1.0.0 prod --url Y` → no env translation, `--url Y` preserved.
- **Doesn't handle `import * as … from "@sentry/cli"`** (sentry[bot]) — fixed. Namespace-import bindings are now collected, so the specifier rewrite applies (`import * as Sentry from "sentry"`) instead of being dropped by the no-binding early return. Added a fixture (17 total).
- **Codemod spreads invalid deploy options** (codemod.ts:49) and **Shim drops deploy env flags** (docs) — these are stale/false: `newDeploy` has no `spread` (committed fixture emits `release.deploy({ orgVersionEnvironmentName: "1.0.0" })`, no `env`/`name`), and `deploys new` already prints the new positional syntax + returns rather than forwarding flags.

BYK:
- **Incorrect TODO on `new SentryCli(cfg, null)`** (sentry[bot], codemod.ts:171) — fixed. A `null`/`undefined` second arg was being assigned to `options` (a truthy AST node), triggering a bogus authToken TODO. Now both the configFile and the null options are dropped → `SentryCli()` with no TODO. Added a fixture (18 total).

The three Cursor findings still showing (`codemod.ts:49` spread, `docs` deploy-env-flags, `docs` `--url`) are the same persistent threads from earlier commits, re-associated with each push — they were already addressed (`newDeploy` has no `spread`; `deploys new` prints the new syntax + returns instead of forwarding flags; `--url` is only translated in the leading global-flag run). Verified via the committed fixtures and a functional shim test.

BYK:
- **Shim drops leading global flags** (High) — fixed. The leading-flag loop now recognizes the still-valid v4 globals (`--org`/`--project`/`--log-level`/`--fields`/`-v`/`--json`) and consumes them (with their values) into a `lead` prefix that's re-applied before the translated command. So `--auth-token`/`--url`/`--header` are still translated even when another global precedes them, and scanning stops only at the real command word.

Verified with a functional harness:
- `--org acme --auth-token X releases list` → `SENTRY_AUTH_TOKEN=X` + `sentry --org acme release list`
- `--url https://self releases list` → `SENTRY_URL=…` + `sentry release list`
- `--auth-token X foo --url http://y` → `SENTRY_AUTH_TOKEN=X` + `sentry foo --url http://y` (command-level `--url` preserved)
- `--log-level debug releases new 1.0.0` → `sentry --log-level debug release new 1.0.0`

BYK:
- **Broken deploy SDK migration** (High) — fixed. You're right: the typed `sdk.release.deploy` collapses version/environment/name into a single `orgVersionEnvironmentName` positional and forwards it as one argv token (`invoke([...], {...}, [params.orgVersionEnvironmentName])`), so it can't supply the required `<environment>` — the call fails at runtime.

  Both the codemod and guide now use the raw `run(...)` escape hatch for deploys:
  - Codemod: `cli.releases.newDeploy("1.0.0", {...})` → `cli.run("release", "deploy", "1.0.0")` + a TODO to add the `<environment>`/`[name]` positionals and `--url`/`--started`/`--finished`/`--time` flags.
  - Guide row: `sdk.run("release", "deploy", v, env, name, "--url", url)`.

BYK:
Three more, all fixed:

- **uploadSourceMaps fails to rename include→directory** (High, sentry[bot]) — the codemod no longer spreads the v3 options. `include` is a path array but v4 `sourcemap.upload` takes a single `directory` string, so spreading emitted an invalid `include` key. Now emits `sdk.sourcemap.upload({ release })` + a TODO to set `directory` (one upload per directory).
- **Shim loops on bare flags** (High) — value-taking leading flags used `shift 2`, which hangs the shell when the value is missing. Changed to `shift 2 2>/dev/null || shift`; verified value-less `--org`/`--auth-token` complete instead of looping.
- **Shim breaks releases deploys** (High) — v3 nested deploys under `releases`, so `releases deploys …` fell through to `release deploys …` with invalid v3 subcommands/flags. Both top-level `deploys` and `releases deploys` now print the manual-migration note (v4 create uses positional environment/name). Verified `releases deploys 1.0.0 new -e prod` → note + exit 64.

BYK:
- **Shim rejects listable deploys** (Medium) — fixed. A small `_scli_deploys` helper (shared by the top-level `deploys` and nested `releases deploys` paths) now maps bare/`list` to `sentry release deploys` (dropping the v3 `list` token); only an invocation containing `new` prints the manual-migration note. Verified:
  - `deploys 1.0.0 list` / `releases deploys 1.0.0 list` → `sentry release deploys 1.0.0`
  - `deploys 1.0.0 new -e prod` / `releases deploys 1.0.0 new …` → note + exit 64
  - `releases list` → `sentry release list`; value-less `--org` no longer hangs

docs/src/content/docs/migrating-from-v3.md
BYK: Fixed — the fictitious `sentry-cli releases files` command is gone. The Sourcemaps section now shows `sentry-cli sourcemaps upload ./dist` → `sentry sourcemap upload ./dist` (only the plural→singular rename, plus the aliased plural), and the mapping note states it's "keyed by debug ID, as it has been since v2." Thanks for the correction.

docs/src/content/docs/migrating-from-v3.md
BYK: Added a note right under that table: "Most of these were already soft-deprecated in v3 (hidden from `--help` in favor of `debug-files` / `proguard`); v4 simply drops the legacy top-level spellings."

docs/src/content/docs/migrating-from-v3.md
BYK: Intentional inconsistency: `debug-files` keeps the legacy `sentry-cli debug-files` spelling verbatim for muscle-memory/compat, while newer resource groups use the singular convention (`sourcemap`, with `sourcemaps` aliased). Not really a migration-guide concern — I can file a follow-up to reconcile the naming if we want, but I'd keep it out of this PR.

BYK:
Ran an independent flag-compatibility review pass. The common path (`--auth-token`/`--url`/`--header` → env vars, dual-meaning command-level `--url`, shim parsing, sourcemap flag deltas) all verified correct. Fixed the real gaps it surfaced:

- **`--allow-failure` / `SENTRY_ALLOW_FAILURE`** (v3 global that forced exit 0) — not implemented in v4, so the shim would have passed it through and errored. The shim now strips it and emulates the old "never fail" behavior, and the Dropped section documents it. Verified: failing cmd → rc=3 normally, rc=0 with the flag/env, and it's never forwarded to `sentry`.
- **Codemod constructor** — v3 `SentryCliOptions` beyond `authToken` (esp. auth-critical `apiKey`, plus `url`/`dsn`/`silent`/`customHeader`/`org`/`project`/`vcsRemote`) were passed through silently. Now flagged with a TODO. Added a fixture (19 total).
- **`setCommits` TODO** now names the actually-dropped `ignoreMissing`/`ignoreEmpty` keys.
- **Guide Dropped section** — added `--silent`, `set-commits --ignore-empty`, `SENTRY_API_KEY`/`SENTRY_VCS_REMOTE` no-longer-honored, and the self-hosted-only note for `SENTRY_CUSTOM_HEADERS`.

BYK:
Two more findings addressed:

- **`ci-status` didn't gate on `codemod-test`** (High) — fixed. Added `codemod-test` to the `ci-status` `needs` array, the failure/cancel result scan, and a skipped-cascade guard (mirroring the `code`/`skill` checks, using the existing `changes.outputs.codemod` filter). A codemod-test failure now blocks merge.
- **`execute(args, live)` → `run(...)` drops `live`** (Medium) — the TODO now calls this out. One correction: `live` isn't a dry-run toggle — in `@sentry/cli`'s `helper.js`, `live:true`/`'rejectOnError'` inherit stdio (stream output), while `live:false`/absent captures and returns stdout. v4 `run()` always captures/returns (the `live:false` behavior), so the TODO notes that callers who passed `live:true`/`'rejectOnError'` for streamed stdio should adjust. No behavior is silently changed toward "live" — if anything it matches v3's default.

BYK:
- **Codemod mislabels valid SDK options** (Medium) — fixed. `token`/`url`/`org`/`project` are first-class `SentryOptions` keys, so I narrowed the constructor TODO to only the genuinely-unmapped v3 keys: `apiKey` (→ `token`), `dsn` (→ `SENTRY_DSN`), and `silent`/`customHeader`/`headers`/`vcsRemote` (dropped). `new SentryCli(null, { authToken, org })` now migrates cleanly with no TODO; only unmapped keys are flagged.

BYK:
- **`new X.default()` namespace imports not migrated** (High) — fixed. Namespace-import names (`import * as X from "@sentry/cli"`) are now tracked separately, and the constructor detection also matches `new X.default(...)` (where `X.default` is the v3 class). `import * as Sentry` + `new Sentry.default({ authToken })` now migrates fully to `Sentry.default({ token })` with `execute`→`run`. Added a `namespace-default` fixture.

PR #1204, opened 2026-07-07
docs: note `sentry cli setup` for package manager installs

sergical:
## Summary

Package manager installs (`npm`/`pnpm`/`yarn`/`bun -g`) place the binary on PATH but — unlike the install script and Homebrew — never run `sentry cli setup`. As a result, shell completions and agent skills are silently missing for those users, with nothing in the docs pointing them to the fix.

This adds a short note in the **Package Managers** section of the getting-started guide directing those users to run `sentry cli setup` once.

## Why here / why this wording

- Scoped to the Package Managers section because that's the only audience affected — curl and Homebrew installs already run setup automatically.
- Explicitly contrasts with the install script / Homebrew so curl/brew readers know it doesn't apply to them.
- States the concrete gap (completions + agent skills) rather than a generic pitch, and omits auth (setup doesn't touch credentials, so it's not relevant to a reader here).

### DIFF
--- a/docs/src/content/docs/getting-started.mdx
+++ b/docs/src/content/docs/getting-started.mdx
@@ -84,6 +84,12 @@ Install globally with your preferred package manager (the npm/pnpm/yarn packages
   bun="bun add -g sentry"
 />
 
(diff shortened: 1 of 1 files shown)


BYK (approved):
Let's go 🚀

PR #1194, opened 2026-07-03
chore(deps): bump @sentry/api to 0.253.0 and adopt token operationIds

betegon:
## What

Bumps `@sentry/api` `^0.180.0` → `^0.253.0` and migrates every SDK call site from the old normalizer-derived names to the clean token `operationId`s produced by the [operationId migration](https://github.com/getsentry/sentry-api-schema/issues/76).

Examples of the rename (34 symbols across 12 files):
- `retrieveAnOrganization` → `getOrganization`
- `listAnOrganization_sIssues` → `listOrganizationIssues`
- `queryExploreEventsInTableFormat` → `listOrganizationEvents`
- `retrieveAnOrganization_sRelease` → `getOrganizationRelease`
- `createANewProject` → `createTeamProject`, `listAProject_sClientKeys` → `listProjectKeys`, etc.

## Aliasing for name collisions

A few new tokens (`getOrganization`, `getProject`, `deleteProject`, `listProjectEnvironments`, `listProjectTeams`) collide with the CLI's own wrapper functions of the same name — the old verbose names never did. Those SDK imports are aliased (`getOrganization as sdkGetOrganization`, …), following the `sdk*` aliasing convention already used elsewhere in the codebase.

## Notes

- **No behavior change** — pure rename + version bump. The response/param shapes the CLI uses are unchanged across 0.180→0.253 (the earlier 0.180 bump already handled downstream type/shape adaptations).
- `generate:sdk` is unaffected (it generates the CLI's own command SDK, not from `@sentry/api`).

## Test plan

- `tsc --noEmit`: clean
- unit tests: API-lib (223), types + commands (2250) — all green

PR #1213, opened 2026-07-09
ref(api): Adopt SDK types for the projects domain

betegon:
## What

A small, bounded **testbed** for the CLI type-cleanup effort — the follow-up to the `@sentry/api` bump (#1194, which renamed call sites but reduced no code). This slice uses the SDK *as intended* for the **projects** domain and establishes the overlay convention the remaining domains will follow. It intentionally does **not** touch all the types.

Projects was chosen because it exercises every realistic case in one cohesive, low-blast-radius slice (all imports flow through the `types/index.ts` barrel): clean SDK adoption, an undocumented-field overlay, the `?collapse` response variant, and a genuine escape hatch.

## Changes

- **Convert `createProjectWithAutoTeam`** from a raw `apiRequestToRegion` POST to the SDK **`createOrganizationProject`** function (same `getOrgSdkConfig` + `unwrapResult` pattern as the rest of the file). `team_slug` (returned by the API but absent from the OpenAPI spec) is kept as a documented typed overlay.
- **Simplify `getProject`**: collapse the whole-argument + result double-cast to a single localized `query: { collapse: 'organization' } as never`. The `collapse` param is server-supported but not in the spec; the trimmed `organization` payload is modeled by the `SentryProject.organization` overlay.
- **Annotate `findProjectByDsnKey`**: its internal `?query=dsn:` filter is not in the public API, so it stays a raw call — now clearly documented as an intentional escape hatch and a backend `@extend_schema` promotion candidate (rather than a silent cast).
- **Codify the overlay convention** on `SentryProject`: the SDK type carries every *documented* field; an overlay adds **only** fields the API returns but the spec omits (`organization`, `status`, `team_slug`), each a backend `@extend_schema` candidate.

## Why not remove more types

Full removal is blocked by four things, none of them CLI-side: loosely-typed endpoints (SDK type thinner than hand-written), undocumented-but-returned fields, `expand`/`collapse` response variants, and private endpoints. These are captured as a backend unlock worklist and are out of scope here.

## Test plan

- `tsc --noEmit`: clean
- `biome check`: clean
- Unit tests: api/types/commands (2738) + api-client coverage / create-sentry-project / project-create (151) — all green. `createProjectWithAutoTeam` coverage asserts the POST target + `team_slug` return, unchanged.
- **Live** (branch source, org `bete-dev`): `project view` returns the project with both the `status` and `organization` overlays intact — behavior unchanged.

No behavior change.

### DIFF
--- a/packages/cli/src/lib/api/projects.ts
+++ b/packages/cli/src/lib/api/projects.ts
@@ -5,6 +5,7 @@
  */
 
 import {
+  createOrganizationProject,
   createTeamProject,
   listOrganizationProjects,
   listProjectKeys,
@@ -27,7 +28,6 @@ import {
 } from "../db/project-cache.js";
 import { getCachedOrganizations } from "../db/regions.js";
 import { type AuthGuardSuccess, withAuthGuard } from "../errors.js";
-import { resolveOrgRegion } from "../region.js";
 import { getApiBaseUrl } from "../sentry-client.js";
 import { buildProjectUrl } from "../sentry-urls.js";
 import { isAllDigits } from "../utils.js";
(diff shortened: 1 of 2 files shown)


BYK:
Jared, people cannot review their own changes reliably 😅 You review it and merge if you think it's good to go

PR #1258, opened 2026-07-16
fix(init): terminate verification process trees

betegon:
## Summary

Post-init verification currently kills only the direct shell process and waits for its close event. Package scripts that launch concurrently, framework dev servers, or file watchers leave descendants holding stdout and stderr open, so sentry init can remain stuck on Verifying setup even though the remote workflow already succeeded.

This starts the verification command in a dedicated POSIX process group and terminates the complete tree during cleanup and signal forwarding. POSIX cleanup is bounded: it sends SIGTERM, escalates to SIGKILL after five seconds, and destroys the local pipe ends so inherited descriptors cannot keep the CLI alive. Windows force-terminates the disposable verification tree with a bounded `taskkill /T /F` invocation and falls back to the direct child if that fails.

The verification path now also handles asynchronous child-process spawn errors. A missing or invalid dev command produces a best-effort verification warning instead of crashing a successful init run.

## Safety properties

- Shell pipelines, concurrently tasks, framework servers, and watchers are terminated as one process tree.
- POSIX descendants that ignore SIGTERM are force-killed after the grace period.
- Cleanup never waits indefinitely for ChildProcess close.
- SIGINT and SIGTERM are forwarded to the POSIX process group; Windows force-terminates the disposable tree before the signal is re-emitted by the CLI.
- Windows `taskkill` calls have their own one-second timeout, debug diagnostics, and direct-child fallback.
- Spawn failures remain non-fatal because verification runs after setup has succeeded.

## Test plan

- `pnpm exec vitest run test/lib/init/verify-setup.test.ts test/lib/init/verify-setup-windows.mocked.test.ts test/lib/dev-script.test.ts test/lib/init/wizard-runner.test.ts` — 75 tests pass, including forced Windows tree cleanup, timeout/status diagnostics, and direct-child fallback.
- `TZ=UTC pnpm exec vitest run test/lib test/commands test/types test/script --coverage --exclude test/lib/completions.property.test.ts` — 400 files, 8,459 tests pass, 14 skipped.
- `pnpm exec tsc --noEmit` — passes.
- `pnpm run lint` — 919 files pass.
- `SENTRY_CLIENT_ID=test-build-only pnpm tsx script/build.ts --single` — darwin-arm64 build passes.
- `git diff --check` — passes.

The unmodified broad test command reaches 8,470 passing tests and 8 unrelated failures. Six time-range assertions assume UTC while the local timezone is Europe/Madrid. Two Bash completion simulations fail on the system Bash 3.2 because the generated script uses extglob syntax without enabling extglob; neither area is touched by this PR.

### DIFF
--- a/src/lib/init/verify-setup.ts
+++ b/src/lib/init/verify-setup.ts
@@ -12,7 +12,7 @@
  * or fatal error patterns in stderr indicate failure.
  */
 
-import { type ChildProcess, spawn } from "node:child_process";
+import { type ChildProcess, spawn, spawnSync } from "node:child_process";
 import { resolve } from "node:path";
 import { captureException } from "@sentry/node-core/light";
 import { createSpotlightBuffer } from "@spotlightjs/spotlight/sdk";
@@ -25,6 +25,18 @@ import type { WizardUI } from "./ui/types.js";
 
 /** Verification timeout in seconds. */
 const VERIFY_TIMEOUT_S = 15;
+
+/** Grace period before escalating verification process cleanup to SIGKILL. */
+const PROCESS_SHUTDOWN_GRACE_MS = 5000;
+
+/** Maximum time to wait after force-killing the verification process tree. */
+const PROCESS_FORCE_SHUTDOWN_MS = 1000;
+
+/** Poll interval while waiting for a verification process tree to exit. */
+const PROCESS_EXIT_POLL_MS = 50;
+
+/** Maximum time for a Windows taskkill invocation to complete. */
+const WINDOWS_TREE_KILL_TIMEOUT_MS = 1000;
 
 /**
  * Patterns in stderr/stdout that indicate a fatal startup failure.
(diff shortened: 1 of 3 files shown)


MathurAditya724 (commented):
Reviewed the diff, ran the suite locally against this branch.

**Verified**
- `pnpm exec vitest run test/lib/init/verify-setup.test.ts test/lib/init/verify-setup-windows.mocked.test.ts` — 5 pass, including the two real-process-tree tests (`shell → parent → child`) and the SIGKILL-escalation path for SIGTERM-ignoring descendants.
- `pnpm exec vitest run test/lib/init` — 425 pass.
- `biome check` clean on all 3 changed files. (typecheck only shows pre-existing missing-generated-module errors, none from this change.)

The core approach is sound: `detached` on POSIX to get a process group, `process.kill(-pid, sig)` to signal the whole tree, bounded poll-for-exit via `process.kill(-pid, 0)`, and destroying the local pipe ends in `finally` so cleanup can't hang on `close`. The tests exercise the actual failure mode (descendants inheriting stdio) rather than mocking it, which is exactly right.

Two things worth a look — neither blocking:

1. **Windows SIGTERM returns before confirming the tree is dead.** In `cleanupProcessTree`, `taskkill /T` (no `/F`) only *requests* graceful shutdown; a status of 0 means the command ran, not that the tree exited. The POSIX path polls for actual exit and escalates to SIGKILL, but Windows returns immediately on a `/T` success and never escalates to `/F` unless `/T` itself fails. A watcher that ignores WM_CLOSE could survive. The pipe `destroy()` in `finally` prevents the CLI from hanging, so this is a leaked-process concern rather than a hang. Given the 5s-then-SIGKILL contract on POSIX, consider whether Windows should also verify/force after the graceful attempt.

2. **`signalProcessTree` swallows taskkill `timeout`/`error` silently.** On the win32 branch, `taskkill.error` (e.g. the 1s timeout firing) makes the function return `false` with no `logger.debug`, unlike the POSIX branch which logs non-ESRCH failures. Minor, but a timed-out taskkill vanishing from debug logs will make Windows cleanup issues hard to diagnose.

Not approving only because I can't exercise the Windows path on real hardware, but the logic and coverage look solid.

MathurAditya724 (approved):

PR #1260, opened 2026-07-16
feat(db): support Node.js 18+ for the npm package via WASM SQLite fallback

BYK:
## Why

The npm package required Node.js 22.15+ purely because our config/cache/auth layer depends on the built-in `node:sqlite` module, which older runtimes don't ship. That excluded users still on Node 18/20 for no fundamental reason — the CLI's actual SQLite footprint is tiny (prepare → get/all/run, exec, manual transactions).

A user asked for Node 20 support and proposed `better-sqlite3`. We pushed back on that: a native addon breaks the single-file bundle, reintroduces per-platform/per-ABI install flakiness, and violates our zero-runtime-dependencies rule.

## Approach

Fall back to a pure-WASM driver (`node-sqlite3-wasm`) on Node < 22.15, selected at runtime behind the existing single-file adapter:

- **22.15+** → native `node:sqlite` (unchanged fast path)
- **18–22.14** → bundled WASM driver

It's a bundled devDependency, so the single-file bundle and no-runtime-deps guarantees both hold.

Critically, the driver is `require()`d lazily so the standalone binary build externalizes it: the SEA binary always embeds a modern LTS Node and uses native `node:sqlite`, making the WASM path dead code there that adds **zero bytes** to the binary. This was the hard constraint for taking this route over `better-sqlite3`.

## Tradeoff

The WASM fallback can't use SQLite WAL mode, so its local cache is slower and less concurrency-friendly. Acceptable for a single-process CLI, and the docs now steer people toward the standalone binary or Node 22.15+.

## Version floors

Dev tooling still needs 22.15+ (it runs the sources against `node:sqlite` via tsx). So `engines.node` advertises the consumer floor (`>=18`) while a new `devEngines` field records the development floor — the doc generator was decoupled to keep the two independent.

CI now runs the npm-package job on a Node 20 matrix entry so the WASM fallback is exercised, not just shipped.


BYK:
Thanks — both findings were valid and are fixed in 174edeb98:

**@sentry-warden (driver never bundled):** Correct, and worse than it looked — the fallback only appeared to work because `node_modules` was present in the dev/CI checkout. A real `npm i -g sentry` on Node < 22.15 would've hit `MODULE_NOT_FOUND` since `node-sqlite3-wasm` is a devDependency. Fixed by adding `db/sqlite` to the `require-alias` filter so esbuild actually inlines the driver. Verified against an isolated dir with **no** `node_modules`: import succeeds, the sibling `.wasm` loads via `__dirname`, schema initializes, and `auth status` returns exit 10. The SEA binary still externalizes the driver (0 driver symbols embedded).

**@cursor (live lock removal):** Valid race. `clearStaleWasmLock` now only removes the `<db>.lock` directory when its mtime is older than a generous staleness window (60s), so a lock a concurrent process just acquired is left to `busy_timeout` instead of being deleted out from under it. Added a regression test covering both stale-cleared and recent-preserved cases.

BYK:
Triaged the latest Seer (sentry[bot]) findings — all four reviewed the cumulative diff and two were already resolved by later commits; the other two are now fixed in 7be144c74:

- **`bundle.ts` missing `db/sqlite` (CRITICAL):** Already fixed in 174edeb98 — `REQUIRE_ALIAS_FILTER` is `/(?:db[\\/](?:index|schema|sqlite)|list-command|telemetry)\.ts$/` and the driver is verifiably inlined (confirmed against a no-`node_modules` install). False positive on current HEAD.
- **Race in `close()` rmdir (MEDIUM):** Already fixed in 90173ec0a — `close()` no longer touches the lock dir; the driver releases our own lock because the `.get()` wrapper finalizes cursors. False positive on current HEAD.
- **Unconditional `PRAGMA journal_mode = WAL` (MEDIUM):** Valid. Now gated on `driverKind === 'node'`; the WASM path stays in rollback-journal mode (fine for a single-process CLI cache) instead of issuing a silent no-op. Verified native still persists `journal_mode=wal`.
- **`__setDriverForTests('node')` silent WASM fallback (MEDIUM):** Valid. `resolveDriver()` now re-throws when the native driver was explicitly forced, so tests can't silently run against the wrong backend; the automatic version-based fallback is unchanged.

PR #1261, opened 2026-07-17
fix(db): recover WASM SQLite locks via PID sentinel, not a 60s age window

BYK:
## Summary

Follow-up to #1260 addressing a post-merge finding from **Cursor Bugbot** and **sentry-warden**.

On the WASM SQLite fallback (npm package on Node < 22.15), `node-sqlite3-wasm` locks the DB with a `<db>.lock` directory (mkdir mutex) and releases it via `rmdir` on close. A CLI process killed by **SIGINT (Ctrl+C) / SIGTERM** mid-write bypasses the `process.on('exit')` handler, so the lock dir leaks. The previous recovery only cleared a leftover lock once it aged past a **60-second** window — so a quick re-run after Ctrl+C failed with `database is locked` for up to a minute.

## Fix

PID-owner sentinel for **precise** liveness detection instead of a time guess:

- On open (WASM), write `<db>.lock.owner` = our PID.
- On the next open, `clearStaleWasmLock` reads it:
  - owner **dead** → lock is provably orphaned → cleared **immediately** (no wait),
  - owner **alive** → genuine live lock → left to `busy_timeout` (no stealing),
  - **no sentinel** → fall back to the age window (older CLI / races).
- On clean close, remove the sentinel.

The sentinel lives **outside** the lock dir — a file *inside* it would make the driver's `rmdir` unlock fail (verified), defeating the driver's own cleanup.

### Why not a SIGINT/SIGTERM handler?
A DB-layer signal handler would risk interfering with commands that legitimately own SIGINT for graceful streaming shutdown (`log list --follow`, `dashboard view`), which keep the process alive after the signal. The recovery-side fix has zero interference risk.

## Verification
- New tests: dead-owner lock cleared immediately despite fresh mtime; live-owner (current PID) lock preserved despite old mtime. 28 sqlite-adapter tests, 296 db tests pass.
- E2E reproduction: a fresh lock dir with a dead-owner sentinel is cleared on the next run in ~startup time (no 60s stall); previously this exact case blocked for 60s.
- SEA binary still excludes the WASM driver (0 driver symbols); native path unchanged.
- npm real-install (no `node_modules`) smoke passes.

Also corrected the now-inaccurate JSDoc that described the (previously wrong) 'live lock is recently re-touched' premise.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -483,9 +483,6 @@
 <!-- lore:019f666a-ac76-7581-82f8-8d2af3f3e428 -->
 * **Always perform rigorous, read-only code reviews with specific focus areas**: The user consistently conducts critical, adversarial code reviews in a read-only manner, focusing on specific technical areas. They provide detailed context about the repository, branch, and changes under review, and specify exact focus areas like path correctness, dependency management, build configurations, and security implications. The user maintains strict read-only discipline, prohibiting any modifications to source files, and often works within isolated worktrees. They expect thorough verification of technical changes against documented requirements and prior decisions.
 
-<!-- lore:019e6134-4be2-75d2-8eb3-a69dd468428d -->
-* **Always plan systemic fixes with structured multi-problem breakdowns before implementation**: When the user identifies documentation or tooling issues, they consistently organize them as numbered problems with precise file locations, line numbers, and root causes before any code is written. They expect the assistant to engage at the planning level first — proposing detection strategies, fix approaches, and tradeoffs — and to consolidate related problems (e.g., merging overlapping tasks) rather than treating each in isolation. Plans are written to files and iterated on. Implementation only follows after the plan is agreed upon. The user prefers systemic/automated fixes (e.g., derive patterns from package.json) over one-off patches.
-
 <!-- lore:019f65d7-0856-76f1-8ef3-252b56b66279 -->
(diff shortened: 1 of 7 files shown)


BYK:
Thanks — addressed all four Cursor Bugbot findings in a62280ae1:

- **Sentinel races under concurrent opens (High)** and **Dead sentinel steals live locks (Medium):** These share a root cause — a stale/overwritten single sentinel could justify deleting a lock another process just acquired (the corruption path). Fixed by adding `LOCK_SAFETY_FLOOR_MS` (3s): **no lock is ever removed while younger than the floor**, regardless of the sentinel. So a freshly-acquired live lock can never be stolen by a leftover dead-PID sentinel; past the floor a dead owner clears the orphan immediately, a live owner keeps it, and concurrent races degrade to the age floor rather than corruption. Added a test proving a fresh lock + dead sentinel is preserved.
- **Duplicate liveness helper (Low):** Extracted `isProcessRunning` into a new dependency-free `src/lib/process-utils.ts`; `binary.ts` re-exports it (existing importers unaffected) and `db/sqlite.ts` imports from there — one implementation, and no dragging `binary.ts`'s heavy graph onto the DB hot path.
- **PID reuse blocks recovery (Low):** Bounded — with no sentinel the age fallback still fires; a reused-PID false positive at worst defers cleanup by one run, never corrupts. Acceptable given the alternative (stealing a possibly-live lock) is strictly worse.

Verified: fresh lock + dead sentinel is kept (no steal); dead-owner lock past the 3s floor recovers in ~1s instead of 60s; SEA binary still excludes the WASM driver; 29 sqlite + 419 db/upgrade tests pass.

PR #1311, opened 2026-07-28
fix(logger): render log timestamps in local time instead of UTC

MathurAditya724:
Log timestamps (the `[auth.login 6:28:59 AM]` prefix) and other Date output were rendered in UTC instead of the user's local time. The CLI ships as Node SEA binaries that often can't resolve the OS timezone and silently fall back to UTC, so a user in US/Pacific saw timestamps hours off from their clock even with correct OS settings.

The fix detects the OS timezone at startup and sets `process.env.TZ` when the runtime fell back to UTC, and overrides consola's reporter `formatDate` to derive local time from `getTimezoneOffset()` in an unambiguous 24-hour format (robust even if ICU state is stale).

## Testing
- `vitest run test/lib/timezone.test.ts test/lib/logger.test.ts test/lib/formatters` — 975 passing (14 new timezone tests).
- Manual: with `TZ` unset in a UTC-fallback runtime, verified the log prefix and `toLocaleTimeString()` render correct local time once repaired; confirmed `initTimezone()` is a no-op when `TZ` is set or the machine is genuinely UTC.
- `biome check` clean on changed files; `tsc` clean on changed files (pre-existing generated-code errors unrelated).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -12,6 +12,7 @@
 
 import { getEnv } from "./lib/env.js";
 import { CliError } from "./lib/errors.js";
+import { initTimezone } from "./lib/timezone.js";
 
 /**
  * Preload project context: walk up from `cwd` once, finding both the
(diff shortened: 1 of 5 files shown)


BYK:
Jared, let's get the patch coverage above 80%. Also let's verify that the [`binpunch`](https://github.com/BYK/binpunch) run we do in fossilize does not interfere with this?

PR #1314, opened 2026-07-29
fix(local): Point `trace view --web` at the trace detail route

sergical:
## What was broken

Three commands printed or opened links that don't work:

- `sentry trace view --web <trace-id>` — opens the traces **list** instead of the trace you asked for
- `sentry trace logs --web <trace-id>` — same
- `sentry log view <log-id>` — the `Link` row it prints is a dead link

In every case you end up on a list page with your trace ID silently dropped.

## Why

We build the URL as:

```
https://my-org.sentry.io/traces/{traceId}/
```

That path doesn't exist. In Sentry's router the trace page lives at `trace/:traceSlug/`, mounted under both `traces/` and `explore/traces/`. So the real URL needs a `trace/` segment in the middle:

```
https://my-org.sentry.io/explore/traces/trace/{traceId}/
```

Without it the URL matches the *list* route instead, and the ID is thrown away. That's why you land on the list.

This is just old code. The other builders in the same file were already moved to the new `explore/` paths (`buildLogsUrl` → `/explore/logs/`, `buildReplayUrl` → `/explore/replays/{id}/`). Traces got missed.

## The fix

1. Build the correct URL — add the missing `trace/` segment and the `explore/` prefix.
2. Teach the URL *parser* about the new path, so pasting a trace URL into the CLI still works.
3. Keep parsing the old `traces/{id}/` shape too, so links people already copied keep working.

Tests updated for the new URL, plus cases covering the old one so the fallback stays working.

### DIFF
--- a/src/lib/sentry-url-parser.ts
+++ b/src/lib/sentry-url-parser.ts
@@ -129,6 +129,11 @@ function matchSubdomainPath(
   if (tracePath.status === "detail") {
     return { traceId: tracePath.traceId };
   }
+  if (tracePath.status === "list") {
+    // A bare trace list URL (e.g. `/explore/traces/`) resolves to the org,
+    // matching the replay-list behavior below.
+    return {};
+  }
 
   const replayPath = matchReplayPath(segments, 0);
   if (replayPath.status === "detail") {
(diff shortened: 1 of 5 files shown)


BYK:
Jared,

>one non-blocking nit: the parser file's top-level JSDoc (the "Recognizes these path patterns" list around lines 291/300) still only mentions the legacy traces/{traceId}/ form — worth adding the new canonical explore/traces/trace/{traceId}/ there for completeness. not a blocker.

Push a commit fixing this please.

Also, check the PR against `.lore.md` file for repo conventions. If all checks out and no Seer or BugBot review comments, merge it away. Don't forget to thank @sergical!

PR #1316, opened 2026-07-30
fix(replays): Use `events` data source for issue replay count

mjq:
The `discover` dataset is deprecated. Searches by `issue.id` like this one were routed to the `events` (errors) dataset behind the scenes already; query it directly. ([Proof this is a valid option](https://github.com/getsentry/sentry/blob/5938102899e45752c97ea2081b5726d31e3fa065/src/sentry/replays/endpoints/organization_replay_count.py#L41), and that [there is no transactions equivalent for issue.id](https://github.com/getsentry/sentry/blob/5938102899e45752c97ea2081b5726d31e3fa065/src/sentry/snuba/events.py#L32-L39), so you'd route to `events` anyway 😄 ).

### DIFF
--- a/src/lib/api/replays.ts
+++ b/src/lib/api/replays.ts
@@ -338,7 +338,7 @@ export async function listReplayIdsForIssue(
     `/organizations/${orgSlug}/replay-count/`,
     {
       params: {
-        data_source: "discover",
+        data_source: "events",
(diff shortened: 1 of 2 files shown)


nikkikapadia (approved):
noice

PR #1328, opened 2026-07-31
fix(logs): Correct severity query examples

jamieQ:
## Summary

- replace misleading `level:error` examples with `severity:error` for Logs queries
- update log and trace command help, documentation fragments, and generated skill references
- align log-specific forwarding tests with the supported field

## Why

The Logs API uses `severity` for the canonical log severity. `level` is treated as an independent custom attribute, not an alias, so the existing examples commonly return no matches.


BYK (approved):
Thanks!

PR #1351, opened 2026-08-03
feat(local): inject Spotlight as a Wrangler Worker binding

MathurAditya724:
Wrangler does not expose inherited process environment variables to Worker code, so the SENTRY_SPOTLIGHT value set by local run never reached @sentry/cloudflare. Users had to hard-code spotlight: true in their SDK config.

Detect Wrangler dev commands in Workers projects and add the ephemeral --var SENTRY_SPOTLIGHT:<url> binding. This works for direct/npx commands, package-manager scripts, auto-detected shell scripts, custom config paths and verify mode; explicit user bindings remain untouched. No wrangler config or .dev.vars files are modified.

### DIFF
--- a/packages/cli/plugins/sentry-cli/skills/sentry-cli/references/local.md
+++ b/packages/cli/plugins/sentry-cli/skills/sentry-cli/references/local.md
@@ -52,8 +52,6 @@ sentry local -f error -f log
 # Run quietly (suppress per-envelope tail output)
 sentry local --quiet
 
---var SENTRY_SPOTLIGHT:http://localhost:<port>/stream
-
 sentry local -f error -f log    # only errors and logs
 
 sentry local -f ai          # only AI/agent spans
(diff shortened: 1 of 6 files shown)


packages/cli/plugins/sentry-cli/skills/sentry-cli/references/local.md
```diff
@@ -52,6 +52,8 @@ sentry local -f error -f log
 # Run quietly (suppress per-envelope tail output)
 sentry local --quiet
 
+--var SENTRY_SPOTLIGHT:http://localhost:<port>/stream
```
BYK: Sorry, what's this?

BYK (commented):
Does this not work: https://developers.cloudflare.com/workers/configuration/environment-variables/#import-env-for-global-access

(we should do this on SDK side)

packages/cli/plugins/sentry-cli/skills/sentry-cli/references/local.md
MathurAditya724: for cloudflare we can't define `SENTRY_SPOTLIGHT` like a global env, cause they don't read that. They have their own scoped env thing inside wrangler

PR #1368, opened 2026-08-05
fix(explore): Remove deprecated `discover` and `transactions` datasets

mjq:
These are now deprecated for all customers, and will be removed as soon as it is safe to do so. Remove them from the `explore` command. Replacements are available (`errors` and `spans`, respectively).

The only docs for this command appear to be autogenerated, so no other changes needed. (Skills still advertise these datasets when talking to the API directly, but I'll address that in a future change.)

See BROWSE-682.


BYK:
Jared, get this merged if the changes look good.

PR #1373, opened 2026-08-06
fix(auth): request team admin OAuth scope

betegon:
## Summary

Add `team:admin` to the CLI standard OAuth scope set so authenticated Team Admins can create projects through team-scoped API endpoints. Regenerate the documentation projections owned by the OAuth scope generator.

Existing stored OAuth grants may predate this scope. When an interactive command receives a scope-specific 403, the CLI now offers one refresh with the current standard scopes and retries exactly once. It does not start OAuth for `--yes`, `--dry-run`, JSON/non-interactive execution, or effective environment-token auth. Organization policy denials are excluded from scope recovery.

This expands what the token may do; it does not grant Team Admin status to users who do not already have that role.

## Test plan

- `pnpm exec vitest run test/lib/oauth.test.ts test/lib/api-scope.test.ts test/lib/scope-recovery.test.ts test/lib/auto-auth.test.ts --coverage=false`
- `pnpm exec biome check src/cli.ts src/lib/api-scope.ts src/lib/scope-recovery.ts src/lib/oauth.ts test/lib/api-scope.test.ts test/lib/scope-recovery.test.ts test/lib/oauth.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm run check:docs-sections`


BYK (approved):
We should make sure to refresh with the new scope on failure for existing tokens.

betegon:
Thanks for the heads up! I'll check it out

betegon:
Implemented the existing-grant path from the review: on a scope-specific 403 from the effective stored OAuth credential, the CLI offers a one-time refresh with the current standard scope set and retries once. Unattended/non-interactive/JSON runs and env-token auth never enter the device flow, and policy denials are not treated as missing OAuth scopes.

PR #1418, opened 2026-08-13
fix(init): detect existing installations safely

betegon:
## Summary

The init preflight now determines existing Sentry state from evidence owned by the selected app directory instead of ambient process state. It reports partial as well as complete installations, ignores build-only tooling on its own, recognizes runtime SDKs and initialization across the supported ecosystems, and returns the scanned cwd so the server can verify that cached detection is scoped to the target.

This also makes non-interactive monorepo selection accept an exact app name, full path, or unique path basename. Exact matches take precedence, while ambiguous basename aliases fail with the matching apps instead of selecting the wrong target.

## Rollout compatibility

This is the CLI side of [getsentry/cli-init-api#216](https://github.com/getsentry/cli-init-api/pull/216). That server change treats both partial and installed state as a safety bail, trusts precomputed state only when its cwd matches the selected app, and falls back to its own scoped detection for older clients. Existing servers can ignore the additive cwd field, so the CLI and server changes do not need an atomic rollout.

## Test plan

- [x] 59 targeted interactive, filesystem-tool, and detect-sentry tests
- [x] pnpm typecheck
- [x] pnpm lint
- [x] pnpm check:deps
- [x] pnpm check:errors
- [x] pnpm check:patches
- [x] pnpm check:stale-refs
- [x] pnpm check:fragments
- [x] pnpm check:docs-sections
- [x] git diff --check

### DIFF
--- a/packages/cli/src/lib/init/interactive.ts
+++ b/packages/cli/src/lib/init/interactive.ts
@@ -50,6 +50,46 @@ const DEFAULT_FEATURE_RANK = new Map<string, number>(
 const UNSUPPORTED_INIT_FEATURES = new Set(["userFeedback"]);
 const FEATURE_SELECTION_CONTEXT =
   "Based on your project, these features are available to set up.";
+const TRAILING_SLASHES_RE = /\/+$/;
+
+function appFlagAliases(app: AppEntry): string[] {
+  const normalizedPath = app.path
+    .replaceAll("\\", "/")
+    .replace(TRAILING_SLASHES_RE, "");
+  const pathBasename = normalizedPath.split("/").at(-1);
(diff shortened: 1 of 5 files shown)


betegon:
Closing this draft because it combined two independent concerns and overlaps the dedicated improve-existing setup flow already underway.

Existing-install detection and safe reruns remain tracked in getsentry/cli-init-api#82 and getsentry/cli-init-api#83. The independent app-alias behavior has been split into #1419. A bail-only CLI patch would conflict with the intended improve-existing UX.

The branch fix/init-detect-existing-install is being kept intact for reference or selective cherry-picking; no code is being deleted.

PR #1427, opened 2026-08-14
feat(cli): add completion command

MunifTanjim:
## Why?

I like to manage my own shell setup, so I do this:

```sh
sentry cli setup --method brew --no-modify-path --no-completions --no-agent-skills
```

But currently there's no way to get the completion script without letting `sentry cli` put it on the filesystem for me.

Adding `sentry cli completion` command solves that. It is a common pattern for most popular CLIs.

## Summary

Adds `sentry cli completion <shell>` to print the shell completion script to stdout for **bash**, **zsh**, and **fish**. When no shell argument is given, the shell is auto-detected from `$SHELL`.

This complements `sentry cli setup` (which writes completion files to disk) by letting users and package managers install completions however they like:

```sh
sentry cli completion zsh > ~/.zfunc/_sentry
eval "$(sentry cli completion bash)"
```

## Details

- New command in `packages/cli/src/commands/cli/completion.ts`, registered under the existing `cli` route group next to `setup`/`uninstall`.
- Reuses the existing `getCompletionScript()` generator (`lib/completions.ts`) and `detectShellType()` helper (`lib/shell.ts`) — no new generation logic. Output is byte-identical to what `cli setup` writes.
- `auth: false`; unsupported shells fail with a `ValidationError` listing the supported shells.
- Regenerated skill/command docs (`cli.md`, `SKILL.md`, `contributing.md`) via `pnpm run generate:docs`.

## Testing

- `packages/cli/test/commands/cli/completion.test.ts` — covers bash/zsh/fish output, the unsupported-shell error, and `$SHELL` auto-detection. All pass.
- `tsc --noEmit` and `biome check` clean.
- Manual smoke test: each shell prints its script, `nonsense` errors with a non-zero exit, and `sentry cli --help` lists the command.

### DIFF
--- a/apps/cli-docs/src/content/docs/contributing.md
+++ b/apps/cli-docs/src/content/docs/contributing.md
@@ -55,7 +55,7 @@ cli/
 │   │   ├── alert/       # create, delete, edit, list, view
 │   │   ├── auth/        # login, logout, refresh, status, token, whoami
 │   │   ├── build/       # download, upload
-│   │   ├── cli/         # defaults, feedback, fix, import, setup, uninstall, upgrade
+│   │   ├── cli/         # completion, defaults, feedback, fix, import, setup, uninstall, upgrade
 │   │   ├── code-mappings/# upload
 │   │   ├── conversation/# list, view
 │   │   ├── dart-symbol-map/# upload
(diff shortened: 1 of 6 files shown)


BYK (approved):

BYK:
@MunifTanjim this is great and the PR quality is top notch. Thanks a lot!

PR #1465, opened 2026-08-25
feat(sdk): Add headers option to createSentrySDK

JPeer264:
Consumers of the programmatic SDK, such as the Sentry bundler plugins, need to send extra HTTP headers to self-hosted instances behind a reverse proxy (Google IAP, Cloudflare Access). Until now the only way was to set `SENTRY_CUSTOM_HEADERS` on `process.env` before invoking a command, which is process-global state.

`SentryOptions.headers` maps onto `SENTRY_CUSTOM_HEADERS` in the isolated env of each invocation, so the existing self-hosted guard, forbidden-header list, and origin trust checks all apply unchanged. Values containing a separator are rejected up front, since they would otherwise parse into a bogus extra header.

### DIFF
--- a/packages/cli/src/lib/custom-headers.ts
+++ b/packages/cli/src/lib/custom-headers.ts
@@ -127,6 +127,34 @@ export function parseCustomHeaders(raw: string): readonly [string, string][] {
 }
 
 /**
+ * Serialize a header map into the `SENTRY_CUSTOM_HEADERS` string format.
+ *
+ * Inverse of {@link parseCustomHeaders}: the result round-trips through it.
+ * A value containing a separator (`;` or a line break) would be split into a
+ * bogus extra header on parse, so it is rejected up front.
+ *
+ * @param headers - Header name/value map (e.g., from `SentryOptions.headers`)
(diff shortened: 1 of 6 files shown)


packages/cli/src/lib/custom-headers.ts
```diff
@@ -126,6 +126,34 @@ export function parseCustomHeaders(raw: string): readonly [string, string][] {
   return results;
 }
 
+/**
+ * Serialize a header map into the `SENTRY_CUSTOM_HEADERS` string format.
+ *
+ * Inverse of {@link parseCustomHeaders}: the result round-trips through it.
```
BYK: Can we not avoid the round-trip and unnecessary string serialize/deserialize waste?

BYK (approved):
I'd love to avoid the unnecessary serialize->deserialize dance but other than that looks pretty good, thanks!

packages/cli/src/lib/custom-headers.ts
JPeer264: Waste has been reduced and dancing got prohibited. It's now as clean as it can get

BYK (approved):

PR #1496, opened 2026-08-27
fix(sourcemap): adopt a debug ID already present on the sourcemap

msonnb:
## Summary

`sentry sourcemap inject` and `sentry sourcemap upload` only looked for an existing debug ID in the **JS file** (`//# debugId=`). If the JS had no such comment but the **sourcemap** already carried `debug_id`/`debugId`, the CLI minted a brand-new content-derived ID, rewrote the JS, and overwrote the map's field.

## Why it matters

Sentry's JavaScript bundler plugins support `sourcemaps.disable: 'disable-upload'`: the plugin injects a runtime `_sentryDebugIds` snippet into the bundle during the build and stamps that same ID into the emitted `.js.map` (getsentry/sentry-javascript#23619). It deliberately does **not** append `//# debugId=` to the bundle, because rewriting the bundle after emit invalidates subresource-integrity hashes computed during the build (getsentry/sentry-javascript-bundler-plugins#949 — the reporter serves `<script integrity="…">` from `webpack-assets-manifest`).

Those users then run `sentry sourcemap upload ./dist`. Before this change the CLI minted a different ID, so the ID the SDK reported at runtime never matched what Sentry indexed and nothing symbolicated — plus the rewrite broke the SRI hashes they were protecting.

## What changed

Debug ID precedence when resolving a file pair:

1. `//# debugId=` in the JS → use it (unchanged; it's the on-disk spec marker)
2. **New:** otherwise, a valid `debug_id`/`debugId` on the sourcemap → adopt it, `wasInjected: false`, **neither file is written**
3. Otherwise, mint from content (unchanged)

`debug_id` wins over `debugId`. A value that isn't a well-formed UUID is treated as absent and falls through to minting.

Case 2 deliberately does **not** prepend the IIFE snippet and does **not** offset `mappings`: the bundle already has the plugin's own `_sentryDebugIds` writer (a second one under a different stack key makes the runtime mapping ambiguous), and the plugin's map already lines up with the un-offset bundle.

Covers external maps, inline (base64 `data:`) maps, `--dry-run`, and the discovery read behind `sourcemap resolve`. This is default behavior — not gated behind a new flag, and `--no-rewrite` is unchanged.

- `src/lib/sourcemap/debug-id.ts` — new `readSourcemapDebugId()`; early return in `injectDebugId()` and `injectInlineDebugId()`
- `src/lib/sourcemap/inject.ts` — new `readMapDebugId()` helper wired into the `--dry-run` branch and `resolveDirectorySourcemaps()`

The upload path needed no change — `buildArtifactPair()` already puts the resolved `debugId` on both entries — but it's covered end to end by new tests rather than assumed.

## On v3 parity

Worth recording, since the flag list in the migration guide reads as a capability loss:

- **`--debug-id-reference` was an `upload` flag**, and narrower than it sounds. Its help text: *"By default Debug ID reference has to be present both in the source and the related sourcemap. But in cases of binary bundles, the tool can't verify presence of the Debug ID. This flag allows use of Debug ID from the linked sourcemap."* It's a verification relaxation for binary bundles (Hermes bytecode, where a `//# debugId=` comment is impossible) — it injects nothing. This is why it shows up in the React Native docs right after `copy-debugid.js`.
- **v3's `sourcemaps inject` already adopted the map's ID by default**, no flag — but it still rewrote both files. The `debug_id_fresh` boolean in `sourcemaps.rs` only selects a report bucket; `fixup_js_file()` still injects the snippet and shifts the mappings either way. So `Ignored: The following sourcemap files already have debug ids` meant *"no new ID minted"*, not *"file untouched"*.

So this PR matches v3 on **which ID wins** and is intentionally stricter on **what gets written** — leaving both files byte-identical is precisely what makes the integrity-hash workflow work.

**Known gap, flagged for review:** a map carrying a `debug_id` whose bundle has *neither* a comment *nor* a plugin snippet now gets no runtime registration at all (v3 would have injected one). It doesn't affect the two motivating workflows — Hermes can't be injected anyway, and `disable-upload` already has the snippet — but if we want v3 parity there, the fix would be to detect `_sentryDebugIdIdentifier` in the bundle and fall back to injecting when absent. Left out because sniffing "does this bundle self-register?" from a substring is fragile.

## Tests

`test/lib/sourcemap/inject.test.ts` — 8 new cases: external `debug_id`; the `debugId` spelling; precedence between the two; inline map; JS comment wins over a conflicting map field; malformed value falls through to minting; `--dry-run` reports the adopted ID; repeat runs stay a no-op. Each asserts the JS and map are byte-identical.

`test/commands/sourcemap/upload.test.ts` — 2 new cases asserting uploaded artifacts carry the adopted ID on both the `minified_source` and `source_map` entries, for external and inline maps.

## Docs

Added a bullet to `migrating-from-v3.md` noting the capability is back as automatic default behavior, with the accurate v3 semantics.

## Verification

- `pnpm lint` — clean
- `pnpm typecheck` — clean
- `pnpm test:unit` — 9334 passed. The 6 failures in `test/lib/time-range.test.ts` are pre-existing and unrelated (timezone-dependent assertions); confirmed identical on a clean tree via `git stash`.

### DIFF
--- a/packages/cli/src/lib/sourcemap/debug-id.ts
+++ b/packages/cli/src/lib/sourcemap/debug-id.ts
@@ -35,6 +35,33 @@ const DEBUGID_COMMENT_PREFIX = "//# debugId=";
 /** Regex to extract an existing debug ID from a JS file. @internal */
 export const EXISTING_DEBUGID_RE = /\/\/# debugId=([0-9a-fA-F-]{36})/;
 
+/** UUID format: 8-4-4-4-12 hex with hyphens. */
+const UUID_RE =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/**
+ * Read a pre-existing debug ID off a parsed sourcemap.å
+ *
+ * `debug_id` wins over `debugId` when both are present. A value that is not a
+ * well-formed UUID is treated as absent so the caller falls through to minting.
+ *
+ * @param map - A parsed sourcemap object (any shape; fields are probed)
+ * @returns The debug ID, or `undefined` when the map carries none
+ * @internal
+ */
+export function readSourcemapDebugId(map: unknown): string | undefined {
+  if (typeof map !== "object" || map === null) {
+    return;
+  }
+  const { debug_id: snake, debugId: camel } = map as SourcemapJson;
+  for (const candidate of [snake, camel]) {
+    if (typeof candidate === "string" && UUID_RE.test(candidate)) {
+      return candidate;
+    }
+  }
+  return;
+}
+
 /**
  * Generate a deterministic debug ID (UUID v4 format) from content.
  *
@@ -118,7 +145,9 @@ export function prependDebugIdSnippet(
  * registered in source code (`constants.ts`) instead of via the IIFE.
  *
  * The operation is **idempotent** — files that already contain a
- * `//# debugId=` comment are returned unchanged.
+ * `//# debugId=` comment are returned unchanged. A debug ID already present
+ * on the sourcemap is likewise adopted as-is, leaving both files untouched
+ * (see {@link readSourcemapDebugId}).
  *
  * @param jsPath - Path to the JavaScript file
  * @param mapPath - Path to the companion `.map` file
@@ -140,6 +169,19 @@ export async function injectDebugId(
   const existingMatch = jsContent.match(EXISTING_DEBUGID_RE);
   if (existingMatch?.[1]) {
     return { debugId: existingMatch[1], wasInjected: false };
+  }
+
+  const map = JSON.parse(mapContent) as SourcemapJson;
+
+  // The JS carries no comment, but the map may already have been stamped by a
+  // bundler plugin that intentionally left the bundle alone. Adopt that ID and
+  // touch neither file: the bundle already registers it via the plugin's own
+  // `_sentryDebugIds` writer (a second snippet under a different stack key
+  // would make the runtime mapping ambiguous), and the map's `mappings` line
+  // up with the un-offset bundle.
+  const mapDebugId = readSourcemapDebugId(map);
+  if (mapDebugId) {
+    return { debugId: mapDebugId, wasInjected: false };
   }
 
   // Derive the debug ID from the minified JS content combined with the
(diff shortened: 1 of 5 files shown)


packages/cli/src/lib/sourcemap/debug-id.ts
```diff
@@ -35,6 +35,33 @@ const DEBUGID_COMMENT_PREFIX = "//# debugId=";
 /** Regex to extract an existing debug ID from a JS file. @internal */
 export const EXISTING_DEBUGID_RE = /\/\/# debugId=([0-9a-fA-F-]{36})/;
 
+/** UUID format: 8-4-4-4-12 hex with hyphens. */
+const UUID_RE =
```
BYK: Don't we already have this regular expression somewhere else??

packages/cli/src/lib/sourcemap/inject.ts
```diff
@@ -139,6 +141,29 @@ export async function injectDirectory(
...
+async function readMapDebugId(map: MapSource): Promise<string | undefined> {
+  if (map.kind === "inline") {
+    return readSourcemapDebugId(map.decoded.map);
+  }
+  try {
+    return readSourcemapDebugId(
+      JSON.parse(await readFile(map.mapPath, "utf-8"))
+    );
+  } catch (err) {
+    log.debug(`failed to read debug ID from sourcemap: ${map.mapPath}`, err);
```
BYK: Should this be reported to Sentry upstream?

packages/cli/test/commands/sourcemap/upload.test.ts
```diff
@@ -4,7 +4,13 @@
  * branches in `buildEmptyDiscoveryError`.
  */
 
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import {
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
```
BYK: Let's use the async versions from `node:fs/promises` for all these

packages/cli/src/lib/sourcemap/debug-id.ts
msonnb: yes, fixed.

packages/cli/src/lib/sourcemap/inject.ts
msonnb: I raised it to `log.warn` like in `findCompanionMap` because it's ultimately something a user needs act on and not a CLI issue. Let me know if you prefer to emit telemetry.

packages/cli/test/commands/sourcemap/upload.test.ts
msonnb: done 👍

BYK (approved):

PR #1499, opened 2026-08-28
feat(migrate): Add `migrate` command and JS v10 -> v11 migration

msonnb:
## `sentry migrate` — upgrade a project across a Sentry SDK major

Upgrading across an SDK major currently means reading the migration guide and finding every renamed option, moved import and removed export by hand. This adds a command that does the mechanical part and hands you a list of what's left.

```
sentry migrate            # apply
sentry migrate --dry-run  # preview
```

- Detects which migration applies by reading the project. One ships today, JavaScript SDK 10.x → 11.x.
- Applies every breaking change that can be made deterministically.
- Marks call sites needing a human decision with `TODO`.
- Writes `*_MIGRATION.md` listing what's left, with a `file:line` per item. Hand that file to a coding agent and it only works on what the codemod already proved it can't do.
- Refuses to run outside git or on a dirty tree, since `git checkout` is the undo. `--allow-dirty` overrides.
- Rewrites `package.json` but installs nothing, so no lockfile churn buries the diff. Safe to re-run.

The engine is generic and the JS v11 knowledge is a plugin under `migrations/sentry-javascript-v11/`. A migration is a list of small tasks, each of which decides for itself whether it applies by looking for the thing it knows how to change. Nothing consults a catalog of what's relevant. A task that finds nothing writes nothing, one failing task can't take down the run, and `--only` / `--skip` address any of them by id.


HazAT:
This looks great for specialized migrations. One thing I’d suggest adding is a generic fallback for upgrades without a dedicated migration: detect the SDK/version, read the relevant changelog or migration guide, and surface the important breaking changes for an agent to act on. Agents should still be able to pick this stuff up when we haven’t built a bespoke codemod, especially changes that could surprise users.

BYK (changes requested):
Preemptively blocking. 

I like the idea but I don't like the specific implementation:

CLI cannot be a dumping ground for all SDKs all migrations. All it can have can be a harness, or a run-time etc. The actual logic should live in the respective SDK's repo and should be dynamically imported as we also cannot afford shipping all this code at all times. 

Moreover, I think this should use the same smart backend with `sentry init`. I think @betegon would have better ideas how to achieve that.

Issue #759, opened 2026-04-15
Allow adding custom header to all requests


kbrooks:
At my company, we have a self-hosted sentry instance behind an IAP authentication layer, which makes it impossible to use the Sentry CLI as it exists today. On the old sentry CLI we did this:

```
export SENTRY_PROJECT="your-project-name"
export CUSTOM_HEADER="SC-LCA-1: ..."
```

Support for this behavior was added in this PR https://github.com/getsentry/sentry-cli/pull/1077

This doesn't seem to be supported on the new CLI which doesn't support a custom header or a header parameter on the `sentry auth login` command 

```
sentry auth login  -H "SC-LCA-1: ..."
```

and the plain command fails with 
```
 ERROR  Failed to initiate device flow                                                                                                                                                          auth.login 1:01:31 PM
  Endpoint: /oauth/device/code/
  Invalid IAP credentials: empty token
```

Can we add support for custom headers to the new cli?

BYK:
@kbrooks thanks for the issue! Looking into how we can get this done.

BYK:
@kbrooks LMK if the design in #761 would work for you and I'll merge it

kbrooks:
This will work for us. Thanks for the quick fix

PR #1500, opened 2026-08-28
fix(sdk): Derive the published SentryOptions type from its source

JPeer264:
In #759 I added the headers and missed that there is also a `bundle.ts` that has stringified types in it. With that we 1. add the missing `headers` type and 2. derive the types directly from `src/lib/sdk-types.ts` - not sure if you're ok with the latter. I could also trim down the PR to just add the `headers` type as string


BYK (approved):

BYK:
Jared, take over the PR, get it ready for merging, fix all your and other bots' findings and then merge when PR is green.

PR #1501, opened 2026-08-28
feat(doctor): add sentry doctor

romtsn:
## Summary

Adds `sentry doctor`: a read-only health check of an existing Sentry install. It scans the project, asks the API, and reports what is configured, what is broken, and what to fix.

- Capture init/build sites from a marker table, then run tier-1 (server) and tier-2 (local) checks. `--send-test-event` does a live round-trip. `--fix` is a dry-run setup plan, never a write.
- JSON report omits judged keys and `cwd`. Interactive runs can send the report to support on an explicit yes, even with telemetry off.
- Live-tested against Android, Empower, and iOS. Known skip: no `project.pbxproj` parsing (Xcode run scripts stay invisible; Fastfile only).

## Test plan

- `pnpm --filter sentry exec vitest run test/lib/doctor test/commands/doctor.test.ts test/lib/dsn/code-scanner.test.ts` — 176 tests
- `pnpm --filter sentry exec tsc --noEmit -p tsconfig.json`
- Live: `sentry doctor --json` on `~/Workspace/android`, `~/Workspace/empower`, `~/Workspace/ios`


packages/cli/plugins/sentry-cli/skills/sentry-cli/references/doctor.md
```diff
@@ -0,0 +1,35 @@
...
+# Doctor Commands
+
+Check whether Sentry is correctly set up and actually working
+
+### `sentry doctor`
+
+Check whether Sentry is correctly set up and actually working
+
+**Flags:**
+- `--sendTestEvent - Send a synthetic event to the configured DSN and confirm it arrives (a write)`
```
BYK, line 19: I'd rename this to `online` or something to make it not sound like a subcommand.

packages/cli/src/lib/doctor/checks/index.ts
```diff
@@ -0,0 +1,7 @@
+/** The ordered check registry. Order here is report order. */
+
+import type { Check } from "../types.js";
+import { TIER1_CHECKS } from "./tier1.js";
+import { TIER2_CHECKS } from "./tier2.js";
```
BYK, line 5: These tiers need explanation. Why do we have 2 tiers? What do they stand for?

packages/cli/src/lib/doctor/checks/tier2.ts
```diff
@@ -0,0 +1,364 @@
...
+import type { Capture, CapturedBlock, Check, CheckResult } from "../types.js";
+
+/** Kinds produced by `autoInit` marker rules — config, not a code call. */
+const AUTO_INIT_KINDS = new Set(
+  INIT_MARKERS.filter((rule) => rule.autoInit).map((rule) => rule.kind)
+);
+
+/** Ecosystems that use a bundler/build plugin to upload symbolication data. */
+const UPLOAD_EXPECTING_ECOSYSTEMS = new Set([
+  "javascript",
```
BYK, line 20: Ideally this list would come from a central place where *all* of our products can read from. Possibly the release registry?

packages/cli/src/lib/doctor/capture-block.ts
```diff
@@ -0,0 +1,313 @@
+/**
+ * One block scanner for every platform doctor understands.
```
BYK, line 2: This concept is a big no. We should not be rolling and maintaining our bonafide half-assed parser. We should be using something like treesitter or ast-grep or something.

packages/cli/src/lib/doctor/capture.ts
```diff
@@ -0,0 +1,303 @@
...
+const DEFAULT_TIME_BUDGET_MS = 1500;
+const DEFAULT_MAX_FILES = 200;
+const MAX_GREP_RESULTS = 5000;
+const MAX_FILE_BYTES = 512 * 1024;
+
+/** Broad enough to catch every marker table entry in a single pass. */
+const SENTRY_PATTERN = /sentry/i;
+
+/** Auto-init config files, even when gitignored. Not every xml/json. */
+const CONFIG_FILE_GLOBS = [
```
BYK, line 59: I don't understand why we are doing this. If this is to find DSNs, our existing DSN discovery code should already cover this. If not, that code should be updated, not built upon here so the rest of the CLI can benefit.

packages/cli/src/lib/doctor/fix.ts
```diff
@@ -0,0 +1,86 @@
+/**
+ * `--fix`: escalate from diagnosis to the setup workflow's plan.
```
BYK, line 2: I think this entire fix system should be agent-assisted. So we should generate a report and either dump that for agents use, or use our existing `sentry init` backend for this. @betegon would be able to advice there.

packages/cli/src/lib/doctor/judge.ts
```diff
@@ -0,0 +1,209 @@
...
+  capture: Capture,
+  opts: JudgeOptions = {}
+): Promise<CheckResult[]> {
+  // Path 1 — an agent is reading this. It is better at the question than a
+  // one-shot classifier, and it costs nothing.
+  if (detectAgent() !== undefined) {
+    return [agentHandoff(capture)];
+  }
+
+  const apiKey = opts.apiKey ?? process.env.ANTHROPIC_API_KEY;
```
BYK, line 123: Okay, this looks more like what I think we should be doing as I mentioned above. Still needs more alignment though.

packages/cli/src/lib/doctor/live.ts
```diff
@@ -0,0 +1,187 @@
...
+
+import { createEventEnvelope, makeDsn, serializeEnvelope } from "@sentry/core";
+import { queryEvents } from "../api/explore.js";
+import { sendEnvelopeRequest } from "../envelope/transport.js";
+import { logger } from "../logger.js";
+import type { Capture, CheckResult, ServerFacts } from "./types.js";
+
+const ID = "live.roundtrip";
+// First attempt is immediate; the rest are 2s apart → ~1 min.
+const DEFAULT_POLL_ATTEMPTS = 31;
```
BYK, line 19: Why 31??

packages/cli/src/lib/doctor/live.ts
```diff
@@ -0,0 +1,187 @@
...
+ * The one write doctor can perform, and only when asked.
+ *
+ * Delivery is the real test: if the POST resolves, this machine can reach
+ * ingest, which is the only failure mode the other liveness signals cannot
+ * see. The search poll is a bonus confirmation, and its absence is a warning
+ * rather than a failure — Sentry's index lags, and calling a healthy install
+ * broken because of that lag is worse than saying "sent, not yet visible".
+ */
+
+import { createEventEnvelope, makeDsn, serializeEnvelope } from "@sentry/core";
```
BYK, line 11: We already have a built-in command to send events to a DSN, we shouldn't be reinventing the wheel here

packages/cli/src/lib/doctor/manifests.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Dependency manifests, reduced to "which Sentry packages, at which versions".
+ *
+ * Two code paths only: JSON manifests get parsed properly; everything else
+ * gets one regex sweep. That is deliberate — doctor needs the SDK name and
+ * version, not a faithful model of nine packaging formats.
+ */
+
+import type { ParsedManifest } from "./types.js";
+
+const MANIFEST_BASENAMES =
```
BYK, line 11: I like that this is only a 100-line file but again, all this meta information should live in an SDK-controlled central place. Probably the release registry. Then we can either dynamically download or compile that information or embed at build time. I prefer the dynamic version though.

packages/cli/src/lib/doctor/markers.ts
```diff
@@ -0,0 +1,180 @@
+/**
+ * Where Sentry gets configured, as data.
```
BYK, line 2: Again, CLI is not the place for this information

packages/cli/src/lib/doctor/redact.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Redaction and untrusted-input validation for captured project files.
```
BYK, line 2: Don't understand why we need this but if we really do need it, it should be a generic helper at the output processing layer, not a command-specific exception.

packages/cli/src/lib/doctor/report.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+import { isatty } from "node:tty";
+// biome-ignore lint/performance/noNamespaceImport: Sentry SDK recommends namespace import
+import * as Sentry from "@sentry/node-core/light";
+import { detectAgent } from "../detect-agent.js";
+import { logger } from "../logger.js";
+import type { DoctorReport } from "./render.js";
+
+const FLUSH_TIMEOUT_MS = 3000;
```
BYK, line 16: Why do we need a custom flush timeout and why is it specific to here?

packages/cli/src/lib/dsn/code-scanner.ts
```diff
@@ -215,7 +215,7 @@ function isCommentedLine(trimmedLine: string): boolean {
  * Get the expected Sentry host for DSN validation.
  *
  * Self-hosted (SENTRY_URL set): only DSNs matching the configured
- * host are valid. SaaS: only `*.sentry.io` DSNs are valid.
+ * host are valid. SaaS: `*.sentry.io` and `*.sentry.gg` DSNs are valid.
```
BYK, line 218: This needs to be reverted I guess?

BYK (changes requested):
I like the idea and I think the implementation is close to being ready. A lot of the things need to be extracted out from the CLI though, otherwise we'll keep piling information here and miss the opportunity to give control back to SDKs and make this proper once again. 

Some stuff can live here, especially for better DSN detection but even some of that (including our existing code) should be moved out. 

Finally, I think there are some missed opportunities to leverage the existing code in the CLI. This PR seems to reinvent a few wheels unnecessarily and is quite large to review and ship.

romtsn:
@BYK sorry, this is definitely far from being ready for review (should've made it a draft), and I haven't had even a slightest look at the implementation myself 🙈 Daniel just asked to open a PR so he could take a look and I just did that.

I like the idea of having SDK-specific bits in the registry, so I'll look into that (and all of your other comments) after I'm back from sabbatical. Tree-sitter is also something I'd love to use here (I thought the agent had opted for it, but apparently it hadn't).

BYK:
@romtsn no worries! Looking forward for the updates!

PR #1516, opened 2026-08-31
fix(alerts): Remove deprecated transactions dataset

mjq:
`transactions` is no longer a valid dataset for alerts (the same use cases are supported by spans). All existing alerts have been migrated off of those datasets and creating new alerts that use them has been blocked on all billing plans.

Remove all references to transactions datasets from alerts code and skills. 

See BROWSE-682.

### DIFF
--- a/packages/cli/src/commands/alert/mutation-utils.ts
+++ b/packages/cli/src/commands/alert/mutation-utils.ts
@@ -7,7 +7,6 @@ import { ValidationError } from "../../lib/errors.js";
 const ISSUE_MATCH_MODES = new Set(["all", "any"]);
 const METRIC_DATASET_VALUES = new Set([
   "errors",
-  "transactions",
   "sessions",
   "events",
   "spans",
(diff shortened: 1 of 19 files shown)


packages/cli/src/commands/alert/mutation-utils.ts
```diff
@@ -33,12 +32,10 @@ const METRIC_DATASET_VALUES = new Set([
 const METRIC_DATASET_ALIASES: Record<string, string> = {
```
BYK, line 37: You should add `span: "spans"` at least and I may even add `transaction: "spans"` and `transactions: "spans"` for a smoother transition.

BYK (approved):

packages/cli/src/commands/alert/mutation-utils.ts
mjq: @BYK The super slick thing to do might be for requests for `transaction(s)` to route to `spans` with `is_transaction:true` in the query. Too magic or just magic enough? Would result in some new special case code of course.

BYK: Love it. We may print a warning/tip using the built-in system we do for gentle nudges. Like do the routing but also tell them "the right way" gently.

BYK:
Jared, take over the PR address all review discussions and then get it merged

PR #1521, opened 2026-09-02
feat(dashboard): Render graphics with Spleen

MathurAditya724:
The Sixel dashboard renderer now embeds Spleen 8x16 glyphs instead of Cozette, while retaining the current terminal-cell scaling and dashboard box, block, and sparkline coverage. The packaged third-party notice now carries Spleen's BSD-2-Clause attribution.

image


packages/cli/src/lib/formatters/spleen-font.ts
```diff
@@ -0,0 +1,105 @@
...
+);
+
+/**
+ * Printable ASCII and glyphs needed by the dashboard's box, sparkline, and
+ * text output, generated from Spleen 2.2.0's 8x16 BDF release. Each glyph has
+ * 16 rows packed into a byte, with bit 7 as the leftmost pixel. Source commit:
+ * 57f9219328c9f5873085320fe8bc8f7dd34b8791. Source BDF SHA-256:
+ * b38b32a66920068965a3101f98071d310c5c74659fe86e55d346140770f8f6e8.
+ */
+const GLYPH_ROWS = Uint8Array.from(
```
BYK: Can we not embed this as an asset and read it at runtime instead of doing this insanity?

BYK (commented):
Screenshots please

packages/cli/src/lib/formatters/spleen-font.ts
MathurAditya724: Moved the Spleen subset to a binary asset. The npm bundle ships it at `dist/assets/spleen-8x16.bin`, while standalone SEA builds embed the same asset and load it with `node:sea`. Local npm packaging and a single-platform SEA build both passed.

packages/cli/test/lib/formatters/pixel-canvas.test.ts
```diff
@@ -16,45 +16,38 @@ function opaquePixelCount(image: { data: Uint8Array }): number {
...
-    expect([...getCozetteGlyph("\u6f22")]).toEqual([...getCozetteGlyph("?")]);
+    expect([...getSpleenGlyph("\u6f22")]).toEqual([...getSpleenGlyph("?")]);
   });
 
   test("includes the dashboard axis, sparkline, and ellipsis glyphs", () => {
-    const questionMark = [...getCozetteGlyph("?")];
+    const questionMark = [...getSpleenGlyph("?")];
     for (const character of [
       "│",
       "└",
```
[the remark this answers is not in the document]
MathurAditya724: Added `■` to the coverage loop. `▀` already has an exact bitmap assertion immediately above, so it remains covered by the dedicated test.

PR #1526, opened 2026-09-02
fix(sourcemap): let `inject` run without authentication

### LINKED ISSUE #1525: `sourcemap inject` requires authentication even though it never calls the API
### Summary

Thank you for maintaining the CLI. `sentry sourcemap inject <dir>` exits with `Not authenticated. Run 'sentry auth login' first.` when no credentials are present, although it only rewrites local JS and `.map` files and never calls the Sentry API. It worked without a token up to 0.22.0. This blocks build steps that inject Debug IDs without a token and leave `sourcemap upload` to a separate, credentialed job.

I searched open and closed issues and PRs (`inject`, `debug id`, `Not authenticated`) and did not find an existing report.

### Environment

sentry 0.44.1 (npm), Node.js 26.8.1, macOS arm64; also reproduced on GitHub Actions ubuntu-24.04.

### Steps to reproduce

```sh
mkdir inject-demo && cd inject-demo
printf 'console.log(1)\n//# sourceMappingURL=a.js.map\n' > a.js
printf '{"version":3,"file":"a.js","sources":["a.ts"],"names":[],"mappings":"AAAA"}' > a.js.map

unset SENTRY_AUTH_TOKEN SENTRY_TOKEN
SENTRY_CONFIG_DIR="$(mktemp -d)" sentry sourcemap inject .
# same, per version: npx -y -p sentry@0.22.0 sentry sourcemap inject .
```

### Actual behavior

```text
Error: Not authenticated. Run 'sentry auth login' first.
```

Exit code 10, no file modified. 0.21.0 and 0.22.0 inject successfully; 0.23.0 through 0.44.1 fail. `sentry sourcemap resolve .` succeeds under the same conditions.

### Expected behavior

`sourcemap inject` runs without credentials, like `sourcemap resolve`. `sourcemap upload` should keep requiring authentication.

### Potential commit/PR that introduced the regression

#611 (`0d322393`, released in 0.23.0) made authentication the default for every `buildCommand` and opted out `auth *`, `help`, `schema`, and `cli *`, but not the sourcemap commands added a week earlier in #547. On current `main`:

- `injectCommand` has no `auth` setting: https://github.com/getsentry/cli/blob/9941f84465902f56d32e2ca0d258aa674968bb71/packages/cli/src/commands/sourcemap/inject.ts#L56
- `resolveCommand` opts out ("Purely local file operation — no Sentry API calls, no auth needed."): https://github.com/getsentry/cli/blob/9941f84465902f56d32e2ca0d258aa674968bb71/packages/cli/src/commands/sourcemap/resolve.ts#L119-L120
- The unit tests do not catch this because `test/preload.ts` sets a placeholder `SENTRY_AUTH_TOKEN` for every test: https://github.com/getsentry/cli/blob/9941f84465902f56d32e2ca0d258aa674968bb71/packages/cli/test/preload.ts#L106

### Other information

Workaround: export any non-empty `SENTRY_AUTH_TOKEN` for the inject step; the guard only checks that a token exists. This report is only about the guard on `sourcemap inject`, not the auth-by-default policy. I have a small fix with a regression test ready and will open a PR referencing this issue. Please let me know if the current behavior is intended.

Gyeonghun-Park:
Fixes #1525

## Problem

`sentry sourcemap inject` throws `AuthError("not_authenticated")` when no credentials exist, although it only rewrites local JS and `.map` files and never calls the Sentry API. #611 made authentication the default for every `buildCommand` and opted out the commands that work without a token, but `sourcemap inject` (added in #547 a week earlier) was not on that list, so it has required credentials since 0.23.0. `sourcemap resolve`, the other local-only command in the group, already sets `auth: false`.

## Fix

Set `auth: false` on `injectCommand`, with the same comment `resolve.ts` uses. `sourcemap upload` still requires authentication; no docs or generated files change.

## Tests

New `test/commands/sourcemap/inject.test.ts` runs the command with a fresh config dir and with `SENTRY_AUTH_TOKEN`, `SENTRY_TOKEN`, and `SENTRY_FORCE_ENV_TOKEN` cleared, then asserts the JS receives a `//# debugId=` comment and the map a `debugId` field. It fails on `main` with `AuthError: Not authenticated. Run 'sentry auth login' first.` and passes with this change. The existing suite did not catch the regression because `test/preload.ts` sets a placeholder token for every test.

Verified locally (macOS arm64, Node.js 26.8.1, pnpm 10.11.0): `biome check` on the touched files, `pnpm typecheck` (no generated-file changes), and `vitest run test/commands/sourcemap` (36 passed). In `pnpm test`, 7 tests in `test/lib/time-range.test.ts` and `test/lib/delta-upgrade.mocked.test.ts` fail in this environment on an unmodified `main` as well (timezone-dependent, local TZ is UTC+9), so they are unrelated. E2E and binary builds were not run.

## Breaking change

No.

Thank you for reviewing. Happy to move the test next to the existing inject tests in `upload.test.ts` or adjust the approach if you prefer.

### DIFF
--- a/packages/cli/src/commands/sourcemap/inject.ts
+++ b/packages/cli/src/commands/sourcemap/inject.ts
@@ -70,6 +70,8 @@ export const injectCommand = buildCommand({
       "  sentry sourcemap inject ./out --dry-run\n" +
       "  sentry sourcemap inject ./maybe-empty --allow-empty",
   },
+  // Purely local file operation — no Sentry API calls, no auth needed.
+  auth: false,
   output: {
     human: formatInjectResult,
   },
--- a/packages/cli/test/commands/sourcemap/inject.test.ts
+++ b/packages/cli/test/commands/sourcemap/inject.test.ts
@@ -1,0 +1,74 @@
+/**
+ * Tests for `sentry sourcemap inject` as a purely local command: it must
+ * run without any stored or env-provided credentials, since it never calls
+ * the Sentry API. Regression test for the `auth: false` opt-out.
+ */
+
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { injectCommand } from "../../../src/commands/sourcemap/inject.js";
+import { useEnvSandbox, useTestConfigDir } from "../../helpers.js";
+
+type InjectFuncArgs = {
+  ext?: string;
+  "dry-run"?: boolean;
+  "allow-empty"?: boolean;
+};
+type CmdFunc = (
+  this: unknown,
+  flags: InjectFuncArgs,
+  dir: string
+) => Promise<unknown>;
+
+function makeContext() {
+  return {
+    stdout: { write: vi.fn(() => true) },
+    stderr: { write: vi.fn(() => true) },
+    cwd: "/tmp",
+  };
+}
+
+describe("sourcemap inject command — runs without authentication", () => {
+  // Fresh config dir → no stored OAuth token; env sandbox → no env token.
+  // preload.ts sets a fake SENTRY_AUTH_TOKEN so the auth guard passes for
+  // most tests; this suite deliberately removes it.
+  useTestConfigDir("sentry-inject-noauth-");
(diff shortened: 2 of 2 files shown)


BYK (approved):
Oops and thanks!

PR #1535, opened 2026-09-04
fix(init): gate unsupported --features and tag dry-run runs

betegon:
Two small `init` hardening fixes surfaced by a wizard-run triage.

## 1. Gate unsupported `--features` (source maps, metrics, attachments)

`sourcemaps`, `metrics`, and `attachments` are non-selector products — the interactive selector never offers them, but `--features` still accepted them, leaving a half-configured integration (e.g. source-map upload needs an auth token this wizard does not provision). Removed from `FEATURE_ALIASES`, `SUPPORTED_FEATURE_NAMES`, and the flag help, so they now fail loudly (`Unknown init feature "sourcemaps"`) until their setup flow is automated.

## 2. Tag `wizard.dry_run` in telemetry

Dry runs (`--dry-run`) apply and install nothing, so the CLI's post-apply `verify-setup` runtime check runs the app's dev command against an **unmodified** project and reports **`init verification failed`** (CLI-2HP) — a false positive. Set `wizard.dry_run` once at `runWizard` start (global scope, so every CLI event inherits it, including that capture). `wizard.dry_run:false` then isolates real runs from dry-run noise.

Companion server change tags the same `wizard.dry_run` for the server-side `Verification found issues` (CLI-SERVER-22) — see below.

## Verification

- `tsc --noEmit`: no new errors in the changed files.
- `test/commands/init.test.ts`: 50/50 — updated the two normalization tests + the "Supported features:" message, and added a `test.each` locking in the rejection of the three gated features.
- `interactive.test.ts` + `clack-utils.test.ts`: 45/45 — the interactive selector's filtering and the display labels are separate code paths, unaffected by the flag change.

### DIFF
--- a/packages/cli/plugins/sentry-cli/skills/sentry-cli/references/init.md
+++ b/packages/cli/plugins/sentry-cli/skills/sentry-cli/references/init.md
@@ -18,7 +18,7 @@ Initialize Sentry in your project (experimental)
 **Flags:**
 - `-y, --yes - Accept non-interactive defaults (requires --features outside a TTY)`
 - `-n, --dry-run - Show what would happen without making changes`
-- `--features <value>... - Features to enable: errors,tracing,logs,replay,metrics,profiling,sourcemaps,crons,attachments,agent-tracing,mcp-observability`
+- `--features <value>... - Features to enable: errors,tracing,logs,replay,profiling,crons,agent-tracing,mcp-observability`
 - `-t, --team <value> - Team slug to create the project under`
 - `--app <value> - App to initialize in a monorepo (required with --yes when multiple apps are detected)`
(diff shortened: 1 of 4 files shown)


betegon:
Companion server PR (the server-side half of the `wizard.dry_run` tag, covering CLI-SERVER-22): getsentry/cli-init-api#267

betegon:
The red `Lint & Typecheck` here is a pre-existing, repo-wide Biome bug — `output.ts` exceeds Biome's 200k type-inference limit via valibot's `GenericSchema`, not this PR's code. It fails #1499 and #1501 identically, and a re-run reproduced it deterministically. Fixed in #1536 (force-ignores `output.ts`, mirroring the existing `custom-ca.ts` exclusion). Once #1536 lands I'll rebase this branch onto main and CI will go green.

PR #1572, opened 2026-09-10
feat(wasm-debug-files): Add prepare command for WASM debug setup

d2anamaria:
## Problem

Preparing WebAssembly for Sentry takes two tools today. You run `wasm-split` to inject a `build_id` and pull DWARF out into a companion file, then run `sentry debug-files upload --type wasm` on that companion. Nothing tells you when a module was built without usable debug info, so the mistake surfaces later as an unsymbolicated stack trace.

## Solution

`sentry debug-files prepare <path>...` does both steps in one command.

```bash
sentry debug-files prepare ./dist
```

It scans the given files and directories for `.wasm` modules, splits the ones carrying inline DWARF, and uploads the companions. Org and project are auto-detected from DSN, env vars, or config defaults. `--dry-run` and `--no-upload` need no credentials.

## What it does per module

For a module with inline DWARF:

1. Inject a `build_id` if it has none.
2. Write a `*.debug.wasm` companion keeping every section, including Code and DWARF.
3. Strip the `.debug_*` sections from the deployable module, in place.
4. Point the deployable at the companion via `external_debug_info`.
5. Upload the companion.

The deployable keeps its original path, so your build artifact does not move. Both files carry the same `build_id`, which is how Sentry matches a stack frame to its debug file. The companion keeps the Code section because DWARF addresses are relative to it.

A module that cannot be split is still stamped with a `build_id`, then reported with a warning. Stamping matters: without a `build_id` a module can never be symbolicated, not even from a debug file uploaded later. Nothing is uploaded for it.

| Outcome | Meaning | Uploaded |
| --- | --- | --- |
| Split | Had DWARF; companion written | Yes |
| Already prepared | Companion with matching `build_id` already exists | Yes |
| Would split | `--dry-run` preview | No |
| Skipped | Not splittable; stamped, warning printed | No |

## Warnings

A module without usable debug info does not fail the run unless `--require-dwarf` is set. The warning says why it was skipped:

- `no line-level symbolication (name/symtab only)`
- `no debug information; rebuild with DWARF (Emscripten -g, wasm-pack dwarf-debug-info)`
- `already stripped (build_id present, no debug sections); splitting would produce a useless companion`
- `has external_debug_info but no local companion with matching build_id`

Re-running is safe. An already-prepared pair is detected and left alone rather than overwritten with an empty companion, and a module keeps the `build_id` it was stamped with on the first run.

## Options

| Flag | Effect |
| --- | --- |
| `--dry-run` | Classify only; write nothing, upload nothing |
| `--no-upload` | Split, but skip the upload |
| `--require-dwarf` | Fail if any scanned module lacks DWARF |
| `--out-dir <DIR>` | Put companions here; deployables are always stripped in place |
| `--strip-names` | Also drop the `name` section from split deployables; the companion keeps it |
| `--build-id <UUID>` | Use a fixed build id instead of a random one, for exactly one module that has none |
| `--include-sources` | Also upload a source bundle per companion |
| `--wait` / `--wait-for <SECS>` | Wait for server-side processing |
| `--ignore <GLOB>` / `--ignore-file <FILE>` | Skip paths while scanning |

`--require-dwarf` is the CI guard, and it runs before anything is uploaded, so a build missing debug info fails without pushing files first. A module pointing at an external companion counts as having DWARF, so a dangling pointer does not fail the build.

`--ignore` globs are relative to the tree you point at: `--ignore 'vendor/**'` with `prepare ./dist` means `./dist/vendor`.

## Scanning rules

Directories are walked recursively. `*.debug.wasm` files are skipped, since they are outputs of an earlier run. Naming a non-`.wasm` file directly is an error; a directory with no modules is just an empty scan.

## Automation

`--json` reports the outcome per module, so a build script can act on the classification instead of grepping logs:

```json
{
  "org": "my-org",
  "project": "my-project",
  "uploaded": true,
  "filesUploaded": 1,
  "modules": [
    {
      "path": "dist/app.wasm",
      "action": "split",
      "quality": "dwarf",
      "buildId": "…",
      "companion": "dist/app.debug.wasm"
    }
  ]
}
```

The command exits non-zero when `--require-dwarf` fails, and when a companion fails server-side processing under `--wait`.

## Relationship to wasm-split

`sentry wasm-split` splits one module and stops there. This command is the pipeline around it: it
scans directories, classifies each module's debug quality, skips what it should not touch, and
uploads the companions. The split itself is the same `splitWasm` call.

Elsewhere the CLI reads debug files through `@sentry/symbolic` that wrapper (only reads).
This command rewrites modules: it injects `build_id`, strips `.debug_*`, and adds
`external_debug_info`. So it parses sections directly.

That is also why the two disagree on `build_id`. This command follows the tool convention;
`symbolic` reads it one byte shifted. The id sent here is advisory — the server re-parses the module
and stores its own key — so nothing breaks today. See
[getsentry/symbolic#1069](https://github.com/getsentry/symbolic/issues/1069).

## Limitations

- A module without DWARF gets a `build_id` but no debug file. Rebuild with DWARF and re-run to get line-level frames; the `build_id` is preserved.
- `external_debug_info` records the companion filename only. `prepare` does not expose `wasm-split`'s `--external-dwarf-url`; reach for `sentry wasm-split` if you need a custom URL. Sentry resolves by `build_id`, so this does not affect symbolication.
- Splitting always strips. There is no mode that writes a companion and leaves the deployable intact.
- With `--out-dir`, re-running without the same `--out-dir` will not find the companion and reports a dangling reference.


d2anamaria:
A deploy `.wasm` can end up with `external_debug_info` (pointer to companion) but no `build_id` when the module was split outside `wasm-split` (custom script/tool) and the build never stamped ids (no `--build-id` / equivalent at compile). The companion may still be on disk with its own id. Today `prepare` skips with a warning and stamps a new random id on the deploy module only, which breaks the pair permanently. Would it be safe to treat this as repairable — read the companion's id (or generate one and write it to both), instead of skipping upload?

BYK:
>Would it be safe to treat this as repairable — read the companion's id (or generate one and write it to both), instead of skipping upload?

Sounds like we should, yes. Also a simple warning in this case sounds like the wrong action as we are essentially uploading broken stuff?

packages/cli/src/commands/debug-files/prepare.ts
```diff
@@ -0,0 +1,536 @@
...
+      flags["ignore-file"]
+    );
+
+    // Companions are inputs to a previous run, never candidates themselves.
+    const candidates = (await scanPaths(paths)).filter(
+      (path) =>
+        isWasmPath(path) &&
+        !isDebugCompanionPath(path) &&
+        !isIgnored(path, ignoreMatchers, paths)
+    );
```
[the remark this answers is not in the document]
d2anamaria: This would require another directory pass to parse every candidate and collect which paths are pointed at by someone else's external_debug_info, and exclude them from the next scan. Is it a too narrow edge case to add this extra pass&parse? 
This solution only helps when the deployable is in the same scan and still carries its pointer, so custom-named companions sitting alone still can't be detected.

packages/cli/src/lib/wasm/prepare.ts
```diff
@@ -0,0 +1,624 @@
...
+    };
+  }
+
+  const split = splitWasm(bytes, {
+    companion: true,
+    strip: true,
+    ...(options.buildId ? { buildId: options.buildId } : {}),
+    stripNames: options.stripNames ?? false,
+    externalDebugInfo: basename(expectedCompanion),
+  });
```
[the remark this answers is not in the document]
d2anamaria: This brings up a design decision:
Do we want our tool to:
- skip stripping, keep old companion, in case the users intent was to upload the unstripped module to browser + warning that stripping was skipped
- always split, replace the old `external_debug_info` section with new one, and provide the user with a warning (although cli will display the action "strip" next to the module's name)

packages/cli/src/commands/debug-files/prepare.ts
```diff
@@ -0,0 +1,536 @@
...
+    for (const path of candidates) {
+      results.push(
+        await prepareWasmFile(path, {
+          dryRun,
+          outDir: flags["out-dir"],
+          buildId,
+          stripNames: flags["strip-names"],
+        })
+      );
+    }
```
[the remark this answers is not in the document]
d2anamaria: changed s.t. --build-id flag is rejected for directories and multiple paths

apps/cli-docs/src/fragments/commands/debug-files.md
```diff
@@ -49,8 +49,47 @@ sentry debug-files upload ./build --il2cpp-mapping --include-sources
...
+# Write companions elsewhere; modules are still stripped in place
+sentry debug-files prepare ./dist --out-dir ./symbols
+
+# Fail the build if any module was compiled without DWARF
+sentry debug-files prepare ./dist --require-dwarf
 ```
 
+## Notes on `prepare`
+
+- `debug-files prepare` replaces the two-step `wasm-split` + `debug-files
```
BYK: Is this information relevant to our users? If this is a behavioural change from the old CLI, it should be under "migrating from v3 to v4" section.

apps/cli-docs/src/fragments/commands/debug-files.md
```diff
@@ -49,8 +49,47 @@ sentry debug-files upload ./build --il2cpp-mapping --include-sources
...
+  keeps its path; only the companion is new.
+- The companion must keep the Code section — DWARF addresses are relative to it,
+  so a companion without it cannot be symbolicated.
+- Modules without DWARF are still stamped with a `build_id` and reported with a
+  warning rather than failing the run. Sentry matches a frame to its debug file
+  by `build_id`, so stamping now keeps symbolication possible later.
+- Name/symtab-only modules are not uploaded: the `name` section stays in the
+  deployable module and runtimes read function names from it directly, so a
+  debug file built from one adds nothing to the stack trace.
+- Running the command twice is safe. A module whose companion already exists
```
BYK: Maybe we can say "the command is idempotent" and leave it there?

packages/cli/src/commands/debug-files/read-file.ts
```diff
@@ -30,3 +34,22 @@ export async function readDebugFile(path: string): Promise<Buffer> {
...
+ * Read a source file for source-bundle resolution.
+ *
+ * A debug file names the sources it was compiled from, but those sources are
+ * often absent on the machine running the upload. A missing one is expected, so
+ * it is skipped rather than failing the bundle.
+ *
+ * @param sourcePath - Path the debug file references.
+ * @returns The contents, or `null` when the file is not available locally.
+ */
+export function readSourceFile(sourcePath: string): Uint8Array | null {
```
BYK, line 48: Why do we need a helper like this and why is it not using async or promisified versions?

packages/cli/src/commands/debug-files/prepare.ts
```diff
@@ -0,0 +1,571 @@
...
+ *
+ * Unlike {@link uploadPath} this counts `would-split`, so a dry run reports the
+ * work it would do rather than the (always empty) set it would upload.
+ */
+function producesCompanion(result: PrepareResult): boolean {
+  return (
+    result.action === "split" ||
+    result.action === "would-split" ||
+    result.action === "already-prepared"
+  );
```
BYK: Might be better to create `const COMPANION_PRODUCING_ACTIONS = new Set(["split", "would-split", "already-prepared"]);` and do `COMPANION_PRODUCING_ACTIONS.has(result.action)` for clarity, especially if there's a chance of adding more action types.

packages/cli/src/commands/debug-files/prepare.ts
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/**
+ * Compile `--ignore` globs and `--ignore-file` entries into path matchers.
+ *
+ * The ignore file is read as gitignore-style lines: blanks and `#` comments are
+ * dropped and every other line is treated as a glob. Patterns are not split on
+ * commas, because a glob's brace group (`{a,b}`) legitimately contains them.
+ */
+async function buildIgnoreMatchers(
```
BYK: We should already have helpers doing similar stuff in our DSN-discovery code. Let's check that and consolidate/reuse as much as possible.

packages/cli/src/commands/debug-files/prepare.ts
```diff
@@ -0,0 +1,571 @@
...
+/**
+ * Path of a scanned file relative to the scan root it came from.
+ *
+ * Ignore globs are written against the tree the user pointed at — `--ignore
+ * 'vendor/**'` for `prepare ./dist` means `./dist/vendor` — so they must be
+ * tested against a root-relative path rather than one relative to the process
+ * working directory. Separators are normalized to `/` because glob patterns
+ * always use them.
+ */
+function pathRelativeToRoot(path: string, roots: string[]): string {
```
BYK, line 249: Same as above, I think all this directory walking code should already exist once or twice in the codebase, we shouldn't be re-inventing the wheel.

packages/cli/src/commands/debug-files/prepare.ts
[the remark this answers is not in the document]
BYK: Not fully familiar with the logic but why can we not build an in-memory tree as we do the first scan and use that later on?

packages/cli/src/lib/formatters/wasm-prepare.ts
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+  return lines;
+}
+
+/** Format human-readable output for the prepare result. */
+export function formatPrepareResult(data: PrepareCommandResult): string {
+  const sections = data.modules.map(moduleSection);
+  return renderMarkdown(
+    [...summaryLines(data), ...sections.flatMap((section) => ["", section])]
```
BYK: Using array spread is quite wasteful here, `summaryLines(data).concat(sections.flatMap...)` should be more efficient.

packages/cli/src/lib/wasm/binary.ts
```diff
@@ -68,7 +71,7 @@ const SECTION_ORDER = [
   [8, "start"],
   [9, "element"],
   [12, "data count"],
-  [10, "code"],
+  [CODE_SECTION_ID, "code"],
```
BYK, line 110: Why only name this??? Either do all or none :D

packages/cli/src/lib/wasm/prepare.ts
BYK, line 1: FYI did not review this file yet.

BYK (changes requested):

packages/cli/src/lib/wasm/prepare.ts
```diff
@@ -0,0 +1,615 @@
...
+ *
+ * @param wasmPath - Path of the input module.
+ * @param outDir - Directory to place the companion in, or omitted for
+ *   alongside the input.
+ */
+export function companionPath(wasmPath: string, outDir?: string): string {
+  const name = basename(wasmPath).replace(WASM_EXTENSION, "");
+  const fileName = `${name}${COMPANION_SUFFIX}`;
+  return join(outDir ?? dirname(wasmPath), fileName);
+}
```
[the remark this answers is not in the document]
d2anamaria: not sure what’s the best solution for this.
- remove flag, companions will always sit in the deploy directory
- name by build_id, filename stops being human recognizable
- another?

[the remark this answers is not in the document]
BYK: >name by build_id, filename stops being human recognizable

I think this makes the most sense.

apps/cli-docs/src/fragments/commands/debug-files.md
d2anamaria: Will remove the bullet point. Not a behavioural change, the other 2 cmds will still exist, this one just combines them and emits clearer warnings

packages/cli/src/commands/debug-files/read-file.ts
d2anamaria: Not new, it was already a private helper in upload.ts. I moved it so the new command can reuse it instead of duplicating the try/catch. Sync because that requirement bubbles up from symbolic.

packages/cli/src/lib/wasm/binary.ts
d2anamaria: it's the only one that gets referenced individually, but i can name them all for the sake of uniformity:))

loewenheim (commented):
On a high level I think this command is misnamed, on two counts:
1. It doesn't apply to debug files in general, only wasm.
2. "Prepare" doesn't sound like it uploads files, it sounds like something you call _before_ you upload them.

I'm just spitballing, but I could see this as a `--split-wasm` flag on `debug-files upload` that splits wasm files with default options (and you can either add options to customize the `wasm-split` behavior or tell users to run `wasm-split` manually if they need to customize it).

d2anamaria:
> I'm just spitballing, but I could see this as a `--split-wasm` flag on `debug-files upload` that splits wasm files with default options (and you can either add options to customize the `wasm-split` behavior or tell users to run `wasm-split` manually if they need to customize it).

I agree the name is a bit misleading, but I'd fix the name rather than folding it into upload. This command rewrites your build artifacts, while upload is read-only. Carrying the wasm-split flags over would bloat upload with options that do nothing unless you're uploading wasm.

loewenheim:
Yeah, that's fair both on the read-only point and the options point. Again, just a thought: maybe the new command could be folded into `wasm-split`.

PR #1573, opened 2026-09-10
fix(explore): Remove deprecated `transaction` dataset

mjq:
Remove calls to the deprecated/unsupported `transactions` dataset from the `explore` surface area.

Following the lead of #1516, you can still ask to explore `transaction(s)` and we'll route you to `spans` with `is_transaction:true` to smooth out the transition for users.

Fixes BROWSE-682.

### DIFF
--- a/packages/cli/src/commands/explore.ts
+++ b/packages/cli/src/commands/explore.ts
@@ -64,7 +64,7 @@ const log = logger.withTag("explore");
 /** Default fields when none specified — top errors view */
 const DEFAULT_FIELDS = ["title", "count()"];
 const DEFAULT_TRANSACTION_FIELDS = ["transaction", "count()"];
-const TRANSACTION_DATASET = "transactions";
+const LEGACY_TRANSACTION_DATASETS = new Set(["transaction", "transactions"]);
 const IS_TRANSACTION_FILTER = "is_transaction:true";
 const IS_TRANSACTION_FILTER_PATTERN = /(?:^|\s)is_transaction:true(?:\s|$)/;
 
@@ -89,8 +89,6 @@ const DATASET_ALIASES: Record<string, string> = {
   error: "errors",
   spans: "spans",
   span: "spans",
-  transactions: TRANSACTION_DATASET,
-  transaction: TRANSACTION_DATASET,
   metrics: "tracemetrics",
   logs: "logs",
   log: "logs",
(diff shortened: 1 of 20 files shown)


mjq:
@cursor review

packages/cli/src/commands/explore.ts
```diff
@@ -63,6 +63,10 @@ const log = logger.withTag("explore");
 
 /** Default fields when none specified — top errors view */
 const DEFAULT_FIELDS = ["title", "count()"];
+const DEFAULT_TRANSACTION_FIELDS = ["transaction", "count()"];
+const TRANSACTION_DATASET = "transactions";
```
BYK: wait, weren't we removing this dataset??

BYK (changes requested):

packages/cli/src/commands/explore.ts
mjq: ...oh, it's already gone. I was trying to replicate what was done in #1516 (`resolveEventsDataset` routes `transaction(s)` requests to the `spans` dataset) but you're right, since this dataset was already removed from the `explore` surface there's no backwards compat to maintain. I'll just get rid of this again.

BYK: I mean we _can_ support the mapping of transactions to span dataset mapping but this mapped `transaction` to `transactions` or something like that?

mjq: It's just a sentinel value for the downstream code to know to treat it differently (`spans` dataset query with `is_transaction:true` added to it)

mjq: Could also try to find a way to remove it from this hash and make the separate code path obvious, agree it's misleading as is

mjq: Does this read better in ac5b70ff2eb53db7db0b7b2d2c069a37ba901bb1?

packages/cli/src/commands/explore.ts
```diff
@@ -168,13 +166,13 @@ type ExploreData = {
...
-    throw new ValidationError(
-      `Invalid dataset "${value}". Must be one of: ${[...VALID_DATASETS].join(", ")}`,
-      "dataset"
-    );
+  if (resolved || LEGACY_TRANSACTION_DATASETS.has(lower)) {
+    return resolved ?? lower;
   }
-  return resolved;
+  throw new ValidationError(
+    `Invalid dataset "${value}". Must be one of: ${[...VALID_DATASETS].join(", ")}`,
```
BYK: Better to do `Array.from(VALID_DATASETS)`

BYK (approved):
Love it, thanks!

PR #1585, opened 2026-09-14
feat(agent-conversation): Link conversation IDs to Sentry

vgrozdanic:
Conversation IDs in `sentry agent-conversation list` now link to their Sentry pages in terminals that support hyperlinks. Responses without `webUrl` keep plain IDs, so older endpoint versions retain existing output.


BYK (approved):
Nice, thanks!

PR #1589, opened 2026-09-16
feat(wasm-split): port Symbolicator wasm-split

d2anamaria:
## Summary

Ports Symbolicator's [`wasm-split`](https://github.com/getsentry/symbolicator/blob/master/crates/wasm-split/src/main.rs) into the CLI as `sentry wasm-split`.

Same flags, same behaviour, same stdout (lowercase hex build id). Drop-in for the Rust binary. Nothing removed from Symbolicator.

## Notes

Validation matches `wasmbin`, not a full wasm engine.

We validate the module envelope (header, section ids, and section order), but we do not read or validate the contents inside each section. Those bytes are passed through unchanged. This matches Symbolicator's Rust wasm-split.

We intentionally do not use WebAssembly.validate, since it performs deeper checks than wasmbin and would reject inputs the Rust tool accepts.


packages/cli/src/lib/wasm/build-id.ts
loewenheim, line 1: All the UUID stuff should probably be handled by the uuid7 library, which is already a dependency.

packages/cli/src/lib/wasm/split.ts
```diff
@@ -0,0 +1,141 @@
...
+    moduleChanged = true;
+  }
+
+  // Before stripping, so the companion keeps every section.
+  const companion = options.companion ? encodeModule(sections) : undefined;
+
+  if (options.strip) {
+    const stripNames = options.stripNames ?? false;
+    sections = sections.filter((section) => !isStrippable(section, stripNames));
+    moduleChanged = true;
```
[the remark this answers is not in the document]
loewenheim: This is true, but it's on par with the Rust version.

packages/cli/src/commands/wasm-split.ts
```diff
@@ -0,0 +1,240 @@
...
+    // Before the main module, so a failure here never leaves a stripped binary
+    // with no companion to symbolicate it.
+    if (debugOut !== undefined && result.companion) {
+      await writeFile(debugOut, result.companion);
+    }
+
+    const output = flags.out ?? input;
+    if (result.moduleChanged) {
+      await writeFile(output, result.module);
+    }
```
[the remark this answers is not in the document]
loewenheim: > and is not drop-in with the Rust tool, which always writes --out.

This is not true, the Rust version behaves the same and only writes `out` if the file changed. See https://github.com/getsentry/symbolicator/blob/6d79a7851780f144c2c6d6de50b788cabeedf838/crates/wasm-split/src/main.rs#L147-L150.

loewenheim (commented):
Seems solid apart from the manual UUID parsing stuff.

packages/cli/src/lib/wasm/split.ts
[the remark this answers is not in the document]
BYK: Well, let's fix it if we can while we are rewriting?

packages/cli/src/lib/wasm/build-id.ts
BYK: Strong yes

packages/cli/src/lib/wasm/build-id.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Read the build id out of already-parsed sections.
+ *
+ * The first readable `build_id` wins, and a malformed section is skipped rather
+ * than treated as an answer — both matching the Rust tool, where a section that
+ * fails to decode never reaches `find_map`. A module should carry at most one
+ * id, but the distinction still matters: minting a fresh id for a module that
+ * already has a good one would orphan every debug file uploaded against it.
+ */
+export function buildIdFromSections(
```
loewenheim, line 66: IMO this is the only worthwhile function left in this file, and it could go in `binary.ts`. The rest are just converting between UUIDs, bytes, and strings.

I think the `UUID` type should be used wherever possible, including in the output of this function, and only be converted to a string or bytes when it needs to be written somewhere.

d2anamaria: I'd keep the wrappers: each handles a different concern in one place, uuidToBytes turns parse failures into CLI errors, and UUID on the read path is wrong because build_id is arbitrary bytes. The other cmd i'm working on will reuse the same helpers when it reads and stamps ids.

loewenheim (approved):

BYK (approved):

PR #1609, opened 2026-09-21
feat(issue): view multiple issues in one invocation

betegon:
## Summary

`sentry issue view` now accepts multiple IDs in one invocation, the same way `event view` already does: space-separated args, newline-separated tokens inside a single arg, bounded parallel fetches, and JSON that stays a flat object for one ID and becomes an array when more than one was requested.

The shared `issueIdPositional` tuple is unchanged, so `explain` / `plan` / `resolve` / `unresolve` / `archive` / `events` stay single-id. Commas are not split (they can be part of a value). `--web` still opens only the first issue.

## Test plan

- [ ] `sentry issue view FRONT-ABC --json` still returns a single object (`shortId` at the top level)
- [ ] `sentry issue view FRONT-ABC BACK-2 --json` returns a 2-element array
- [ ] `sentry issue view $'FRONT-ABC\nBACK-2' --json` matches two separate args
- [ ] `sentry issue view FRONT-ABC,BACK-2` is treated as one identifier, not two
- [ ] `sentry issue view A B -w` opens only A and warns that extras are ignored
- [ ] `sentry issue explain A B` is still single-id (shared positional untouched)


BYK:
> sentry issue view A B -w opens only A and warns that extras are ignored

I'd say it should open all up to a limit and the limit can be bypassed with `-f`

BYK:
> sentry issue explain A B is still single-id (shared positional untouched)

I'd also make this support multiple issues

betegon:
Addressed both points from @BYK:

- `issue view A B -w` now opens both. The default safety cap is 5 tabs, and `--force` opens all requested issues. I kept `-f` bound to the shared `--fresh` flag used across read-only commands, so the override is long-form `--force` rather than giving `-f` two meanings.
- `issue explain A B` now batches too, with the same space/newline input, concurrency bound, partial-failure behavior, and single-invocation compatibility guarantees as `issue view`.

The cleanup also fixed the CI trace nullability error, extracted issue formatting from the command, and centralized issue-batch orchestration.

PR #1615, opened 2026-09-21
fix(api): preserve cursor for partial final pages

betegon:
## Summary

`paginate()` currently uses the same `per_page` for every request. A limit of 150 therefore fetches 100 + 100 rows, trims the last 50, and correctly drops the server cursor because using it would skip those trimmed rows.

Track the remaining budget inside `paginate()` so that request becomes 100 + 50 and the final cursor points at the first unreturned row. `autoPaginate()` remains unchanged: same signature, same cursor loop, and the same defensive trim-and-drop behavior when an endpoint ignores `per_page`.

Split out of #1606 so the shared helper behavior can be reviewed independently.

## Test plan

- `pnpm exec vitest run test/lib/api/infrastructure.test.ts test/lib/auto-paginate.property.test.ts test/lib/api/traces.test.ts test/lib/api/replays.test.ts test/lib/api/logs.test.ts test/lib/api/conversations.test.ts`
- `pnpm run lint`
- generate schema/docs/SDK, then `pnpm exec tsc --noEmit`


betegon:
@BYK added you here as you added pagination if i'm not wrong. this should be just an improvement

BYK (approved):
Yeah, good catch!

